PR #21, opened 2015-02-05
Fixed incorrect name for source file of Nucleotide exercise

ErikSchierboom:
The source file of the Nucleotide exercise was named `Nucleotide.java`, whereas the expected class is named `DNA`. The class should thus be named `DNA.java`, which is fixed by this PR.


kytrinyx:
Thank you!

Issue #3, opened 2014-02-28
Implement Java Exercises


kytrinyx:
Copied from exercism/exercism.io#933 /cc @sit

## Minimum Exercises to Implement

The following exercises are required to be created for the path to be launch.

- [x] bob
- [x] hamming
- [x] word-count
- [x] anagram
- [x] nucleotide-count
- [x] phone-number
- [x] grade-school
- [x] robot-name
- [x] etl
- [x] meetup

## Additional Exercises to Implement

After the initial set of exercises the following should be implemented to have a more complete test suite.

- [x] space-age
- [x] gigasecond
- [x] triangle
- [x] scrabble-score
- [x] roman-numerals
- [x] binary
- [x] prime-factors
- [x] raindrops
- [x] allergies
- [x] strain
- [x] atbash-cipher
- [x] accumulate
- [x] crypto-square
- [x] trinary
- [x] rna-transcription
- [x] sieve
- [x] simple-cipher
- [x] octal
- [x] luhn
- [x] pig-latin
- [ ] pythagorean-triplet
- [ ] series
- [ ] difference-of-squares
- [ ] secret-handshake
- [x] linked-list
- [ ] wordy
- [x] hexadecimal
- [ ] largest-series-product
- [ ] kindergarten-garden
- [ ] binary-search-tree
- [ ] matrix
- [ ] robot-simulator
- [x] nth-prime
- [ ] palindrome-products
- [x] pascals-triangle
- [ ] beer-song
- [ ] sum-of-multiples
- [ ] queen-attack
- [ ] saddle-points
- [ ] ocr-numbers
- [x] simple-linked-list
- [x] linked-list
- [ ] point-mutations

## Not Allowed to Be Implemented 

These exercises were inspired by Java Ranch and are not allowed to be implemented.

- leap
- grains
- say

kytrinyx:
@hrj did you have a chance to talk to your team about possibly simplifying the test dependencies here?

kytrinyx:
We have a new process that you can use to find exercises to implement.
You can read all about it [here](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#porting-an-exercise-to-another-language-track).

I'm closing this issue in favor of the process described in the Contributing guide linked to above.

Thanks!

PR #27, opened 2015-04-02
Added exercise space-age (see #3)

pietvandongen:


kytrinyx:
Awesome thank you!

pietvandongen:
@kytrinyx Awesome, thank you too! I'll do more when I have some time in between stuff.

kytrinyx:
That would be great. Right now Java, Objective-C and PHP are the languages with the fewest exercises.

space-age/example.java
```diff
@@ -0,0 +1,69 @@
...
+        EARTH, MERCURY, VENUS, MARS, JUPITER, SATURN, URANUS, NEPTUNE
+    }
+
+    private double seconds;
+
+    private static final double EARTH_ORBITAL_PERIOD_IN_SECONDS = 31557600.0;
+    private static final int PRECISION = 2;
+    private static final Map<Planet, Double> relativeOrbitalPeriods = new HashMap<>();
+
+    static {
```
sit: Could fold this into the `Planet` enum itself. (Oddly enough, it's in the official tutorial :-)

https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

sit:
@pietvandongen hope you don't mind all the comments. Thanks for picking up the Java and even Java 8!

pietvandongen:
@sit I don't mind, thanks! I'm not sure if I can / will respond to all of them, but it's good to have these comments here anyway :)

PR #30, opened 2015-04-02
Added triangle exercise (see #3)

pietvandongen:


kytrinyx:
Thanks!

triangle/example.java
```diff
@@ -0,0 +1,53 @@
...
+    private double side2;
+    private double side3;
+
+    public Triangle(double side1, double side2, double side3) throws TriangleException {
+        this.side1 = side1;
+        this.side2 = side2;
+        this.side3 = side3;
+
+        if (allSidesAreZero() || hasImpossibleSides() || violatesTriangleInequality()) {
+            throw new TriangleException();
```
sit: From a more Java-y standpoint (and ignoring the checked vs unchecked exception debate), I would probably just want to use an `IlllegalArgumentException`.

Also, it might be helpful if the exception included a message indicating which condition was violated (and possibly which specific side in question is in violation).

triangle/example.java
```diff
@@ -0,0 +1,53 @@
+public class Triangle {
+
+    private double side1;
+    private double side2;
+    private double side3;
```
sit: `final`

triangle/example.java
```diff
@@ -0,0 +1,53 @@
...
+        this.side1 = side1;
+        this.side2 = side2;
+        this.side3 = side3;
+
+        if (allSidesAreZero() || hasImpossibleSides() || violatesTriangleInequality()) {
+            throw new TriangleException();
+        }
+    }
+
+    public TriangleKind getKind() {
```
sit: Might make sense to embed the `TriangleKind` enum directly in this class and just call it `Kind`.

triangle/example.java
```diff
@@ -0,0 +1,53 @@
...
+    private boolean hasImpossibleSides() {
+        return side1 < 0 || side2 < 0 || side3 < 0;
+    }
+
+    private boolean violatesTriangleInequality() {
+        return side1 + side2 <= side3 || side1 + side3 <= side2 || side2 + side3 <= side1;
+    }
+
+
+    public int getNumberOfUniqueSides() {
```
sit: `private`

PR #31, opened 2015-04-02
Moved Gigasecond solution code from main file to example file (closes #29)

### LINKED ISSUE #29: Gigasecond's main file contains the solution and should be move to the example file
And the ``Date()`` method should be called ``getDate()``.

pietvandongen:


kytrinyx:
Good catch.

PR #35, opened 2015-04-03
added strain exercise

PaulNoth:


pietvandongen:
Thanks a lot @PaulNoth! 

(I've renamed the main file to maintain track consistency (the C# track you used as an example does thing slightly different))

PR #36, opened 2015-04-04
Added exercise atbash-cipher

pietvandongen:


atbash-cipher/example.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class Atbash {
+
+    private static final int GROUP_SIZE = 5;
+    private static final String PLAIN = "abcdefghijklmnopqrstuvwxyz";
+    private static final String CIPHER = "zyxwvutsrqponmlkjihgfedcba";
+
+    public static String encode(String input) {
+        String encoded = stripInvalidCharacters(input).toLowerCase();
+        String cyphered = "";
```
sit: `ciphered`?

atbash-cipher/example.java
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        return filteredValue;
+    }
+
+    private static char applyCipher(char input) {
+        int idx = PLAIN.indexOf(input);
+
+        return idx >= 0 ? CIPHER.toCharArray()[idx] : input;
```
sit: Could extract a constant of `CIPHER.toCharArray()` --- not sure if the compiler would optimize this to avoid having to run this conversion every time.

atbash-cipher/example.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private static String splitIntoFiveLetterWords(String value) {
+        List<String> words = new ArrayList<>();
+
+        for (int i = 0; i < value.length(); i += GROUP_SIZE) {
+            words.add(i + GROUP_SIZE <= value.length() ? value.substring(i, i + GROUP_SIZE) : value.substring(i));
```
sit: Could extract local variables to make the calculation easier to understand.
```java
int end_index = i + GROUP_SIZE;
words.add(end_index > value.length() ? value.substring(i, end_index) : value.substring(i));
```
Could also rename `i` something like `start_index`.

atbash-cipher/example.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private static char applyCipher(char input) {
+        int idx = PLAIN.indexOf(input);
+
+        return idx >= 0 ? CIPHER.toCharArray()[idx] : input;
+    }
+
+    private static String splitIntoFiveLetterWords(String value) {
+        List<String> words = new ArrayList<>();
```
sit: Could pre-size this array based on `value.length() / GROUP_SIZE`.

PR #37, opened 2015-04-04
Added exercise accumulate

pietvandongen:


accumulate/example.java
```diff
@@ -0,0 +1,17 @@
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+public class Accumulate {
+
+	public static <T> Collection<T> accumulate(Collection<T> collection, Function<T, T> function) {
```
sit: Why use a `Collection`? The tests don't demand a collection... actually, they assume that the order of elements will be preserved.

(Also, not probably necessarily the right place for this discussion @kytrinyx, but this function seems more like `map` than `accumulate`. See for example [SICP §2.3.3](https://mitpress.mit.edu/sicp/full-text/book/book-Z-H-15.html#%_sec_2.2.3).)

pietvandongen: Good point about order specifically, I've created issue #38 for this.

kytrinyx: Yeah, in the Ruby standard library this behavior is called both `map` and `collect` (one is an alias of the other). I'm not sure why I chose `accumulate` as the name for this :/

PR #47, opened 2015-05-10
Define the missing class (with no content)

vdemeester:
Some of the ``java`` files to implement have a class defined, some are not. This pull request is there to add the ``public class`` definitions that miss.

I think it's better to have the class defined as, and it felt strange to have some where it was already defined and some not..

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/accumulate/src/main/java/Accumulate.java
+++ b/accumulate/src/main/java/Accumulate.java
@@ -1,2 +1,0 @@
-public class Accumulate {
-}
--- a/allergies/src/main/java/Allergies.java
+++ b/allergies/src/main/java/Allergies.java
@@ -1,2 +1,0 @@
-public class Allergies {
-}
--- a/anagram/src/main/java/Anagram.java
+++ b/anagram/src/main/java/Anagram.java
@@ -1,0 +1,2 @@
+public class Anagram {
+}
--- a/atbash-cipher/src/main/java/Atbash.java
+++ b/atbash-cipher/src/main/java/Atbash.java
@@ -1,2 +1,0 @@
-public class Atbash {
-}
--- a/binary/src/main/java/Binary.java
+++ b/binary/src/main/java/Binary.java
@@ -1,2 +1,0 @@
-public class Binary {
-}
--- a/bob/src/main/java/Bob.java
+++ b/bob/src/main/java/Bob.java
@@ -1,5 +1,0 @@
-public class Bob {
-    public String hey(String s) {
-        return null;
-    }
-}
--- a/crypto-square/src/main/java/Crypto.java
+++ b/crypto-square/src/main/java/Crypto.java
@@ -1,2 +1,0 @@
-public class Crypto {
-}
--- a/etl/src/main/java/Etl.java
+++ b/etl/src/main/java/Etl.java
@@ -1,0 +1,2 @@
+public Etl {
+}
--- a/gigasecond/src/main/java/Gigasecond.java
+++ b/gigasecond/src/main/java/Gigasecond.java
@@ -1,2 +1,0 @@
-public class Gigasecond {
-}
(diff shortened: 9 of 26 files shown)


kytrinyx:
I don't have strong opinions as to whether or not we should have the empty class or not. I think it makes sense to have it in the very first exercise, to help people get started, but part of the idea is that you should recognize the error message of an undefined class when running the test suite, and then respond to that error message by creating the missing class.

I'd be curious to hear what people involved in the Java track think.

/cc @chelseakomlo, @DouglasBrunner, @lackita, @SteffenBauer, @pietvandongen, @blemale, @marciojrtorres, @coulas, @sarod, @adericbourg, @Akrome, @ChristianWilkie

DouglasBrunner:
I think that the first few exercises should have a very low barrier to entry so that people can get used to the exercism workflow. After that, exposing users to issues/concepts they may not come across on their own will help them *gain a deep understanding of the fundamentals of [their] craft*.

With that said, it might be a good idea to include the class definitions until the user has been exposed to Java's various class related constructs (inheritance, interfaces, access modifiers, etc) and then remove them in subsequent exercises. Put another way, supplying the class definitions provides users with the tools they'll need to complete the exercises, and removing the definitions will force them to apply their newly acquired tools.

vdemeester:
@kytrinyx @DouglasBrunner Yes. The first one, in order, are : Etl, Nucleotide Conut, Word Count, Anagram, Hamming, Bob, … And 4 of them have no class defined. I also don't have a strong opinions of that but :
- It's getting in the way of the IDE(s) *helper*. For example, in Inteliij you can't use the ``import class`` action (because it doesn't exists, even though the file is present) or the ``create class`` (because when trying to create the class, it will complain the file is already there). Not sur if that is a goal or not, important or not, but I view exercism also as a way to *getting used the IDE(s)/editor* you're using (and thus, their features).
- There is no clear reason why on one exercise it is defined and on another it is not ; I do agree with @DouglasBrunner that, if not always, it might be better to have them define in the first exercises and then not.

Another way to handle this, would be to not have the ``.java`` file implementation in ``src/main/java`` (like in the [xgo](https://github.com/exercism/xgo) when the only file defined is the test files) — when additional classes/enums are provide (like [allergies](https://github.com/vdemeester/xjava/tree/master/allergies/src/main/java)), we would just leave them in plac. It's too bad git doesn't handle empty folder but a way to handle could be to had a ``.placeholder`` file in ``src/main/java`` when needed so the folder is there and seems empty.

Akrome:
I agree with @DouglasBrunner , maybe add for the first few, but not for the other. My rationale about not adding to all of them is that, by providing ANY kind of code, you are somehow influencing (albeit minimally) the coder. I can see some people going "Oh, this class is already defined, which means that, if other classes had been necessary, they would have also been already defined", thus limiting one's possibilities.

kytrinyx:
The idiomatic git hack for an empty directory is `.keep`, I think. So if the problem is that the IDE has a conflict between adding a class and an empty file, that would be a reasonable workaround. That still leaves the open question of _should_ the class be defined.

DouglasBrunner:
> @vdemeester : I do agree with @DouglasBrunner that, if not always, it might be better to have them define on the first exercises and then not.

> @Akrome : I agree with @DouglasBrunner , maybe add for the first few, but not for the other.

A small consensus is building, ;)

ChristianWilkie:
I haven't really contributed much but FWIW I agree with the consensus of including the class definitions for at least the first few exercises. I think it would help with lowering the barrier to entry for newer programmers.

vdemeester:
While waiting others to share their thoughts, I'll update my PR based on the small concensus we currently have :).

vdemeester:
Any more thoughts on this ? :blush:

adericbourg:
I might be late (only one month...) but I'm ok with that. Thanks for having done it!

kytrinyx:
Thanks! (And also, thanks for checking back in, it's been a hectic few weeks, and a few things have been lost in the cracks)

PR #52, opened 2015-06-27
scrabble-score: replace illegal word

kytrinyx:
multibillionaire can't be played as a word in a real scrabble game. See https://github.com/exercism/x-common/issues/86

### DIFF
--- a/scrabble-score/src/test/java/ScrabbleScoreTest.java
+++ b/scrabble-score/src/test/java/ScrabbleScoreTest.java
@@ -23,7 +23,7 @@ public class ScrabbleScoreTest {
                 {"f", 4},
                 {"street", 6},
                 {"quirky", 22},
-                {"MULTIBILLIONAIRE", 20},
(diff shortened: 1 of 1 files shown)


jtigger:
Dig the replacement word. :smile:

I ran the tests locally (since that isn't yet happening in CI); ran green, here.

kytrinyx:
Thanks! I don't have the java tests running locally. I really should get on that...

jtigger:
You shouldn't have to (beyond just installing gradle and running the build).  I suspect you've got higher priorities. ;)

After I finish with the getting started documentation, having a complete and automated build seems to be the next priority for the Java track.

I mentioned running the tests purely for documenting the fact that it happened.

kytrinyx:
> I suspect you've got higher priorities

How did you guess? :)

> After I finish with the getting started documentation, having a complete and automated build seems to be the next priority for the Java track.

Yeah, that sounds like a good set of priorities.

PR #58, opened 2015-08-10
Add test for same word in anagram.

jmluy:
Existing test (testIdenticalWordIsNotAnagram) for words only check that it contains the anagram but not that it shouldn't contain the same word.

### DIFF
--- a/anagram/src/test/java/AnagramTest.java
+++ b/anagram/src/test/java/AnagramTest.java
@@ -21,6 +21,13 @@ public class AnagramTest {
     }
 
     @Test
+    public void testSameWordIsNotAnagram() {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Can we change the assertion on `testIdenticalWordIsNotAnagram` so that it asserts the full list of expected results, rather than just that it contains the expected anagram instead of adding a new test?

jmluy:
Updated to compare list containing the only anagram.

kytrinyx:
I like this--it's a bit simpler than adding an extra test. Thanks!

kytrinyx:
Merged in 04a082121af34013d0a30e4a50d3d57202734876

PR #59, opened 2015-08-17
Test for Robot name allows other characters for the first 2 characters

jmluy:
First 2 characters in test allows numbers, lowercased letters or underscores.

### DIFF
--- a/robot-name/src/test/java/RobotTest.java
+++ b/robot-name/src/test/java/RobotTest.java
@@ -7,7 +7,7 @@ import static org.junit.Assert.assertThat;
 
 public class RobotTest {
 
-    private static final String EXPECTED_ROBOT_NAME_PATTERN = "\\w{2}\\d{3}";
+    private static final String EXPECTED_ROBOT_NAME_PATTERN = "[A-Z]{2}\\d{3}";
     private final Robot robot = new Robot();
 
     @Test


kytrinyx:
This looks good to me, but I'll defer to @jtigger to merge.

Would you mind squashing the commits please? (Here's a quick guide in case it's helpful. It assumes that you have a git remote named `upstream` that points to the exercism/xjava repository )http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request)

kytrinyx:
One question--do we want a test that clarifies that lower-case names are not allowed?

jmluy:
thanks, will check on squashing.

regarding the test, it's the test that's not accurate but we can have the test for lowercased names separately

kytrinyx:
Yeah, the lowercase thing is a separate problem (if it's a problem at all). Thanks!

jmluy:
Should it allow uppercased words? If it does, it might just need some more examples on the readme file or a clarification.

kytrinyx:
I'm not sure I understand the question. At the moment the README implicitly suggests that the name should be all uppercase. I don't think there's any reason to allow lowercase letters.

jtigger:
Hey y'all, I've been out-of-pocket for the past month.  I apologize for not being responsive, here.  Catching up now.

jtigger:
@jmluy, thank you for the contribution!  Your implementation is inline with how we're generally thinking about this problem (see exercism/exercism.io#2441).

In lieu of CI that runs the test against the example, I did so manually (tickling the example implementation to generate lowercase names, verifying that the assertion catches it).

(Always looking forward, I'd love to see a more telling test signal than...

```
RobotTest > hasName FAILED
    java.lang.AssertionError at RobotTest.java:33
```

... an aspect of making Exercism an exemplar would be showing how test output makes plain the exact failure (an important aspect of "good" test systems)...

```
Generated robot name was invalid; it did not match the pattern [A-Z]{2}[0-9]{3}.
```

...)

kytrinyx:
Reading error messages is one of those habits that I've struggled to impart to newer developers. I would love it if exercism helped encourage better error-reading skills (since that seems like it's a pre-requisite for improved error-writing skills...)

PR #60, opened 2015-08-19
Corrected error in RandonKeyCipherTest.java

JTMaher2:
I changed the cipherKeysAreRandomlyGenerated test so that it does not compare the new Cipher to the length of the old Cipher.


kytrinyx:
D'oh! That's a great catch, thank you :)

PR #62, opened 2015-09-08
atbash: Add test case and sample implementation for decode.

jmluy:
Based on the README file, it should have a decode equivalent.


jtigger:
Thank you for taking the time to do this, John: this is a pretty significant gap you closed, here.  I appreciate your use of the Enclosed runner to keep the two cases completely separate.

Cheers!

jmluy:
Glad to help with the java exercises.

PR #63, opened 2015-09-10
allergies: Add test case to check when not allergic to anything

jmluy:
Add test case for also checking isAllergicTo() when not allergic to anything.


jtigger:
Thank you!  Yup, this is a fine way to kick off this exercise.

Cheers!

Issue #55, opened 2015-07-01
crypto-square — tests incorrectly specify how chunking of normalized cipher text works.


jtigger:
*(originally exercism/exercism.io#2432)*

jtigger:
@kytrinyx — moved this issue to the Java track; feel free to close exercism/exercism.io#2432.

jtigger:
@matthewmorgan — thanks for the worked out tests!

matthewmorgan:
No worries, happy to contribute a little.

jtigger:
@kytrinyx I notice there is no <problem>.json for crypto-square.  Seems like the intent behind these files is to prevent this kind of deviation.  True?

If so, I don't mind pulling together that data for this problem and reworking the Java version of the problem to use it.  If we did the same for the JavaScript and Ruby tracks, we've got a seed of goodness in this corner of the garden. :house_with_garden:

kytrinyx:
True! I was just thinking about that. I'll write up a generator for the Ruby and the Go tracks.

matthewmorgan:
@kytrinyx I think we're going to need you to comment on this one. @jtigger and I had another look at this, and it appears #64 made the tests conform to what the majority of the tracks are doing.  But, what the majority of the tracks are doing is wrong:

Chunk sizes are wrong: The README gives several examples of how to deal with strings of text that can be made into perfect ( ie n X n ) squares, and those that can be made into imperfect ( r X c ) squares.  In all the README examples, the imperfect squares are r X c with the additional constraint that r >= c.

As documented by @jtigger [here](https://github.com/exercism/exercism.io/issues/2432#issuecomment-118467422) most of the tracks take a 26-letter string and chunk it in to 5, 5, 4, 4, 4, 4.  This is not the behavior specified in the README, and seems like a error from early on that got propagated to many tracks.

There's no `json` file for this in x-common, but it appears this needs to be changed in many places, not just on the Java track.  @jtigger and I are happy to change it here, but what do you think about putting an issue in to x-common?

kytrinyx:
Yeah, we need to discuss this in x-common, and make sure everyone agrees on the format, and each track can fix their own issue. I also have a new tool that lets me submit identical issues to all the relevant track repositories.

Would you mind creating the x-common issue?

matthewmorgan:
[...]

PR #64, opened 2015-09-16
crypto-square: Fix invalid tests

jmluy:
The existing tests were normalizing the strings based on the square size and not the actual sequence of characters from the columns.
You can refer to the source material where "If man was meant to stay on the 
ground god would have given us roots" should be "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau" but were grouped by square size on the existing tests. This is a temporary fix until #55 uses the json file.


jtigger:
Thanks again, John!  Looks great.

PR #67, opened 2015-09-27
NucleotideTest Guidance about Return Types

masters3d:
https://github.com/exercism/xjava/issues/66
https://github.com/exercism/xjava/issues/65

### DIFF
--- a/nucleotide-count/src/test/java/NucleotideTest.java
+++ b/nucleotide-count/src/test/java/NucleotideTest.java
@@ -10,6 +10,9 @@ public class NucleotideTest {
     assertThat(dna.count('A')).isEqualTo(0);
   }
 
+// nucleotideCounts() <Character, Integer> 
+// Remember "String" 'Character'
+
   @Test
   public void testEmptyDnaStringHasNoNucleotides() {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Aaaah, nice! Leaving for @jtigger to review and merge. Thanks @masters3d

jtigger:
Looks like you resolved the confusion, @masters3d.

Because we're in a practice environment, I'd tend to make such clarifications when we see a pattern of folks getting stuck.  We'll keep an eye on it.

In the event we would make such a clarification (i.e. providing additional context around a language feature), I'd err on being more verbose: if someone is getting help from it, they'll likely lack the experience to interpret a clue.

If you still believe your PR should be included, be more thorough in your description and be prepared for the PR to sit a while as we wait for a pattern to emerge.

Thanks for taking the time to submit!

masters3d:
Interesting. I always wondered what users would do If they get stuck. To me if the API is asking me for a Key, I'd always assume is a String. I come from languages that don't care about single or double quotes.

You are right about patterns though. I think I'll do another PR later with a new type signature after I do the rest of the track!

Thanks for all the excersises.

PR #68, opened 2015-12-02
Add ABOUT.md documentation to describe the Java language.

jjstopforth:
I wrote a description for the Java language as specified on the language page [here](http://exercism.io/languages/java). The file basically contains: 

># OOPs it's Java
>From web applications to games, the Java language is as diverse as it is portable meaning that it is     found nearly everywhere. With a very strong focus on object oriented programming, it is an ideal language in which to implement complex data structures and algorithms.

As this is really my first time contributing to exercism I tried to follow the [contribution instructions](http://help.exercism.io/git-workflow.html) as closely as possible, but I am not entirely sure if what I did was correct. Any feedback or suggestions would be greatly appreciated.

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,0 +1,2 @@
+#OOPs it's Java
+From web applications to games, the Java language is as diverse as it is portable meaning that it is found nearly everywhere. With a very strong focus on object oriented programming, it is an ideal language in which to implement complex data structures and algorithms. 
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks great!

A few things -- it's going to show up on a page that already has a header, so we don't need the first line.

I really like the conversational tone. What do you think of adding some examples about where it's used.

This says: _it is found nearly everywhere_. Could we add some examples e.g. Games like Minecraft, and big corporations like banks and insurance companies--or something like that.

jjstopforth:
I have updated the file with some examples and also removed the header.
I think the examples (below) gave the best scope in terms of "popular" usage of Java without being too obscure.

>Some notable examples include Minecraft, Android apps, and large parts of websites like Amazon.

I am trying to keep the description short and sweet, but do let me know if you want it fleshed out; I will happily write more.

kytrinyx:
This is great! Let's deploy it and see if other people have any further suggestions once it's in the wild. Thanks so much.

PR #69, opened 2015-12-04
Add hello-world exercise

winterchord:
This implementation does not generate a test suite based on the json
meta the hello-world exercise.
See https://github.com/exercism/todo/issues/13

Fixes https://github.com/exercism/todo/issues/35


config.json
```diff
@@ -34,7 +34,8 @@
     "octal",
     "luhn",
     "pig-latin",
-    "pascals-triangle"
+    "pascals-triangle",
+    "hello-world"
```
kytrinyx: Hello world should be the very first exercise in this list.

kytrinyx:
This looks good to me--just the one thing about the ordering of the exercises.

@jtigger would you have a moment to look at this?

config.json
winterchord: You got it. Do you prefer that I squash the fix, or just add another commit?

winterchord:
I originally had more tests (trailing space after name, `"    "` as "empty" string), but removed them. While I think they're correct things to handle in general. But if this exercise is supposed to serve beginners, I thought it best to not include them (don't have to learn about `String#trim()` on your first exercise).

kytrinyx:
Yeah, I agree. This exercise is about making sure that all of the exercism pieces are wired together and that all the dependencies are installed, so it makes sense to keep it very, very simple.

jtigger:
@kytrinyx sorry for the delay; will look over tomorrow morning.

jtigger:
Thank you, Brian! Apologies for not getting to your PR, sooner. Looks great, sir.  I absolutely agree with all the edits you made, too.

PR #70, opened 2015-12-13
Add new exercise pangram

matthewmorgan:
As described in [xcommon #133](https://github.com/exercism/x-common/pull/133).

This is my first PR to the Java track-- I hope I have everything set up right!


pangram/src/test/java/PangramTest.java
```diff
@@ -0,0 +1,38 @@
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+public class PangramTest {
+
+    @Test
+    public void emptySentence() {
+        Pangram pangram = new Pangram("");
```
sit: Not sure what the other tracks look like, but I think it's odd to return an object of type `Pangram` that is .. not a pangram.

In terms of Java, I'd recommend instead something like
```java
@Test(expected = IllegalArgumentException.class)
public void itRejectsEmptySentence() {
    Pangram p = Pangram.of("");
}
```
Or if you want to stick with an `isPangram` method, I'd stick with a static method on a `Pangrams` helper class.
```java
@Test
public void itRejectsEmptySentence() {
{
    assertFalse(Pangrams.isPangram(""));
}
```

sit:
Nice! I'm not an active maintainer so please don't hold up on any of my comments, just making some suggestions on API design.

matthewmorgan:
@sit thanks for your feedback-- I think you make an excellent point.

As far as 'other tracks', this is a new exercise that has only been implemented three other times (and two of those were by me!)  My implementations on the EcmaScript track and the JavaScript track were very similar to this one, and thus contain some of the same choices as far as design.

My intention was to do something more like your second suggestion, especially since it seems like testing for a Pangram is really the point, not creating Pangram objects. 

The Python implementation is a more pure version of what I was trying to do, and very similar to what you are recommending it seems.

jtigger:
thanks for the contribution, @matthewmorgan !

I'll also give it a look over in the morning.  I'm pretty familiar with the setup.  It ought to be that if the build succeeds, you're good to go.

I did notice that your change set includes removing a starter solution from the word-count exercise.  Could we do this in a separate commit, please?

Thanks!

matthewmorgan:
@jtigger The word count change was a mistake on my part.  I was seeing a complaint about a duplicate class definition and thought I had somehow changed the word-count directory.

jtigger:
Thanks again, Mr. Morgan — looks great!  You've been a regular contributor to this community, thank you so much for all your efforts.

Great adds, @sit.  For a "not active maintainer" that's a lot of value add, right there. ;)

jtigger:
Gotcha.

Issue #56, opened 2015-07-03
WordCount — method required by test is sentence-cased.


jtigger:
Instead of

```java
actualWordCount = wordCount.Phrase("word");
```

it should be

```java
actualWordCount = wordCount.phrase("word");
```

In Java, method names are always camelCased (i.e. the first letter is never capitalized).

matthewmorgan:
Closed by #76

PR #76, opened 2015-12-17
Fix example and test to use proper casing for method names

matthewmorgan:
As described in #56


jtigger:
Thanks, @matthewmorgan!

PR #85, opened 2016-01-03
Add linked-list exercise

winterchord:
The tests assume that the Deque class is implemented with generics.

The example implementation keeps track of a circular head Element.

Completes one task in https://github.com/exercism/xjava/issues/3.

### DIFF
--- a/linked-list/src/example/java/Deque.java
+++ b/linked-list/src/example/java/Deque.java
@@ -1,0 +1,45 @@
+public final class Deque<T> {
+    private Element head;
+
+    public void push(T value) {
+        if (head == null) {
+            head = new Element(value, null, null);
+            head.setNext(head);
+            head.setPrev(head);
+            return;
+        }
+
+        Element oldTail = head.getPrev();
+        Element tail = new Element(value, oldTail, head);
+        oldTail.setNext(tail);
+        head.setPrev(tail);
+    }
+
+    public T pop() {
+        head = head.getPrev();
+        return shift();
+    }
+
+    public void unshift(T value) {
+        push(value);
+        head = head.getPrev();
+    }
+
+    public T shift() {
+        Object value = head.getValue();
+
+        Element newHead = head.getNext();
+        Element newTail = head.getPrev();
+
+        if (newHead == head) {
+            head = null;
+        }
+        else {
+            newHead.setPrev(newTail);
+            newTail.setNext(newHead);
+            head = newHead;
+        }
+
+        return (T) value;
+    }
+}
--- a/linked-list/src/example/java/Element.java
+++ b/linked-list/src/example/java/Element.java
@@ -1,0 +1,52 @@
+public final class Element {
+    private final Object value;
+    private Element prev;
+    private Element next;
+
+    public Element(Object value, Element prev, Element next) {
+        this.value = value;
+        this.prev = prev;
+        this.next = next;
+    }
+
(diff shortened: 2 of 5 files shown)


jtigger:
Thank you!  Looks really good, @winterchord.  Thanks for following suit and using AssertJ!  Also, this is a great exercise to introduce generics; nice touch.

I'm not sure how strict we want to be with the progression.  This should definitely be later in the track as you've placed it.  I do notice that in the check list (#3), it's *after* Pascal's Triangle.

winterchord:
@jtigger Yeah, I wasn't sure exactly where to place it in the progression. I tried to place it relative to where it was in the ruby progression. I didn't even think about using the checklist in #3 as the progression order.

linked-list/src/example/java/Deque.java
```diff
@@ -0,0 +1,45 @@
...
+        if (newHead == head) {
+            head = null;
+        }
+        else {
+            newHead.setPrev(newTail);
+            newTail.setNext(newHead);
+            head = newHead;
+        }
+
+        return (T) value;
```
sit: `Element` should be generic as well and instead of storing an `Object` should store an object of type `T`. Then this cast becomes unnecessary.

linked-list/src/example/java/Element.java
```diff
@@ -0,0 +1,52 @@
+public final class Element {
```
sit: This class should be package private. Or just be a private static inner class of `Deque`.

linked-list/src/example/java/Element.java
```diff
@@ -0,0 +1,52 @@
...
+    public void setNext(Element next) {
+        this.next = next;
+    }
+
+    public Object getValue() {
+        return value;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
sit: Why do we need these methods?

sit:
I don't know what the other tracks look like, but there should be tests for popping and shifting off an empty `Deque`.

linked-list/src/example/java/Element.java
winterchord: Since the java implementation (heavily inspired by the ruby implementation) creates a circular list, we need the next-setters during `shift()`. The value-getter is for shifting and popping.

Though, you're right, we don't need the `equals()` or `hashCode()` implementations. We compare pointers when checking for circular references not the values.

linked-list/src/example/java/Element.java
winterchord: Will make it package private.

linked-list/src/example/java/Deque.java
winterchord: You're absolutely right. I had that in a previous implementation. Don't know why I got rid of it.

winterchord:
The instructions for the linked-list exercise (https://github.com/exercism/x-common/blob/master/linked-list.md) state the following:

> To keep your implementation simple, the tests will not cover error conditions. Specifically: pop or shift will never be called on an empty Deque.

winterchord:
I'll create another PR for with the suggested changes and cc you @sit.

linked-list/src/example/java/Element.java
winterchord: You also argue that we unwrap the next/prev fields into public-scoped fields and use them directly since `Element` has tighter scoping. I'm don't have a strong opinion either way.

PR #88, opened 2016-01-20
Fix broken image on installation instructions

michaelkunc:
Also updated installation instructions to reflect latest version of Java.


kytrinyx:
Ah, thanks! We changed how the images are served, and clearly missed a spot.

The new link needs to reference the full url to the API: `http://x.exercism.io/v3/tracks/java/docs/img/mac-osx--install-java-dialog.png`

jtigger:
Thanks @michaelkunc for the update!

PR #92, opened 2016-01-30
Add new exercise acronym

matthewmorgan:
A new exercise!


jtigger:
@matthewmorgan strikes again!  Well-crafted (short, telling test names and a succinct solution) and complete (easy to miss the `config.json` or add it to the `gradle.settings`).  Thank you, sir.

PR #93, opened 2016-01-31
Fix error on edge case for triangle

matthewmorgan:
As described in #91

### DIFF
--- a/config.json
+++ b/config.json
@@ -46,7 +46,9 @@
   "ignored": [
     "_template",
     ".gradle",
-    "docs"
+    "docs",
+    ".idea",
+    "gradle"
   ],
   "foregone": [
     "leap",
--- a/exercises/triangle/src/example/java/Triangle.java
+++ b/exercises/triangle/src/example/java/Triangle.java
@@ -40,7 +40,7 @@ public class Triangle {
     }
 
     private boolean violatesTriangleInequality() {
-        return side1 + side2 <= side3 || side1 + side3 <= side2 || side2 + side3 <= side1;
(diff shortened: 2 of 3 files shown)


kytrinyx:
I may have jumped the gun on the whole triangle inequality thing. First off: I'm technically correct. On the other hand, a 2, 2, 4 triangle is considered a degenerate triangle. Third of all, should we really get into the nitty gritty details of geometry here? Probably not.

See discussion here: https://github.com/exercism/xerlang/issues/85

matthewmorgan:
OK, should I change this PR to revert the code changes and just edit the README?

And change the JS/ES tracks back as well?

kytrinyx:
Yah, probably. Let's actually wait for a week and see if the discussion settles, then I'll do a summary and see if we've landed on some form of agreement, and then we can clean everything up.

matthewmorgan:
@kytrinyx looks like this is settled?  I'm going to close this PR without merging it.

kytrinyx:
Yeah, that sounds good.

svnk:
Just ran into this issue. Test and documentation do contradict each other. Maybe add a sentence to the HINT Section like "Even if traingles where the the sum of the lengths of two sides are equal to the third technically ARE triangles, they are "degenerate" triangles and should raise errors".

kytrinyx:
@svnk thanks, this has popped up again in a few tracks. I don't think we got it straightened out quite well enough.

I also don't know if all of the tracks are handling this the same way. If they are, then we can update the main description over in exercism/x-common. If not, then perhaps updating the HINTS here is the right way to go.

Issue #98, opened 2016-02-29
Add new exercise simple linked list


matthewmorgan:
I have been working on this one and will put the PR as soon as it's fully tested.

PR #99, opened 2016-03-01
Add new exercise simple linked list

matthewmorgan:
As described in #98


sit:
Why both this one and #85?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+/**
+ *
+ * @author matt
```
sit: You can teach your IDE not to generate these comments. They probably aren't relevant for this repo.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
...
+
+        Element(int value) {
+            this.value = value;
+        }
+    }
+
+    private Element head;
+    private int size = 0;
+
+    public SimpleLinkedList() {
```
sit: Java will provide a [default constructor](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.8.9), so this isn't necessary.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
...
+        int value;
+        Element next;
+
+        Element(int value) {
+            this.value = value;
+        }
+    }
+
+    private Element head;
+    private int size = 0;
```
sit: As `0` is already the [default value for an int](https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5), this initializer is unnecessary.

exercises/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -0,0 +1,70 @@
+import java.util.NoSuchElementException;
+import static org.assertj.core.api.Assertions.assertThat;
+import org.junit.Test;
+
+public class SimpleLinkedListTest {
+
+    @Test
+    public void canCreateANewList() {
```
sit: Perhaps a stronger initial test is that `aNewListIsEmpty()` and then `assertThat(list.size()).isEqualTo(0)`.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        head = previous;
+    }
+
+    final public int[] asArray() {
+        int[] result = new int[size];
+        int index = 0;
+        Element current = head;
+        while (this.size() > 0) {
+            result[index++] = this.pop();
```
sit: Should `asArray` mutate the list in this way?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
...
+    public SimpleLinkedList() {
+    }
+
+    public SimpleLinkedList(int[] values) {
+        for (int ii = values.length - 1; ii >= 0; ii--) {
+            push(values[ii]);
+        }
+    }
+
+    final public void push(int value) {
```
sit: Why `final`?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private Element head;
+    private int size = 0;
+
+    public SimpleLinkedList() {
+    }
+
+    public SimpleLinkedList(int[] values) {
+        for (int ii = values.length - 1; ii >= 0; ii--) {
```
sit: It's interesting that this is backwards (in that this linked list is implemented as a stack).

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: no good reason!

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: no.

exercises/simple-linked-list/src/test/java/SimpleLinkedListTest.java
matthewmorgan: good idea.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: good point.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: true!  thanks.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Yes, for some reason my IDE was complaining about an uninitialized value when the property was dereferenced later.  I changed it and never looked back at it later.  I will have another look.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Yes, the Java `LinkedList` class provides `push` and `pop` methods that both operate on the head (LIFO), so that's what I was emulating, but if it should be different please let me know.

matthewmorgan:
@sit "linked-list" is supposed to be doubly linked.  This exercise is singly linked.  Both are specified in x-common.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: I went back and checked this out, and turns out the issue with an uninitialized int was later, in the `isArray` method.  I saw it there and changed it everywhere.  Still don't know why it would be a problem when the int is used in an index.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: The other reason I did it this way is that this is how the Ruby track implements it.  There is no test metadata for this exercise, so I used the Ruby `example.rb` as a template.

matthewmorgan:
Refactored to allow generics.  Feedback appreciated!

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+        int index = 0;
+        Element<T> current = head;
+        while (Objects.nonNull(current)) {
+            result[index++] = current.value;
+            current = current.next;
+        }
+        return (T[]) result;
+    }
+
+    private <T> T[] getArray(Class<T> clazz, int size) {
```
sit: `newArray`?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+
+        Element(T value) {
+            this.value = value;
+        }
+    }
+
+    private Element<T> head;
+    private int size;
+
+    public SimpleLinkedList() {
```
sit: Can be omitted.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
+
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+public class SimpleLinkedList<T> {
+
+    public class Element<T> {
```
sit: Can be `private static` as this is a purely internal class.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+
+        T value;
```
sit: Could be `final`.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public T[] asArray(Class<T> clazz) {
+        T[] result = getArray(clazz, this.size);
+        int index = 0;
+        Element<T> current = head;
+        while (Objects.nonNull(current)) {
+            result[index++] = current.value;
+            current = current.next;
+        }
+        return (T[]) result;
```
sit: Cast should be unnecessary.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+    public SimpleLinkedList() {
+    }
+
+    public SimpleLinkedList(T[] values) {
+        for (int ii = values.length - 1; ii >= 0; ii--) {
+            push(values[ii]);
+        }
+    }
+
+    public void push(T value) {
```
sit: This allows null values, though that may be fine.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
```diff
@@ -0,0 +1,86 @@
...
+            throw new NoSuchElementException();
+        }
+        T value = head.value;
+        head = head.next;
+        this.size--;
+        return value;
+    }
+
+    public void reverse() {
+        Element current = head;
```
sit: Should perhaps be `Element<T>`?

matthewmorgan: Yes.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Unless you know another way, this is required by the test framework.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Are you suggesting changing the method name?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Great!

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: thanks!

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: Yes, it seemed to be earlier, but it must have been something I hadn't set up correctly with the types.

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
jtigger: The warning about calling non-final methods in constructors is conservative.  It's an attempt to enforce the [Open-Closed Principle](https://en.wikipedia.org/wiki/Open/closed_principle).

Specifically, if a method that is being invoked in a constructor is non-final, it technically can be overridden in a subclass.  If someone created a subclass and overrode that method, then the base class' constructor would invoke that overriding implementation (example at: https://gist.github.com/jtigger/61ce1c48543a78846141).

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
sit: Does `new SimpleLInkedList()` not work if you don't have this?

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
sit: Yeah, to indicate that it's constructing something (as opposed to getting something).

exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
matthewmorgan: No, the test class fails to compile.

matthewmorgan:
@sit @jtigger is this OK to merge?  Any remaining concerns?  Thanks

sit:
LGTM.

kytrinyx:
Thanks!

jtigger:
btw, @sit, we were singing your praises, this morning.  Great spot
feedback/questions!  Thank you, sir!  Hoping to help carry forward that
which you started.

PR #106, opened 2016-04-01
Initial POC of how to ignore all-but-first test

### LINKED ISSUE #101: Add @Ignore to track unit tests other than first one
In the JS and ES tracks, all tests except the first for any given problem are disabled by default.  This way the user is encouraged to make their solution pass one test at a time, enabling the next test as he or she proceeds.

In JUnit we could use `@Ignore` in front of the `@Test` annotation to enable this.

There's a second step, how do we tell gradle to ignore `@Ignore` so all the tests get run during development and CI builds?

sit:
Introduce a new JUnit test category that lets us mark certain tests as
not ready for execution. The test task in the individual project
can then be configured to not run tests marked as `NotReady`.

For the CI server builds, add a fullTest test task that simply runs
tests as normal, and link it into the check task so it runs on builds.

This is not perfect, as the class under test must still compile, so it
must have at least the correct signatures of all the methods tested.
Also, it may be worth finding a way to distribute the NotReady interface
that doesn't require checking it into every project.

This will fix #101.

### DIFF
--- a/exercises/build.gradle
+++ b/exercises/build.gradle
@@ -19,4 +19,11 @@ subprojects { project ->
       println "  (source = " + compileTask.source.asPath + ")"
     }
   }
+
+  task fullTest(type: Test) {
+    group = LifecycleBasePlugin.VERIFICATION_GROUP
+    description = "Runs tests without exclusions."
+  }
+
+  check.dependsOn(fullTest)
 }
--- a/exercises/grade-school/build.gradle
+++ b/exercises/grade-school/build.gradle
@@ -11,3 +11,8 @@ dependencies {
   testCompile "org.assertj:assertj-core:3.2.0"
 }
 
+test {
+    useJUnit {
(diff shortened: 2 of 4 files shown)


jtigger:
Having an example to talk to is very helpful, @sit; thanks!

Chewing... :)

jtigger:
@sit do you have time to flesh this out?  Otherwise, either @matthewmorgan or myself could do the grunt work.

jtigger:
@matthewmorgan and I are picking this one up.  We agreed that @sit's approach is the way forward, here.  Thanks!

jtigger:
Yikes!  So sorry for squatting on this issue. :(  Really want to see this implemented.

Recently, @matthewmorgan and I have been focusing on some amateur data science'ing, but perhaps we'll clear out some of these PRs and issues too.

jtigger:
Starting with Emil's (@sit) work.  I'm hoping this will preserve history/blame/credit.

PR #114, opened 2016-05-11
Fix typo in test name

dguardado:
As I don't think the month of Map exists, I'm guessing this is meant to be May ;)


kytrinyx:
That's a pretty good guess :heart:

PR #120, opened 2016-07-08
CI runs journey test

bronhuston:
implements exercism/xjava#83.


matthewmorgan:
This looks great!

After reviewing this with @jtigger, we are thinking one possible improvement (for the future, if not now) would be to combine `run-journey-test-in-ci.sh` and `run-journey-test-locally.sh` in to one script.

Our thought is that all the environment cleanup in the local script will be harmless in the CI container, so we can leave that.  Then, we just need to account for different ruby managers and act accordingly.

But, having said that, this is an awesome improvement to the contributor experience.  The refactor can be for a future PR, and either @bronhuston, @NeoValkyrion,  and Cen Ge can take it on or @jtigger and I would be happy to do it.

PR #121, opened 2016-07-10
Merge journey test scripts into one.

jtigger:


jtigger:
This contributes to exercism/xjava#83

Issue #116, opened 2016-05-25
Implement Java 8's DateTime Class in Meetup Exercise


matthewmorgan:
Java 8 provides a much nicer way to manipulate dates and times than past version of Java did.

This means we no longer need to use the JodaTime library, as many of its good ideas now inform the syntax of the Java 8 DateTime and LocalDateTime classes.

Let's switch the Meetup test and example problem to the standard libraries and remove the Joda dependency.

PR #124, opened 2016-08-04
Meetup: dependencies to org.joda.time removed in favor of java.time API.

samilaine:
A one proposition to handle [#116](https://github.com/exercism/xjava/issues/116).

### DIFF
--- a/exercises/meetup/src/test/java/MeetupTest.java
+++ b/exercises/meetup/src/test/java/MeetupTest.java
@@ -1,647 +1,643 @@
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeConstants;
+import java.time.LocalDateTime;
+import static java.time.DayOfWeek.*;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.joda.time.DateTimeConstants.*;
 
-/*
- * We use Joda Time here to encourage the use of a saner date manipulation library.
- */
 public class MeetupTest {
     @Test
     public void test_monteenth_of_may_2013() {
-        DateTime expected = new DateTime(2013, 5, 13, 0, 0);
+        LocalDateTime expected = LocalDateTime.of(2013, 5, 13, 0, 0);
         Meetup meetup = new Meetup(5, 2013);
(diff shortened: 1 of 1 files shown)


samilaine:
I'd rather not paste a working example right here to make it easier to pass configuous-integration tests. How should I try to deliver it?

matthewmorgan:
HI @samilaine thanks for doing this.  I think the issue with the CI is that the example at `meetup/src/example/java` does not use the Java 8 LocalDateTime classes.  It's still using JodaTime, so the tests won't pass.

Do you want to refactor the example as well?  I would be happy to help with that if you like.  LMK.

samilaine:
The tests in `MeetupTest.java` have been refactored as well. However, because the original tests used `org.yoda.time` enums, they will fail even though the tests have been refactored.

You see, the original version of the exercise had `public DateTime day(int day, MeetupSchedule schedule)`. However, when using `java.util.time` the method signature is different: `public LocalDateTime day(int d, MeetupSchedule m)`. I believe this is why continuous integration tests fail: It's not that the tests themselves area faulty, but the return type is different and hence the tests for continuous integration would have to be fiddled with.

This means that all the previous solutions for this exercise will fail in the future, should the pull request go through. I'm not sure if that's a good idea. Maybe a better idea would be to create a new exercise which uses `java.util.time` interfaces/classes, and leave the old meetup exercise as it is?

jtigger:
I really appreciate your thinking about how changing the exercise will have an impact on folks who have done the exercise in the past, @samilaine.

I sense that there have been (many?) conversations about this in the Exercism community.  Anyone aware of if/when/how those went?

My initial thought is that if the change is a clear improvement, we should evolve the exercises to improve the overall quality of the experience.

Opinions?

jtigger:
https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#exercise-versioning

matthewmorgan:
@jtigger @samilaine for my part I feel that moving forward is they way to go.  Especially given the explicit support for and requirement of Java 8 in this track.

My vote: update all the things!

matthewmorgan:
I'm going to close this PR due to lack of activity for the last several weeks.

PR #127, opened 2016-08-23
Fixed typo "but the tests are expects"

rcraggs:
Fixed a typo in the getting started guide

but the tests are expects -> but the tests expect


kytrinyx:
Great catch, thank you!

Issue #138, opened 2016-09-18
Add @Ignore to "hello-world" unit tests other than first one


jtigger:
First increment of #101

jtigger:
Resolved by #139

PR #139, opened 2016-09-24
Ignore all but first test in hello-world

jtigger:
Started with @sit's suggestion, but then tacked to use `@Ignore` and defined a copy with filter in the root `build.gradle` that removes these annotations in a maintainers' build.


jtigger:
This addresses #138 and #140.

scripts/increase-test-logging.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
sit: What are these scripts for? Just for developer use?

jtigger: Yes.  They are not at all generalized, but will be.  In the meantime, a starting point to build other scripts?

jtigger:
Thanks for the review @sit!  Yeah, I thought that the added complexity in the maintainer build (which you only have to understand if you need to modify it, at this point) was worth the trade-off of removing the complexity for the exercism programmer.

There are additional benefits.  By turning up test logging, the ignored tests are mentioned, signalling to the programmer there's more to do.

matthewmorgan:
👍 Thanks @jtigger !

PR #146, opened 2016-10-01
Add missing tests to Pangram

jtigger:
- ensure titles of tests match those in
  https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json


jtigger:
Addresses #117.

PR #153, opened 2016-10-16
Replace JodaTime in Meetup with standard lib

matthewmorgan:
As of Java 8, the DateTime libraries have become more usable and less bizarre.

This PR replaces the Joda dependency from the meetup exercise with references to the standard library, as discussed in #116.


jtigger:
Thanks, @matthewmorgan !

PR #155, opened 2016-10-26
Adding leap for java

hajush:
The jump from hello-world to Etl was very jarring to new Java programmers. Hope adding the "leap" that's used in other languages like JavaScript will help make the path easier.

### DIFF
--- a/exercises/leap/build.gradle
+++ b/exercises/leap/build.gradle
@@ -1,0 +1,18 @@
+apply plugin: "java"
+apply plugin: "eclipse"
+apply plugin: "idea"
+
+repositories {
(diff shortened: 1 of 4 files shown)


jtigger:
@hajush, we really appreciate your efforts.  However, "leap" cannot be implemented on the Java track.  Please see #3 for details (and a list of exercises that *can* be implemented!).

This is a failure on our part: it is not clear what additional exercises could be implemented; that should be front-and-center.  Apologies for not making that clearer.  We're a bit under construction and the instructions on how to contribute and exercise has been shuffled. :-/

It's my bedtime; I'll tackle this in the morning.

PR #156, opened 2016-10-28
PPA for gradle

mienaikage:
I spent way too long trying to figure out why gradle wasn't working properly. Ended up with gradle 1.4 which doesn't seem to play nice with java 8.


sit:
Let's not do this. Instead we should upgrade everything to Gradle 3.x. @exercism/java

jtigger:
@sit, given all we've learned, are you okay with this tack?

sit:
Go for it.

Issue #168, opened 2016-11-05
etl: usage of 3rd party libraries in exercism.io


alebaffa:
Hi,
in the [tests of the Etl exercise](https://github.com/exercism/xjava/blob/master/exercises/etl/src/test/java/EtlTest.java) a 3rd party library (Guava) is used to create the [ImmutableMap](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/ImmutableMap.html) objects. 
The problem with this library is that the ImmutableMap class cannot handle maps with more than 5 elements, and so the tests cases are written differently. See these two example from two test cases in this exercise:

```
import com.google.common.collect.ImmutableMap; //3rd party library

// from the test case testTransformMoreValues()
Map<Integer, List<String>> old = ImmutableMap.of(
                1, Arrays.asList("A", "E"),
                2, Arrays.asList("D", "G")
        );

// from the test case testFullDataset()
Map<Integer, List<String>> old = ImmutableMap.<Integer, List<String>>builder().
                put(1, Arrays.asList("A", "E", "I", "O", "U", "L", "N", "R", "S", "T")).
                put(2, Arrays.asList("D", "G")).
                put(3, Arrays.asList("B", "C", "M", "P")).
                put(4, Arrays.asList("F", "H", "V", "W", "Y")).
                put(5, Arrays.asList("K")).
                put(8, Arrays.asList("J", "X")).
                put(10, Arrays.asList("Q", "Z")).
                build();
``` 
As you can see, the two test cases are written differently since the first one creates a map with 2 elements, while the second one a map with more than 5.

In my opinion there are two problems with these tests:
1. they are incoherent and they make it hard to clearly understand the difference among the different test cases. 
2. These test cases could have been written also using the standard library and achieving the same result. Here below the second test cases previously mentioned, but rewritten :
``` 
Map<Integer, List<String>> old = new HashMap<Integer, List<String>>() {
            {
                put(1, Arrays.asList("A", "E", "I", "O", "U", "L", "N", "R", "S", "T"));
                put(2, Arrays.asList("D", "G"));
                put(3, Arrays.asList("B", "C", "M", "P"));
                put(4, Arrays.asList("F", "H", "V", "W", "Y"));
                put(5, Arrays.asList("K"));
                put(8, Arrays.asList("J", "X"));
                put(10, Arrays.asList("Q", "Z"));
            }
        };
        Collections.unmodifiableMap(old);
``` 

[...]

PR #170, opened 2016-11-07
Remove Guava and use JUnit assert

alebaffa:
As described in issue #168, I removed the dependency from Guava library in order to use `Collections.unmodifiableMap` from the standard library. The last test case is longer than before, but at least now all the test cases are written coherently and so they are easier to understand.

Also, I noticed that this exercise was using the Guava assertions `assertJ`. So I removed it and used the standard `assertEquals` from JUnit (though, I could have used the Hamcrest library for the `assertThat`, or the `assertThat` from JUnit + a matcher from Hamcrest. Let me know if you prefer that way for the assertion instead of the standard assertEqual).

### DIFF
--- a/exercises/etl/src/test/java/EtlTest.java
+++ b/exercises/etl/src/test/java/EtlTest.java
@@ -22,7 +22,7 @@ public class EtlTest {
                 put("a", 1);
             }
         };
-        Collections.unmodifiableMap(expected);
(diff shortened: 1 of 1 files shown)


alebaffa:
Done.

alebaffa:
Done. Sorry!

alebaffa:
awesome! Thanks @jtigger

PR #171, opened 2016-11-10
queen-attack: add to track

stkent:
Test data sourced from https://github.com/exercism/x-common/blob/master/exercises/queen-attack/canonical-data.json


stkent:
@jtigger no idea what caused that failure. Any clues? If not, how can I trigger a rebuild? Thanks!

stkent:
Never mind, rebuild seems to be working just fine!

jtigger:
hmmmmmm.  From [the logs](https://travis-ci.org/exercism/xjava/builds/174942720):

```
$ bin/unit-tests.sh
FAILURE: Build failed with an exception.
* What went wrong:
Could not open terminal for stdout: $TERM not set
```

... but like you said, the next build succeeded.

stkent:
Yeah, likely to be a CI hiccup I'm thinking.

exercises/queen-attack/src/example/java/QueenAttackCalculator.java
```diff
@@ -0,0 +1,86 @@
+import java.util.Arrays;
+
+public final class QueenAttackCalculator {
+
+    private final int[] queen1Coordinates;
+
+    private final int[] queen2Coordinates;
+
+    public QueenAttackCalculator(final int[] queen1Coordinates, final int[] queen2Coordinates)
```
jtigger: I know that the README doesn't call it out, but what do you think of naming the parameters `blackQueenCoordinates` and `whiteQueenCoordinates`?

Personally, I find exercises that immerse me in the domain a bit more engaging... as if I'm solving a trimmed down version of "the real thing."

jtigger: Just as an FYI: I'm proposing to the group to switch to SAN: exercism/x-common#446.  That is a significant change and should *not* hold up this work.

stkent: R.E. a `Coordinate` type - I agree that primitive types are a little ugly here, but I avoided introducing a second separate type that requires implementation as I did not find another example which follows that pattern. (This is probably why other exercises used e.g. Guava to provide a nicer parameter type, though I agree with the removal of that third party dependency moving forward). It shouldn't be hard to adapt the problem, however, so I'll make the change to use a separate `Coordinate` type and we can see how that feels!

R.E. renaming to white/black - yep, great call, will make the change!

R.E. notation change - there's a note about using chess notation in the canonical data set that mentions it, but IMO it's not really clear how the note should be interpreted:

> Some languages implement tests beyond this set, such as checking for two pieces being placed on the same position, representing the board graphically, or using standard chess notation. Those tests can be offered as extra credit.

Is there build-in support in Exercism for optional/extra credit tests? If not, I'm guessing this means we could expose two methods that accept differently-formatted `Coordinate` types and require implementations that satisfy both to consider the problem complete. That feels like relatively uninteresting duplication, however.

exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.assertTrue;
+
+public final class QueenAttackCalculatorTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
```
stkent: I like the second option - we can save a search and provide a link to: https://github.com/junit-team/junit4/wiki/Rules, but still leave reading and discovery to the user?

exercises/settings.gradle
```diff
@@ -24,6 +24,7 @@ include 'pangram'
 include 'pascals-triangle'
 include 'phone-number'
 include 'pig-latin'
+include 'queen-attack'
```
stkent: Hah, everything in its right place!

exercises/queen-attack/src/example/java/QueenAttackCalculator.java
jtigger: subject `Coordinates`:
It is absolutely "in bounds" to include other bits in the starter code (i.e. `main` sourceset).  There are examples of this in other exercises ([`Allergen`](https://github.com/exercism/xjava/blob/master/exercises/allergies/src/main/java/Allergen.java) for one).

The way I think about it, we are defining the API for the exercising programmer.  We are free to define as little as an interface and as much as a whole class as we see appropriate.

jtigger: subject "SAN":
There is *not* currently a mechanism for extra-credit, per se.  It's merely a mention of where the exercising programmer could go from here.

Personally, I'd like to see this exercise be defined as using SAN, but that takes rough consensus across the tracks (read: not happening, today).

It could be an interesting exercise to declare two contracts (one with cartesian coordinate and another for SAN coordinates) and a test class for both of those interfaces.  The SAN test class would have to make it clear that it is optional; there ought to be additional readme (`EXTRA-CREDIT.md`?) to provide a proper framing.

None of that is at all required for this PR.  I'm just thinking out loud, now.

stkent: R.E. `Coordinates`: ah, thanks for the reference to the allergies exercise; symlinking into the example source set is a nice solution there. What are your thoughts, then, for the queen-attack exercise: require the implementer to build the `Coordinate` class, or provide it and allow the implementer to focus on the attack logic only?

stkent: R.E. notation: all good points! Perhaps you could open a new issue to capture the ideas you described regarding extra credit, etc.?

exercises/queen-attack/src/main/java/Coordinate.java
```diff
@@ -0,0 +1,5 @@
+public final class Coordinate {
```
jtigger: I suspect this was left over from an intermediate effort?

exercises/queen-attack/src/example/java/BoardCoordinate.java
```diff
@@ -0,0 +1,39 @@
+public final class BoardCoordinate {
+
+    private final int rank;
+
+    private final int file;
+
+    public BoardCoordinate(final int rank, final int file) throws IllegalArgumentException {
```
jtigger: This is not *at all* a request/demand... just a thought: value objects often come with come with a factory method (usually with a short name).  Then, in places where it is used, they employ a static import.

```java
public static BoardCoordinate at(final int rank, final int file) {
   return new BoardCoordinate(rank, file);
}
```

exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java
```diff
@@ -3,111 +3,132 @@
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public final class QueenAttackCalculatorTest {
 
+    /*
```
jtigger: 👍

exercises/queen-attack/src/main/java/Coordinate.java
stkent: Ah, yes, I did mean to leave a `BoardCoordinate` class in here since that's a required part of the exercise API; renamed!

stkent:
@jtigger thanks for the comments - replied/updated!

exercises/queen-attack/src/example/java/BoardCoordinate.java
jtigger: oooohhhh!  I missed the part where `BoardCoordinates` is in the `example` sourceset.  I absolutely agree.

exercises/anagram/src/test/java/AnagramTest.java
```diff
@@ -1,9 +1,10 @@
-import static org.assertj.core.api.Assertions.assertThat;
+import org.junit.Ignore;
+import org.junit.Test;
 
 import java.util.Arrays;
 import java.util.List;
-import org.junit.Test;
-import org.junit.Ignore;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
jtigger: OMG, you're going to kill me — this is a long PR; thank you for your patience.

Looks like we have a freeloader, here.

stkent: The pitfalls of auto-optimized imports! Fixed this commit :)

stkent:
Rebased to fix conflicts after largest-series-product was merged.

jtigger:
Thanks again, @stkent.  Good add.

stkent:
@jtigger I don't see this exercise on the website (http://exercism.io/languages/java/exercises) 4 days after the merge. Is there a periodic rebuild of the website that will expose the new exercise, or did I miss something in implementing this PR? Thanks!

jtigger:
You have done your job, @stkent.

Your work typically gets swept-up in the periodic deploys that @kytrinyx does from time-to-time.  With that mention, she's now explicitly aware of this one. ;)

stkent:
Hah, as I suspected. Thanks for confirming!

PR #174, opened 2016-11-12
minesweeper: add to track

stkent:
Data from https://github.com/exercism/x-common/blob/master/exercises/minesweeper/canonical-data.json, plus some extra tests to validate input.

The long annotate method is kind of ugly, so I'm happy to refactor to non-static methods to make it easier to break apart. Input welcome :)


exercises/minesweeper/src/example/java/Minesweeper.java
```diff
@@ -0,0 +1,89 @@
...
+
+                // Count mines in the squares surrounding (row, col).
+                for (int rowToInspect = minRowToInspect; rowToInspect <= maxRowToInspect; rowToInspect++) {
+                    for (int colToInspect = minColToInspect; colToInspect <= maxColToInspect; colToInspect++) {
+                        if (inputBoard.get(rowToInspect).charAt(colToInspect) == MINE_CHAR) {
+                            mineCount += 1;
+                        }
+                    }
+                }
+
```
matthewmorgan: @stkent this is really the only part of this method that makes me wish for refactoring.  I find myself wondering if there's not some way to grab the entire subset of characters concerned and reduce them to a number without nested `for` loops and `if`s.

Otherwise this PR is generally very nice work.

stkent: Thanks @matthewmorgan! I'll think about a neater implementation for that heavily-nested portion of the code, and probably switch to an object-based implementation per discussion in  https://github.com/exercism/xjava/issues/177.

stkent:
@matthewmorgan updated this implementation; I don't see a neat way around manually looping to check the cells surrounding a target, but I've at least broken the calculation of the annotated board down into smaller chunks now.

matthewmorgan:
@stkent, fair enough.  I asked because we did this exercise over a code kata lunch at work a few weeks back, and the nested loops always provoke discussion!

If I think of anything myself I'll be sure to share it with you.

PR #182, opened 2016-11-21
Javaeeeee

javaeeeee:
Updated config.json to include Series exercise.


exercises/series/src/test/java/SeriesTest.java
```diff
@@ -0,0 +1,144 @@
...
+    public void hasDigitsShort() {
+        Series sut = new Series("01234");
+        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4);
+        List<Integer> actual = sut.getDigits();
+        assertNotNull(actual);
+        assertFalse(actual.isEmpty());
+        assertEquals(expected, actual);
+    }
+
+    @Test
```
jtigger: There's no instruction that would have prompted you to do so, but...

Please `@Ignore` all but the first test.  Doing so makes for a much more pleasant user experience when practitioners run the tests.  Without it, she would get a wall of text from all the failures.

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -0,0 +1,144 @@
...
+import static org.junit.Assert.assertNotNull;
+
+public class SeriesTest {
+
+    @Test
+    public void hasDigitsShort() {
+        Series sut = new Series("01234");
+        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4);
+        List<Integer> actual = sut.getDigits();
+        assertNotNull(actual);
```
jtigger: This is a really nice touch: checking for null, then non-empty, then actual contents.  Test systems like this give great signal.  Thank you, @javaeeeee!

exercises/settings.gradle
```diff
@@ -33,6 +33,7 @@ include 'robot-name'
 include 'roman-numerals'
 include 'prime-factors'
 include 'scrabble-score'
+include 'series'
```
jtigger: +1 for keeping with the alphabetical ordering.  It's little details like this that keep the place tidy and encourage others to do so. :)

javaeeeee:
Hello John,

Thank you for reviewing my request. Forgot about @Ignore, added; Thank you
for reminding.

As to method return types, if one looks at README, substrings should be
returned. But I've taken a look at JavaScript and C# versions and looks
like arrays are returned everywhere.

Also, we return digits from the string in one of the methods in other
languages, so it seemed to me logical to go with arrays of numbers.

Anyways, methods could be easily converted into ones returning arrays of
substrings.

Please inform me about your decision.

Best Regards,

Dmitry

jtigger:
Hey Dmitri!  I dig how you looked at "prior art" and allowed that to inform your design choice.  That said, I tend to come back to the question, "how can I surprise my user/consumer the least?"  Yeah, that's a bit subjective, but not a bad goal to at least aim for.

This is the first I've seen of this exercise.  I find myself wishing that it was a bit more contextualized.  Meaning, I wish I knew *why* I might identify all the substrings of a particular length... knowing that would inform what interface would be most appropriate.

In the absence of that, what do you think of going with the simplest data structure?

javaeeeee:
"*why* I might identify all the substrings of a particular length"

Great question, thank you. The quick and easy answer would be to attain the
so-called unconcious competence (
https://en.wikipedia.org/wiki/Four_stages_of_competence) in the desired
programming language.

Digging deeper and Googling shows that extracting substrings might not be
only an exercise in coding but can actually be used in computational
biology to identify some DNA fragments (
https://books.google.ca/books?id=3Ss-ws2Zm6IC&pg=SA24-PA6&lpg=SA24-PA6&dq=why+I+might+identify+all+the+substrings+of+a+particular+length&source=bl&ots=xUZ_JWyWzb&sig=HHS7DYcnWxiekTnkUiP3WoEE1BM&hl=en&sa=X&ved=0ahUKEwj2ufflir_QAhVh64MKHT5DBdwQ6AEILzAD#v=onepage&q=why%20I%20might%20identify%20all%20the%20substrings%20of%20a%20particular%20length&f=false),
but those are letter strings.

The final result is the hash of the string, which is used to search the
database in the form of a hash table.

Going back to the exercise, the README says that we work with strings of
numbers. This should be offered to try various approaches of splitting
strings into digits. The tests offered in other tracks test the splitting.
Without it, we throw away a part of the exercise.

After that, one can try to work with arrays. Also, it's not difficult to
create a string from an array. Maybe producing strings from arrays is not a
direct route to the goal, but could offer some additional training.

Some exercises ask directly to abandon some constructs like a filter and to
reinvent the wheel in order to improve coding skills.

Personally, I lean towards the opinion that one shouldn't look for some
sense in such exercises, those are not research or production code, just to
concentrate on language constructs.

In addition, tests are also a kind of code documentation, so one can learn
about the importance of reading tests to extract useful information about
the code, method contracts in this particular case. So, some surprise could
be useful to teach that information about code can be extracted not only
from the textual description but also from the tests.

Finally, it's not difficult to change the return type of methods, but then
we'll be forced to abandon the part that produces a digit array from the
number string, also such array can be obtained by using segment length of
one.

jtigger:
Thank you for your thorough thought process, Dimitri.  I looked at the other interpretations as well and see how a big part of the point is that these are literally a series of integers.

...and, you are absolitely right that there will be some exercises that are just exercises, with no specific background.

jtigger:
Thank you for implementing this exercise, @javaeeeee!  Great work, sir!

javaeeeee:
Hello John,

thank you for the interesting discussion and reviewing my code.

In addition to what was previously mentioned, found an exercise that can
build on the Series which is
http://exercism.io/exercises/java/largest-series-product/readme.

So, using arrays, not Strings could be justified.

Thank you.

PR #186, opened 2016-11-25
Binary search

javaeeeee:
Added the binary search exercise to the Java track.


stkent:
@javaeeeee thanks for this addition! Overall the exercise implementation looks very solid; I'll comment inline with a few comments/questions and we can discuss from there :)

exercises/binary-search/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+apply plugin: "eclipse"
+apply plugin: "idea"
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  testCompile "junit:junit:4.12"
+  testCompile "org.assertj:assertj-core:3.2.0"
```
stkent: In general, my preference (moving forward) would be for all new exercise implementations to rely on the default JUnit assertions. While I definitely think AssertJ assertions read better, it's another API surface that newcomers need to explore/understand to make sense of the tests we present.

@exercism/java thoughts on this? Is there somewhere we can capture standards such as the one I'm proposing for future reference? Thanks!

exercises/binary-search/src/example/java/BinarySearch.java
```diff
@@ -0,0 +1,70 @@
+
+import java.util.List;
+
+public class BinarySearch<T extends Comparable<T>> {
+
+    public static String ARRAY_MUST_BE_SORTED = "Array should be sorted.";
```
stkent: This field can be marked `final` to make it an honest-to-goodness constant :)

exercises/binary-search/src/test/java/BinarySearchTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public void shouldRequireASortedListOK() {
+        BinarySearch<Integer> sut = new BinarySearch<>(SORTED_LIST);
+        List<Integer> actual = sut.getArray();
+        assertNotNull(actual);
+        assertFalse(actual.isEmpty());
+        assertEquals(actual.size(), SORTED_LIST.size());
+        assertEquals(actual, SORTED_LIST);
+    }
+
+    @Ignore
```
stkent: Thanks for ignoring all but the first test by default! Implementors thank you.

exercises/binary-search/src/test/java/BinarySearchTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void shouldReturnMinusOneIfNotFound() {
+        BinarySearch<Integer> sut = new BinarySearch<>(SORTED_LIST);
+        final int number = 10;
+        final int actual = sut.indexOf(number);
+        final int expected = -1;
+        assertEquals(expected, actual);
+    }
```
stkent: The `binary-search` problem has associated canonical test data located [here](https://github.com/exercism/x-common/blob/master/exercises/binary-search/canonical-data.json).

In general, when implementing a new exercise in a track we should reference those canonical test cases (if they exist) to ensure consistency across languages. I think you have most of the expected cases covered, but for future maintainability it would be awesome if you could adjust your tests to match the canonical naming/data exactly. Thanks!

exercises/binary-search/src/test/java/BinarySearchTest.java
```diff
@@ -0,0 +1,100 @@
...
+                    Arrays.asList(0, 1, 2, 2, 3, 10, 12)
+            );
+
+    public static final List<Integer> UNSORTED_LIST
+            = Collections.unmodifiableList(
+                    Arrays.asList(10, 2, 5, 1)
+            );
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
stkent: Nice use of Rules! Since this is a relatively advanced concept that we've only just begun integrating into the track, we've been marking all usages with a helpful comment (for an example, see https://github.com/exercism/xjava/blob/master/exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java#L11-L14). Please add that same comment here. Once we have established a progression of exercises per #142, then we'll likely reduce this redundancy some, but better safe than sorry for now.

javaeeeee:
Hello, Stuart and thank you.

exercises/binary-search/build.gradle
jtigger: Agreed, @stkent for the exact reason you gave.

javaeeeee: Hello, also agree. Actually used jUnit assertions in the test. In some other projects AssertJ is used, so It looks like a standard for the project. Removed the dependency.

exercises/binary-search/src/example/java/BinarySearch.java
javaeeeee: Sorry, missed this. Static analyzers had several other complaints including Javadocs, so didn't spot it. Are Javadocs actually allowed in the project?

stkent: No problem! As far as I know, adding Javadocs to example implementations (when appropriate) should be just fine. Pinging @exercism/java for confirmation of that though.

exercises/binary-search/src/test/java/BinarySearchTest.java
javaeeeee: Removed the @Rule from the tests altogether as according to the [canonical test data](https://github.com/exercism/x-common/blob/master/exercises/binary-search/canonical-data.json) one shouldn't test if the array is sorted.

exercises/binary-search/src/example/java/BinarySearch.java
javaeeeee: Hello, Stuart, thank you for the answer. So the takeaway is that comments may be used only in the example implementations, not the tests. Am I right?

stkent: I think comments are just fine in the tests too - as mentioned somewhere else in this PR, we have them when JUnit rules are implemented, so that users can find relevant context more quickly. However, as is generally considered good practice, we definitely don't want comments on every line of code just for the sake of it ;)

exercises/binary-search/src/example/java/BinarySearch.java
```diff
@@ -0,0 +1,40 @@
+
+import java.util.List;
+
+public class BinarySearch<T extends Comparable<T>> {
```
stkent: Just noticed that this generic type means the example implementation is a lot more general than the tests require. @exercism/java is this a good feature in a reference solution, or does it veer towards YAGNI territory?

exercises/binary-search/src/example/java/BinarySearch.java
jtigger: In general, the code should be self-explanatory will little need for comment.

If one is tempted to add a comment, ask first, "is there a way I can reword this code to make it so clear, a comment is no longer necessary?"

Comments are best used to explain unusual/surprising context... that is to say, they share *why* a chunk of code is shaped in the way it is.

For example, in `robot-simulator`, @stkent smartly overloaded `Object::equals` instead of overriding it and didn't implement `Object::hashCode`.  He left a comment explaining why and how one would do it in the real world.

exercises/binary-search/src/example/java/BinarySearch.java
jtigger: Personally, I currently give *wide berth* to implementors in what happens in the example.  As long as it passes the tests and represents a true solution, it's generally been implementor's choice.

I think this way because practitioners don't see these.  The high value conversations in PRs are about how the test suite flows, the API decisions therein and whether we're implementing the exercise correctly.

exercises/binary-search/build.gradle
jtigger: You are right, @javaeeeee, we haven't completely removed "not technically necessary" third party dependencies.  Sorry for the confusion that caused.

exercises/binary-search/src/example/java/BinarySearch.java
javaeeeee: Hello, and thank you for your comments. Here is a question: Do we need *.keep* files? Spotted that some were removed in one of the commits.

stkent: Since there are other files in `exercises/binary-search/src/main/java/` and `exercises/binary-search/src/test/java/`, it should be safe to remove `exercises/binary-search/src/main/java/.keep` and `exercises/binary-search/src/test/java/.keep`. Thanks for checking!

stkent:
One more merge with master needed!

javaeeeee:
Merged!

stkent:
Sorry for the silence yesterday; I should be able to take another look at this tonight :o)

javaeeeee:
It's Okay. Thank you. There is no hurry.

stkent:
Looks good! One more merge, then I'll hit the button as fast as possible so we don't need any more :)

jtigger:
Take your time; I'll make sure not to merge until this one goes through (assuming no corrective feedback).

javaeeeee:
Thank you. Done.

stkent:
@jtigger hah, yes, I wouldn't really race, but thank you for holding off on others. (I haven't dealt with my own merge conflict for that same reason :)

javaeeeee:
Thank you. :)

PR #187, opened 2016-11-25
Added Pythagorean Triplet assignment

javaeeeee:
Added Pythagorean Triplet assignment to the Java path.


javaeeeee:
Config JSON has data that differs from binary-search branch as pythagorean-triplet slug was added instead to make possible to work with two requests independently.

jtigger:
Sorry about all the conflicts, @javaeeeee... been merging in a bunch of adds (which is AwEsOmE!🎈) which all will conflict on `config.json`.

Getting to yours so that the rebasing can end. :)

exercises/pythagorean-triplet/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+apply plugin: "eclipse"
+apply plugin: "idea"
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  testCompile "junit:junit:4.12"
+  testCompile "org.assertj:assertj-core:3.2.0"
```
jtigger: Let's pop this bad-boy (i.e. `assertj`) out as we're aiming to minimize the number of 3rd party dependencies (very recent thing).  ("policy" at exercism/exercism.io#3113).

javaeeeee:
Ok, thank you. Just corrected and once again. :)

exercises/pythagorean-triplet/src/test/java/PythagoreanTripletTest.java
```diff
@@ -0,0 +1,95 @@
...
+        PythagoreanTriplet sut = new PythagoreanTriplet(5, 6, 7);
+        assertFalse(sut.isPythagorean());
+    }
+
+    @Test
+    @Ignore
+    public void shouldMakeTripletsUpToTen() {
+        final Map<String, Integer> searchData
+                = new HashMap<>();
+        searchData.put(PythagoreanTriplet.MAX_FACTOR, 10);
```
jtigger: We're referencing a constant that's defined in the example code... code that the practitioner will not have when the exercise is fetched.

It's legit to include a starter class that includes the constants.

exercises/pythagorean-triplet/src/test/java/PythagoreanTripletTest.java
```diff
@@ -0,0 +1,95 @@
...
+        final List<Long> actual
+                = PythagoreanTriplet.createTripletsList(searchData)
+                        .stream()
+                        .map(t -> t.calculateProduct())
+                        .sorted((p1, p2) -> Double.compare(p1, p2))
```
jtigger: Is `sorted()` sufficient (i.e. without a comparator)?

jtigger:
API design...

In the spirit of playing with different ideas...

This looks like a great opportunity to introduce the [Builder](https://en.wikipedia.org/wiki/Builder_pattern) pattern...

It might read something like:

```java
PythagoreanTriplet.makeTripletsList()
  .withFactorsGreaterThan(10)
  .withFactorsLessThan(21)
  .build();
```

or

```java
PythagoreanTriplet.makeTripletsList()
  .withFactorsLessThan(101)
  .thatSumTo(180)
  .build();
```

Given that the description of the exercise is lean on details, it might be helpful to use a fluent interface like this to convey the intent even more crisply.

To be clear, this is just me being a design buddy; this is **not** a requirement for your PR to be accepted, Dmitry.

exercises/pythagorean-triplet/src/test/java/PythagoreanTripletTest.java
javaeeeee: Done.

exercises/pythagorean-triplet/src/test/java/PythagoreanTripletTest.java
javaeeeee: If elements implement Comparable, they are sorted in natural order, and the ClassCastException may be thrown otherwise. In our case we have Long elements, so can get by without the comparator. :)

javaeeeee:
Thank you for the idea. I love Builder pattern. :)

javaeeeee:
Hello John, changed the proposed method names to accommodate equality.

javaeeeee:
Thank you. Special thanks for  the builder idea, it was interesting to
apply.

PR #190, opened 2016-11-27
Delete all unused .keep files

stkent:
These were leftovers from cloning the template exercise; if a directory contains any other tracked file, they become unnecessary.


jtigger:
Thank you, sir!

PR #206, opened 2016-12-05
custom-set: add to track

javaeeeee:
dibs: I will implement custom-set


matthewmorgan:
@javaeeeee you have been contributing so much lately, thank you!  Sorry that it's taken so long to get to these pull requests, but I wanted to give them proper attention.

I have to challenge the implementation of this contribution, and ask the following question: Do you think there might be a better way to do this?  Or do you feel this implementation makes sense the way it is?

Let me explain: one of the core features of the Java language is the ability to apply object-oriented principles, like extending base classes.  Your `CustomSet` class uses a Java `Set` for storage under the hood, and provides methods such is `isEmpty` by simply returning `set.isEmpty()`.

I would suggest that one of two approaches might provide a better lesson for students of Java on Exercism:
1. Use the base class by extending it to define `CustomSet`.  Add or override only the methods required to meet the spec.
2. Go 'closer to the metal.'  Don't use `Set` at all.  Instead create an `Element` class (or some name of your choosing) that implements 'hashCode` and `equals()`, and implements Set behaviors without leveraging any class from the Java Collections library.

What do you think about one of those two approaches?

javaeeeee:
Hello Matthew,

Thank you for taking time and commenting on the exercise.

If I were solving this exercise, I would implement it as a Tree set or a
Hash set and wouldn't use Java Set at all, as it would fly in the face of
the goal of this exercise.

In my opinion, this particular exercise could be seen as an opportunity to
reinvent the wheel and try to implement the Set interface.

As my goal was to add the code to "test the test", I've selected the
quickest way to rely on the Java Set implementation.

The README files in some exercises explicitly state that if the language
used to implement the assignment has a construct like a set, a filter, etc.
that will make the solution a snap should be avoided.

Going back to Set extension, I didn't select this approach because of
the lesson learned from the Java Stack class, which relies on
extension rather than composition. The idea is that in the exercise, which
is common for all language tracks, one has to implement some custom
contract which is not the same as offered by the Java Set interface's one.
So, in order not to expose the undesired methods, I preferred composition.

Please inform me about your decision if it is necessary to rewrite the
code. I would be glad to go with any approach.

Best regards,

Dmitry

matthewmorgan:
@javaeeeee I completely respect your decision to use composition over inheritance.  Thank you for your explanation!

javaeeeee:
Thank you.

PR #208, opened 2016-12-15
Issue #193: Changed assertions in AnagramTest to use junit instead of…

FridaTveit:
… assertj. Removed assertj from Anagram gradle build file. This should fix issue #193.


jtigger:
Perfect!  Thanks for picking these up, @FridaTveit.

PR #211, opened 2016-12-16
Issue #195: Changed assertj assertions to junit and hamcrest assertio…

FridaTveit:
…ns in grade-school. Removed assertj from grade-school gradle build file. This should fix issue #195 .


matthewmorgan:
👍

PR #212, opened 2016-12-16
Issue #196: Replaced assertj assertions with junit and hamcrest asser…

FridaTveit:
…tions in linked-list. Removed assertj from linked-list gradle build file. This should fix issue #196 .


matthewmorgan:
Nice!

PR #214, opened 2016-12-17
Issue #198: Replaced assertj assertions in Meetup test with junit and

FridaTveit:
hamcrest assertions. Removed assertj from meetup gradle build file. This should fix issue #198 .


matthewmorgan:
Thanks @FridaTveit !  A lot of tests in this one...

PR #215, opened 2016-12-17
Issue 199: Replaced assertj assetions in nth-prime test with junit an…

FridaTveit:
…d hamcrest assertions. Removed assertj from nth-prime gradle build file. This should fix issue #199 .


matthewmorgan:
Lovely!  Thanks so much.

PR #217, opened 2016-12-18
Issue #200: Replaced assertj assertions in nucleotide-count tests wit…

FridaTveit:
…h junit and assertj assertions. Removed assertj from nucleotide-count gradle build file. Added extra hamcrest library to nucleotide-count gradle build file to be able to use hasEntry for maps. This should fix issue #200 .

### DIFF
--- a/exercises/nucleotide-count/build.gradle
+++ b/exercises/nucleotide-count/build.gradle
@@ -8,7 +8,7 @@ repositories {
 
 dependencies {
   testCompile "junit:junit:4.12"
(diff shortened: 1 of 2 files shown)


jtigger:
Yup!  That's perfect, @FridaTveit.  Thanks for your help!  These PRs move us toward making the experience more straightforward for all practitioners coming through the Java track.

If you see other things that qualify as "unnecessary complication" feel free to point those out.

jtigger:
Fixes #200.

PR #221, opened 2016-12-20
Make Oracle JDK default for Debian distros

jtigger:
Without this instruction, the system continues to use the default OpenJDK.  This is confusing since we instruct the practitioner to install Oracle's JDK.


stkent:
I don't know enough about this to be anything like definitive, but this change _sounds_ very reasonable indeed!

jtigger:
Yeah, this is an annoyingly labor intensive thing to validate.  You'd have to have a fresh OS install and go through the steps to verify that a) the symptom was occurring and b) that the suggested fix addresses that symptom.

I'm doing all this work on a VirtualBox image so that I can reset back to a known state and try again.

If you'd like, you can tacitly approve and just merge. :)

stkent:
I'm fine to roll with that; we'll hear from users if it's a problem.

jtigger:
:+1:

Issue #151, opened 2016-10-12
Ordered lists in documentation are failing to continue after codeblocks


jtigger:
---
## tl;dr

fix the broken numbering in the `docs/.md` files.

---
## Context

When properly formatted, markdown converts this:

```
1. first item
- second item
- third item
```

to this:
1. first item
2. second item
3. third item

There is a slight difference between how GitHub is parsing ordered lists and how the templating implementation being used in exercism.io is parsing/rendering it when those ordered lists include codefences.

Take for example who this:

> https://github.com/exercism/xjava/blob/master/docs/INSTALLATION.md

varies from this:

> http://exercism.io/languages/java/installing
## The Work

Your mission (should you choose to accept it) is to figure out how to format the markdown in those files such that the numbering continues even after a codefence is included.  (i.e. that it renders on the website the same way that GitHub renders it).

PR #223, opened 2016-12-21
Issue #151: Changed markdowns to have actual number to prevent them b…

FridaTveit:
…reaking when the lists include codefences.


jtigger:
Terrific!  Thank you, @FridaTveit!

Issue #152, opened 2016-10-12
simple-cipher: make more approachable


jtigger:
## Context

This particular exercise is painful to approach:
- the goal is unclear, yet the path is apparently (if the size of the description is any indication) long (this problem is detailed in exercism/x-common#404 )
- the test structure is unnecessarily confusing:
  - for the first time in the track, we include a suite.  The suite is unnecessary when running the suite with Gradle.
  - there are multiple failures as there are missing `@Ignore`s.
  - the test classes don't immediately line-up with which "step" called out in the description.
- there's a relatively significant API that's spread across the various tests; perhaps we should give the programmer a leg-up here since we're asking so much of them... and provide a skeleton starter.

bleh!  This exercise really really frustrated me (and I've seen complains on the site too).

Let's make some lemonade...

---
## The Story

As a programmer solving the `simple-cipher` exercise:
1. after I read the description, when I first review the tests, it is clear to me which step/iteration is being covered by which test class.
2. when I first run the tests, exactly one test case fails and it is an assertion failure.
3. as I read the test, the three strings: plaintext, ciphertext, and key jump out at me (because they are given descriptive names).
4. as I complete the first test class, it is obvious which test class I should move on to, next.
5. as I work my way through a single test class, it covers the corresponding step/iteration from the description and _only_ the corresponding step/iteration.

Notes:
- while `SubstitutionCipherTest#cipherCanDoubleShiftEncode` is interesting and fun, it is misplaced: it is the only test case in the test class that ignores the test setup and does its own thing.

PR #224, opened 2016-12-22
Issue #152: Made simple-cipher more approachable

FridaTveit:
Issue #152: Made simple-cipher more approachable by splitting test classes in a more sensible way and making clear which steps they relate to.


stkent:
I'd like to defer to @jtigger for review of this PR, since I haven't worked through the exercise yet!

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Ignore
+    @Test(expected = IllegalArgumentException.class)
+    public void cipherThrowsWithAnyNumericKey() {
+        new Cipher("abcd345ef");
+    }
+
+    @Ignore
+    @Test(expected = IllegalArgumentException.class)
+    public void cipherThrowsWithEmptyKey() {
```
jtigger: What do you think of introducing this in step two?  That's when the key gets included and there's really only an empty check required.

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    public void cipherKeysAreRandomlyGenerated() {
+        String newKey = new Cipher().getKey();
+        assertFalse("Cipher constructor without argument should generate a random key. No two calls to the" +
+                " constructor should generate the same key. Two calls to the constructor " +
+                "both returned key: " + newKey, newKey.equals(new Cipher().getKey()));
+    }
+
+    @Ignore
+    @Test(expected = IllegalArgumentException.class)
```
jtigger: This is slightly out of scope for the work you're doing, @FridaTveit, but if you're so inclined...

... since we've upgraded to JUnit 4.12, we've been migrating to using JUnit's [`ExpectedException`](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) `@Rule` feature for this situation.

If you so choose, since we're making changes, here could convert these to use that notation.

exercises/simple-cipher/src/test/java/SimpleCipherStepTwoTest.java
```diff
@@ -1,79 +1,74 @@
...
 import static org.junit.Assert.assertEquals;
 
-public class SubstitutionCipherTest {
-
+/**
+ * Step 2: Specify key and use that for shift distance : substitution cipher
+ */
+public class SimpleCipherStepTwoTest {
+    private Cipher cipherWithSetKey;
     private static final String KEY = "abcdefghij";
```
jtigger: Another opportunistic improvement suggestion (feel free to ignore, if you disagree/don't want to)...

Personally, I'm not a fan of using constants as fixtures in tests.  The fact that it is a value used in a test case is far far more important than the incidental fact that it's constant (all String literals *are* constant in Java).

Seems to me that `"abcdefghij"` is playing the role of an important piece of data for *each* test case (exactly like `cipherWithSetKey`) ... call it `key`?

exercises/simple-cipher/src/test/java/SimpleCipherStepTwoTest.java
```diff
@@ -1,79 +1,74 @@
...
     @Test
     public void cipherKeepsTheSubmittedKey() {
-        assertEquals(KEY, cipher.getKey());
+        assertEquals(KEY, cipherWithSetKey.getKey());
     }
 
     @Ignore
     @Test
     public void cipherCanEncodeWithGivenKey() {
         String expectedOutput = "abcdefghij";
```
jtigger: Can we come up with a more intention revealing name than `expectedOutput`?  What is the string that's returned from `encode()` represent?

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeTest.java
FridaTveit: Done :)

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeTest.java
FridaTveit: Sure :)

exercises/simple-cipher/src/test/java/SimpleCipherStepTwoTest.java
FridaTveit: Sure, done :)

exercises/simple-cipher/src/test/java/SimpleCipherStepTwoTest.java
FridaTveit: I've renamed it to cipherText :)

jtigger:
@FridaTveit, what do you think of including a little instruction explaining that there are multiple test classes and how to run a single test class in `HINT.md`?

The way it works is if there's a `HINT.md` file in the root of an exercise directory, that markup will [get merged in to the `README.md`](https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L89) delivered to the practitioner.

This can be helpful in a case like this.  Right now, if the practitioner runs the tests:

```
$ gradle test
```

They get a huge wall of text.  However, if they run just the first step:

```
$ gradle test --tests *StepOne*
```

They get a smaller more manageable wall of text. :)

Part of the spirit, but not the letter of #152; if you'd rather not, the rest of this is great work (thank you!).

FridaTveit:
Sorry, forgot about that comment! Sure will do! :)

exercises/simple-cipher/HINT.md
```diff
@@ -0,0 +1,9 @@
+This exercise has three test classes: StepOne.java, StepTwo.java and StepThree.java. To run all test classes:
+```
+$ gradle test
+```
```
jtigger: This is rather subtle, but the "Track Hint" [SETUP.md](https://github.com/exercism/xjava/blob/master/SETUP.md) is inserted before the "Implementation Hint" (this file, `HINT.md`), so you'll want to consider the contents of the first file too.

In that merge, this instruction to run `gradle test` will appear twice.

exercises/simple-cipher/HINT.md
```diff
@@ -0,0 +1,9 @@
+This exercise has three test classes: StepOne.java, StepTwo.java and StepThree.java. To run all test classes:
```
jtigger: Practitioners might find it confusing to refer to the tests by filenames that doesn't match the actual files.

FridaTveit: Sorry, forgot to check that those were the actual names! Will change it :)

exercises/simple-cipher/HINT.md
FridaTveit: Oh, okay. Will change it :)

jtigger:
@FridaTveit, thank you so much for putting in the time to clean this exercise up.  Frankly, this was an unnecessarily painful experience before.  The problem is already challenging enough without confusing instructions and test suite.  This is sooooooo much better.

💩  => ♻️  => 🌸❕

Issue #225, opened 2016-12-25
pangram: Test and Class do not match


jtigger:
It's a convention in Java for unit tests to be the concatenation of the class under test and the word `Test`.  We missed that on this exercise.

Please fix!

PR #226, opened 2016-12-26
Renamed PangramTest to PangramsTest to be consistent with class it's …

FridaTveit:
…testing. This should fix issue #225 .


stkent:
I ❤️ consistency  - thanks!

Issue #235, opened 2016-12-31
all-your-base: only the first test should be unignored


stkent:
Currently, no all-your-base tests are marked with the `@Ignore` annotation.

We should add the `@Ignore` annotation to every test but the first.

PR #236, opened 2017-01-03
Added Ignore annotation to all tests but the first one in all-your-base.

FridaTveit:
This should fix issue #235 .


stkent:
Thank you! ❤️

PR #239, opened 2017-01-07
twelve-days: add to track

FridaTveit:


stkent:
Hi @FridaTveit! Thanks for your work implementing this exercise :)

First off, let's get your branch passing the CI build. Click [here](https://travis-ci.org/exercism/xjava/builds/189853989) to see the build logs. Near the top, I spot the following message:

```
>>> Running configlet...
Evaluating .
-> config.json does not include [twelve-days].
```

Our first clue! This suggests that you'll need to update `config.json` to make it aware of your brand new exercise. [Here's an example](https://github.com/exercism/xjava/pull/192/files#diff-06b2d3b23dce96e1619d2b53d6c947ec) of the config changes that a new exercise requires. I think that should give you enough information to update your own branch; if not, let me know!

(If you're wondering why you need to edit the `config.json` file in two different spots, then you can find some context [here](https://github.com/exercism/discussions/issues/60).)

NobbZ:
The problems key can get removed... It is deprecated. All necessary information is in the exercises key, and afaik the problems key isn't even read anymore. When I'm back at my laptop I'll lookup the corresponding upstream issue.

stkent:
@NobbZ seems like we would want to take care of that (if possible) outside this PR. AFAIK we're following the sequence of tasks outlined by https://github.com/exercism/xjava/issues/125. If it is true that the `problems` key is completely ignored upstream, that's useful to know!

stkent:
@FridaTveit this needs a merge with master now, but it shouldn't be too painful!

config.json
```diff
@@ -305,7 +306,12 @@
       "slug": "palindrome-products",
       "difficulty": 1,
       "topics": []
-    }
+    },
+	{
+	  "slug": "twelve-days",
+	  "difficulty": 1,
+	  "topics": []
+	}
```
stkent: Oops, indents in this file seem a little wonky; please see if you can tweak them to be consistent with the existing content!

exercises/twelve-days/src/test/java/TwelveDaysTest.java
```diff
@@ -0,0 +1,172 @@
...
+        assertEquals(expectedVerseEleven, TwelveDays.verse(11));
+    }
+
+    @Ignore
+    @Test
+    public void testVerseTwelve() {
+        String expectedVerseTwelve = "On the twelfth day of Christmas my true love gave to me, " +
+                "twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, " +
+                "eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, " +
+                "four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.\n";
```
stkent: Hmm, is the final newline an intrinsic part of these verses? In my mind that's a verse separator, so I'd not necessarily expect it to be included when requesting an individual verse. I do think the double newline separators make sense when printing out more than one verse though. What do you think, @FridaTveit?

exercises/twelve-days/src/test/java/TwelveDaysTest.java
```diff
@@ -0,0 +1,172 @@
...
+                "\n" +
+                "On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, " +
+                "ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, " +
+                "six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, " +
+                "and a Partridge in a Pear Tree.\n" +
+                "\n" +
+                "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, " +
+                "eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, " +
+                "seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, " +
+                "three French Hens, two Turtle Doves, and a Partridge in a Pear Tree.\n";
```
stkent: It seems like a "gotcha" that this test string ends in a single newline, but the test string for `testMultipleVerses` ends in a double newline. I could imagine that this might confuse implementors a little bit! Do you think it would be possible to make them consistent (as above, I would lean towards no trailing newlines in either case, but I think consistency of some kind here is more important than the actual choice we make)?

config.json
FridaTveit: That's strange, it's not wonky when I open it...

exercises/twelve-days/src/test/java/TwelveDaysTest.java
FridaTveit: I didn't have the newline there originally, but I changed it when I saw that the exercise in other languages had the newline at the end of every verse. I think it's because if you copy a line of the song from the description, you get the newline as well.

exercises/twelve-days/src/test/java/TwelveDaysTest.java
FridaTveit: There's no double newline in this one, because there's no double newline at the end of the song in the description. Personally, I don't think the multiple verses should end in a double newline. However, in the implementation of twelve days in other languages the multiple verses end in a double newline. So I'm happy to change the multiple verses to end with a single newline if you're happy with my implementation not being exactly the same as in other languages. I wasn't sure how important it was to be consistent with other implementations...

jtigger:
I can't help but comment on the quality of the intent and review, here.

@FridaTveit, it's exactly things like looking at how the same exercise is implemented in other tracks that make for a cohesive experience on Exercism; excellent instincts.

And @stkent, your attention to detail is something that pays dividends every time someone types `exercism fetch java`.

The specific topics may seem like nitpicks at times, but these are the ingredients of the programmer's UX.  We want the challenge of designing the solution to be the singular focus of practitioners; your focusing on getting all the other bits "right" is what makes that happen.

Thank you, both! 🙏

config.json
jtigger: mixture of tabs and spaces?

FridaTveit: Thanks, that was it! :)

exercises/twelve-days/src/test/java/TwelveDaysTest.java
stkent: Ah, that makes sense r.e. the description of the problem. Thanks!

exercises/twelve-days/src/test/java/TwelveDaysTest.java
stkent: I would say; let's break with the other tracks here for the sake of consistency in our own track. I'll add a further note below in response to @jtigger's comments regarding cross-track checks :)

stkent:
@jtigger thank you for the kind words! I agree wholeheartedly that referencing implementations in other tracks as @FridaTveit did is a great way to seed our own implementations. However, I also like to look at them as inspiration and make sure I ask "can we do better?". In this particular instance, I think that double newline in the entire song is something the Java track can do better than other tracks (for the same reasons I originally stated); whether or not it's a big enough issue to submit upstream, I'm not sure.

@FridaTveit this looks ready for merge as soon as the final trailing newline is chopped off the entire song! 🏆

stkent:
Edit to the above:

@FridaTveit this looks ready for merge as soon as the final trailing newline is chopped off **the multiple verses test, so that both it and the entire song end in a single newline** 🏆

stkent:
Lovely, thanks! A great addition to the track; let's get this merged!

jtigger:
>... However, I also like to look at them as inspiration and make sure I ask "can we do better?". ...

@stkent, I couldn't agree more.

PR #240, opened 2017-01-09
sublist: add to track

stkent:
[x-common exercise directory](https://github.com/exercism/x-common/tree/master/exercises/sublist)

The tests require that the implementor accommodate `String`, `Integer` and `Character` list element types. All three implement `Comparable`.


stkent:
Agreed on the overlap, especially the start/middle/end checks. I would expect that implementors would find they could uncomment chunks of tests quite easily after building an implementation that works for the early ones.

jtigger:
Yeah, and this is not the only exercise like that... might be something of a discussion for the Jedi Council. ;)

FWIW, I added `sublist` to the [difficulty spreadsheet](https://docs.google.com/spreadsheets/d/15UbVn4cBUxI7tekKmNNls9dXsT1rDhtG7Z8lp1nmjlM/edit#gid=0); when you get a sec, throw down a number on it.

stkent:
Added a rating and also tweaked my opinion of bracket-push (5 -> 6)!

PR #244, opened 2017-01-10
dibs: I will implement exercise Matrix

FridaTveit:


jtigger:
Go Frida, go!

jtigger:
Hey @FridaTveit, would you mind connecting with me over email (on my GitHub profile)?  I have an Exercism-related question for you.

stkent:
I'm curious to see what you choose to do with the problem interface here, @FridaTveit - I have pondered this puzzle a little in the past!

jtigger:
You've pondered the puzzle of this problem's protocol? (sorry, couldn't help myself... carry on)

stkent:
Hah, indeed! In preparation for pontification!

FridaTveit:
I added a commit with the implementation a while ago but I rebased it on top of master instead of merging it, is that a problem?

stkent:
Ah, no, that's perfectly fine, I had just missed it! I normally update the PR title to `exercise-slug: add to track` once I had an implementation that's ready for review, just to call out the difference :)

FridaTveit:
Ah, okay :)

jtigger:
Hey @stkent I'm assuming that you're reviewing this PR; that your intent?

stkent:
Sure, I'm happy to; probably won't get to it for a couple days though, so if you want to jump in, that's fine by me.

exercises/matrix/src/example/java/Matrix.java
```diff
@@ -0,0 +1,35 @@
...
+    private int[][] matrix;
+
+    public Matrix(String matrixAsString) {
+        String[] rows = matrixAsString.split("\\n");
+        matrix = new int[rows.length][];
+        for (int i = 0; i < rows.length; i++) {
+            String[] columnValues = rows[i].split(" ");
+            matrix[i] = new int[columnValues.length];
+            for (int j = 0; j < columnValues.length; j++) {
+                matrix[i][j] = Integer.valueOf(columnValues[j]);
```
stkent: Minor detail, but it's probably ok to use `Integer.parseInt` here instead, which avoid extra boxing/unboxing. See e.g. [this SO post](http://stackoverflow.com/questions/508665/difference-between-parseint-and-valueof-in-java) for some interesting discussion. Not going to hold the PR for this change alone, just noting for interest :)

exercises/matrix/src/example/java/Matrix.java
```diff
@@ -0,0 +1,35 @@
+public class Matrix {
+    private int[][] matrix;
+
+    public Matrix(String matrixAsString) {
+        String[] rows = matrixAsString.split("\\n");
+        matrix = new int[rows.length][];
+        for (int i = 0; i < rows.length; i++) {
+            String[] columnValues = rows[i].split(" ");
```
stkent: Another FYI, not a required change: note that there are some "gotchas" when using `String.split` with a regex. [Here's a nice breakdown](http://chrononsystems.com/blog/hidden-evils-of-javas-stringsplit-and-stringr). If this was high-performance code, you'd probably want to compile the `Pattern` ahead of time, and split using that instead. I only learned about this a couple months back.

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -0,0 +1,184 @@
...
+    public static class GetFirstRowTest {
+        private Matrix matrix;
+        private int[] firstRow;
+
+        @Parameterized.Parameters(name = "{index}: checking first row of matrix constructed with string \"{0}\".")
+        public static Collection<Object[]> data() {
+            return Arrays.asList(new Object[][]{
+                    {"0", new int[] {0}},
+                    {"0 1", new int[] {0, 1}},
+                    {"0\\n1", new int[] {0}},
```
stkent: Do these strings all need to include the escape newline? Everything seemed to work ok for me if I replaced all occurrences of "\\n" with a simple "\n"? On a related note, would that change make the `matrixAsString.replaceAll("\\\\n", "\n")` lines removable too?

stkent: Ohh... is this for formatting the parametrized test output? If so, that's a tricky balance; I fear adding then adjusting the newlines in this way might cause implementors some confusion.

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -0,0 +1,184 @@
...
+        private int[] firstColumn;
+
+        @Parameterized.Parameters(name = "{index}: checking first column of matrix constructed with string \"{0}\".")
+        public static Collection<Object[]> data() {
+            return Arrays.asList(new Object[][]{
+                    {"0", new int[] {0}},
+                    {"0 1", new int[] {0}},
+                    {"0\\n1", new int[] {0, 1}},
+                    {"0 1\\n2 3", new int[] {0, 2}},
+                    {"0 1 2\\n3 4 5\\n6 7 8", new int[] {0, 3, 6}}
```
stkent: I wonder if it might be possible to format the matrices here in a more intuitive way? Something like the following (last row example):

```java
@Parameterized.Parameters(name = "{index}: checking first row of matrix constructed with string \"{0}\".")
public static Collection<Object[]> data() {
    return Arrays.asList(new Object[][]{
            {"0", new int[] {0}},
            {"0 1", new int[] {0, 1}},
            {"0\n1", new int[] {0}},
            {"0 1\n2 3", new int[] {0, 1}},
            {  "0 1 2\n"
             + "3 4 5\n"
             + "6 7 8", new int[] {0, 1, 2}}
    });
}
```

I find it a little easier to draw the connection between the input and the expected output with that formatting, but others might disagree...?

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -0,0 +1,184 @@
...
+            });
+        }
+
+        public GetFirstColumnTest(String matrixAsString, int[] firstColumn) {
+            this.matrix = new Matrix(matrixAsString.replaceAll("\\\\n", "\n"));
+            this.firstColumn = firstColumn;
+        }
+
+        @Ignore
```
stkent: I think these `@Ignore`s could also be applied so the inner classes themselves instead of the nested `@Test` methods, which could make them a little easier to spot? e.g.

```java
@Ignore
@RunWith(Parameterized.class)
public static class GetLastRowTest {
```

As before, though, it might just be me that finds the top-level annotation easier to see!

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        public GetLastColumnTest(String matrixAsString, int[] lastColumn) {
+            this.matrix = new Matrix(matrixAsString.replaceAll("\\\\n", "\n"));
+            this.lastColumn = lastColumn;
+        }
+
+        @Ignore
+        @Test
+        public void getLastColumnTest() {
+            assertArrayEquals(matrix.getColumn(matrix.getColumnsCount() - 1), lastColumn);
```
stkent: JUnit assertions typically request the expected value before the actual value, so we should reverse the parameter ordering of all of these calls to `assertArrayEquals` - otherwise our failure messages will be backwards and potentially confusing!

stkent:
@FridaTveit some comments in; mostly musing out loud 😄 . Let me know what you prefer in the cases where I suggested alternatives!

exercises/matrix/src/example/java/Matrix.java
FridaTveit: Okay, will change it! :)

exercises/matrix/src/example/java/Matrix.java
FridaTveit: Thanks, good to know! :)

exercises/matrix/src/test/java/MatrixTest.java
FridaTveit: Yes, it is for formatting the test output and I was unsure about it. I agree that adjusting the newlines looks confusing when you look at the code. So having the test output look a bit less nice is probably the better option. I'll change it :)

exercises/matrix/src/test/java/MatrixTest.java
FridaTveit: Sure, that looks good, I'll change it :)

exercises/matrix/src/test/java/MatrixTest.java
FridaTveit: Sure, that makes sense, I'll change it :)

exercises/matrix/src/test/java/MatrixTest.java
FridaTveit: Thanks, will change it! :)

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        public GetLastRowTest(String matrixAsString, int[] lastRow) {
+            this.matrix = new Matrix(matrixAsString);
+            this.lastRow = lastRow;
+        }
+
+        @Test
+        public void getLastRowTest() {
+            assertArrayEquals(lastRow, matrix.getRow(matrix.getRowsCount() - 1));
```
stkent: Since this test uses `getRowsCount` to identify the final row of the matrix, I wonder if we should move the tests for `CountColumnsTest` and `CountRowsTest` classes to the very beginning of the test suite? Otherwise those test classes are effectively redundant, as the implementor will already have had to build working implementations of `getRowsCount` and `getColumnsCount` to pass earlier tests.

stkent:
@FridaTveit thanks, changes look great! I added one more question that I didn't spot the first time around - let me know what you would like to do with that. Once that's been discussed and/or resolved, let's get this merged!

stkent:
(Small merge with master also needed now :)

exercises/matrix/src/test/java/MatrixTest.java
FridaTveit: True, will move them! :)

stkent:
I took care of those config conflicts using the new inline GitHub merge tool. It worked well! Merging on successful build 🏆

Issue #241, opened 2017-01-09
Reorder queen-attack tests


stkent:
Per discussion in https://github.com/exercism/xjava/issues/234, the tests in `queen-attack` should be reorganized as follows:

- happy path tests first
- exceptional case tests last.

PR #248, opened 2017-01-16
Reorder queen-attack tests

colin-mullikin:
Re-ordered the tests according to issue #241.

### DIFF
--- a/exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java
+++ b/exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java
@@ -8,6 +8,13 @@ import static org.junit.Assert.assertTrue;
 
 public final class QueenAttackCalculatorTest {
 
+    /*
+     * See https://github.com/junit-team/junit4/wiki/Rules for information on JUnit Rules in general and
+     * ExpectedExceptions in particular.
+     */
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
     @Test
     public void testQueensThatDoNotShareRankFileOrDiagonalCannotAttack() {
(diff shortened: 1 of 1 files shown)


exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java
```diff
@@ -8,127 +8,127 @@
...
-        assertFalse(calculator.canQueensAttackOneAnother());
+        assertTrue(calculator.canQueensAttackOneAnother());
     }
 
+    /*
+     * See https://github.com/junit-team/junit4/wiki/Rules for information on JUnit Rules in general and
+     * ExpectedExceptions in particular.
+     */
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
```
FridaTveit: I would keep this at the top of the class, where it was before, even though you're not using it until here. If you keep all field declarations at the top of the class then they're easy to find if you're looking for them :)

colin-mullikin: Okay. That was the only bit that I was unsure of. I'll move that back to the top where it was.

FridaTveit:
By the way, for future reference it's probably better to make a pull request from a feature branch instead of from your master branch. When you've made a pull request any further changes pushed to the same branch will automatically be added to the pull request. So it's easy to accidentally add changes you didn't intent to :)

colin-mullikin:
Yeah, I realized that immediately after I did it. Figured it would get approved and merged pretty quickly, so I didn't bother to go back and spend the time to do it that way.

FridaTveit:
Fair enough :)

PR #249, opened 2017-01-18
Make root gradle project aware of binary-search again

stkent:


jtigger:
doh!

stkent:
I think it probably died when `binary-search-tree` was added and we missed it during review! All better now.

PR #254, opened 2017-01-20
rectangles: polish example implementation

stkent:
Following on from https://github.com/exercism/xjava/pull/243.

Renames `Grid.Entry` to `Grid.Element` and adds a couple query methods to the `Grid.Element` type for readability.

### DIFF
--- a/exercises/rectangles/src/example/java/RectangleCounter.java
+++ b/exercises/rectangles/src/example/java/RectangleCounter.java
@@ -15,34 +15,42 @@ final class RectangleCounter {
 
     private static final class Grid {
 
-        private enum Entry {
+        private enum Tile {
             CORNER,
-            HLINE,
-            VLINE,
+            HORIZONTAL_WALL,
+            VERTICAL_WALL,
             SPACE;
 
-            private static Entry fromChar(final char rawGridEntry) {
-                switch (rawGridEntry) {
+            private static Tile fromChar(final char rawGridTile) {
+                switch (rawGridTile) {
                     case '+': return CORNER;
-                    case '-': return HLINE;
-                    case '|': return VLINE;
+                    case '-': return HORIZONTAL_WALL;
+                    case '|': return VERTICAL_WALL;
                     case ' ': return SPACE;
-                    default:  throw new IllegalStateException("Grid entry " + rawGridEntry + " not recognized.");
+                    default:  throw new IllegalStateException("Grid tile " + rawGridTile + " not recognized.");
                 }
+            }
+
+            private boolean isHorizontalConnector() {
+                return this == CORNER || this == HORIZONTAL_WALL;
+            }
+
+            private boolean isVerticalConnector() {
+                return this == CORNER || this == VERTICAL_WALL;
             }
         }
 
         private int nRows, nCols;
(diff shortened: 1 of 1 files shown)


exercises/rectangles/src/example/java/RectangleCounter.java
```diff
@@ -15,13 +15,13 @@ int countRectangles(final String[] rawGrid) {
...
 
-        private enum Entry {
+        private enum Element {
             CORNER,
             HLINE,
             VLINE,
             SPACE;
 
-            private static Entry fromChar(final char rawGridEntry) {
+            private static Element fromChar(final char rawGridEntry) {
```
jtigger: I sense an intent to want to pair-up the input and the output, by picking a similar name; if so, that got lost in the rename of the output type.

----

tl;dr — `Piece`?

I couldn't help myself... I had to attempt to suggest names for these little bits of the diagram (not that `Element` isn't "right"... just nice having options).  In that journey I discovered that the Dictionary/Thesaurus that comes OOTB with OS X has a neat little section on some words called "Choose the right word".  There's one such entry for "part".  It reads like this:

> The whole is equal to the sum of its parts — **part** being a general term for any of the components of a whole. But how did the whole come apart? **Fragment** suggests that breakage has occurred (fragments of the vase) and often refers to a brittle substance such as glass or pottery. **Segment** suggests that the whole has been separated along natural or pre-existing lines of division (a segment of an orange), and **section** suggests a substantial and clearly separate part that fits closely with other parts to form the whole (a section of the bookcase). **Fraction** usually suggests a less substantial but still clearly delineated part (a fraction of her income), and a **portion** is a part that has been allotted or assigned to someone (her portion of the program). Finally, the very frequently used **piece** is any part that is separate from the whole (we're missing two pieces from the jigsaw puzzle).

exercises/rectangles/src/example/java/RectangleCounter.java
```diff
@@ -30,19 +30,27 @@ private static Entry fromChar(final char rawGridEntry) {
...
                 }
             }
+
+            private boolean isHorizontalConnector() {
+                return this == CORNER || this == HLINE;
+            }
+
+            private boolean isVerticalConnector() {
+                return this == CORNER || this == VLINE;
+            }
```
jtigger: ooh! I ❤️  `...Connector`, here!

exercises/rectangles/src/example/java/RectangleCounter.java
stkent: I'm sick, so apologies for the abbreviated reply; but your `Piece` made me think of another option: `Tile`! I'm happy to use whichever of those two you prefer :)

jtigger: > I'm sick, so apologies for the abbreviated reply ...

Unacceptable; sickness is for the weak. 😷 

Oooohhh I love `Tile` too!  I dig the physicality of it.  Seems like the best names are those that are the most concrete but not overly-so.

Hope you're feeling better, soon, Stuart! 🤒

stkent:
Updated!

jtigger:
`Tile` looks GREAT! :1st_place_medal:

PR #255, opened 2017-01-20
flatten-array: add to track

stkent:
[x-common exercise directory](https://github.com/exercism/x-common/tree/master/exercises/flatten-array)


stkent:
Note: this implementation is not type-safe. I wrote an alternate type-safe implementation too: https://github.com/exercism/xjava/compare/flatten-array-add-to-track-type-safe (needs some cleanup/de-duplication). Inviting thoughts on which is more suitable here!

matthewmorgan:
@stkent we did this as our Code Kata lunch at work a few weeks ago!  I would be in favor of seeing the typesafe version.  I think it's a very useful lesson for Java learners.

stkent:
@matthewmorgan ah, great timing! I'm curious, then, what you think of the linked type-safe version? Some of the API felt a _little_ clumsy to me, so any tips would be appreciated!

matthewmorgan:
@stkent there's a lot to like about that implementation.  The recursion of course, for example.  I do find myself wondering about the use of the `Optional`-style class of `ValueOrList`...at first reaction, I thought it might be better to overload the `flatten` method to take either a List or a Value, but then, we don't really know what our 'values' might be, except that we think they're not Lists, correct?

Your approach may be the best way to handle that.  I'd be curious to hear what @jtigger and @FridaTveit think about it.

stkent:
With regards to overloading: I'm not sure that can handle the requirement that we support inhomogeneous lists. We are forced to confront that inhomogeneity head on: the only two choices that I see are to give up on typing entirely (since the nearest shared ancestor of `T` and `List<T>` is `Object`) or to use the specialized `Either`-style construct from the other branch.

stkent:
And yes, definitely interested in what our other maintainers think on this one!

jtigger:
Hmmm... trying to reconcile this conversation with what I see in the code.  I'll do my best, but warning I might be off base. :)

### inhomogeneity 

It took me a while to get there, but I see that this is a great example of where a raw collection is very useful.  `Flattener` makes no assumption whatsoever about the contents of the list.  I love it! 👍

In fact, I'd recommend that we really double-down on this fact and deliberately put all kinds of objects in any given list. 💡❔ 

### Overloading

With respect to:

```java
if (nestedList.get(0) instanceof List) {
	result.addAll(flatten((List) nestedList.get(0)));
} else {
	result.add(nestedList.get(0));
}
```

I'm pretty sure the overloading will work.  It would eliminate the type-check and cast.  It will handle the inhomogeneity since the signature of the value overload will be `Object`.

stkent:
Thanks for the input, both of you! I agree that the type-safe implementation necessarily involves a lot of pedantry; the non-type-safe code is a lot more intuitive to parse. I'll try adding the overload described to the non-type-safe version and we can see how that feels :)

stkent:
Sanity check please: if I try to replace

```
if (nestedList.get(0) instanceof List) {
    result.addAll(flatten((List) nestedList.get(0)));
} else {
    result.add(nestedList.get(0));
}
```

with something like

```
result.addAll(flatten(nestedList.get(0)));
```

then my overloaded `flatten` that takes an `Object` would need to take that `Object` and wrap it in a `List`, which feels ugly to me. The fact that there's no `List.add` overload would seem to scupper these plans, no?

jtigger:
No, you're sane.

I got tripped up thinking that method selection was based off execution-time types.  It's compile-time types.  Ahh geez! :man_facepalming:

matthewmorgan:
@jtigger my problem exactly...

I am with you in thinking it would be nice to have an array of mixed types, but the canonical data doesn't call for it.  @stkent @jtigger @FridaTveit do we feel OK expanding the scope a little?

stkent:
I think so! I tried to mix it up while preserving the spirit of the original tests (rebased).

exercises/flatten-array/src/example/java/Flattener.java
```diff
@@ -0,0 +1,25 @@
...
+
+final class Flattener {
+
+    List flatten(final List nestedList) {
+        if (nestedList.isEmpty()) {
+            return new ArrayList<>();
+        } else {
+            final List result = new ArrayList();
+
+            if (nestedList.get(0) instanceof List) {
```
stkent: I should avoid calling `get` three times here!

stkent:
Good for another look!

PR #256, opened 2017-01-20
diamond: add to track

stkent:
[x-common exercise directory](https://github.com/exercism/x-common/tree/master/exercises/diamond)


stkent:
I'd like to propose the tests in this suite as canonical. Most other tracks test only the 3 examples from the exercise description, which means they never test:

- A diamond with an even side length;
- The extreme case in which all letters of the alphabet are used.

If you think that the suite in my implementation needs augmentation before being adapted to a canonical set, please let me know!

exercises/diamond/src/main/java/DIamondPrinter.java
```diff
@@ -0,0 +1,5 @@
+final class DIamondPrinter {
```
stkent: Typo.

exercises/diamond/src/test/java/DiamondPrinterTest.java
```diff
@@ -0,0 +1,111 @@
...
+                                     "                J                 J                ",
+                                     "                 I               I                 ",
+                                     "                  H             H                  ",
+                                     "                   G           G                   ",
+                                     "                    F         F                    ",
+                                     "                     E       E                     ",
+                                     "                      D     D                      ",
+                                     "                       C   C                       ",
+                                     "                        B B                        ",
+                                     "                         A                         ")));
```
stkent: 💎

matthewmorgan: 👍 for those two additional tests.  I'm in favor of opening a PR following on this to make those tests canonical-- should there be a 'discussion' issue first?

^^^ `DIamondPrinter.java` should probably be `DiamondPrinter.java`, no?

stkent: I'm imagining that there should be a discussion issue, yes. Presumably x-common is still the right place for that to happen?

Regarding the file name: I thought I pushed a fix for that already. Are you seeing the incorrectly-named version still when you pull down this branch?

matthewmorgan: You're right about the filename!  I was looking at the history in the Github UI.

matthewmorgan:
Thanks @stkent !

PR #258, opened 2017-01-21
Phone number exceptions

### LINKED ISSUE #252: Phone-number: Throw exception, instead of "0000000000"?
This is more of a question than an issue, but it could spawn a workable issue.

Currently, in the phone-number exercise, if the input is invalid (less than 10 digits, 11 digits w/o starting with 1, or more than 11 digits), the expected return is "0000000000". Wouldn't it be better practice to throw some sort of InvalidArgumentException (or similar), instead of garbage data? 

This seems especially needed since there is no rule in the exercise that prohibits 000-000-0000 from being a valid phone number.

colin-mullikin:
Fixes #252. Passing invalid input to the constructor should now expect
an IllegalArgumentException.


exercises/phone-number/src/test/java/PhoneNumberTest.java
```diff
@@ -39,25 +39,15 @@ public void validWhen11DigitsAndFirstIs1() {
     }
 
     @Ignore
-    @Test
+    @Test(expected = IllegalArgumentException.class)
```
FridaTveit: We've been migrating to using JUnit's ExpectedException @Rule feature for these cases (like in your PR #248), so if you could convert these two tests to use that, that would be great :)

exercises/phone-number/src/test/java/PhoneNumberTest.java
```diff
@@ -39,25 +39,15 @@ public void validWhen11DigitsAndFirstIs1() {
...
     public void invalidWhenOnly11Digits() {
-        final String expectedNumber = "0000000000";
         final String actualNumber = new PhoneNumber("21234567890").getNumber();
```
FridaTveit: You don't need to make a variable here since you're not using it again. You could just do <code>new PhoneNumber("21234567890").getNumber()</code> :)

exercises/phone-number/src/test/java/PhoneNumberTest.java
```diff
@@ -39,25 +39,15 @@ public void validWhen11DigitsAndFirstIs1() {
...
-
-        assertEquals(
-            expectedNumber, actualNumber
-        );
     }
 
     @Ignore
-    @Test
+    @Test(expected = IllegalArgumentException.class)
     public void invalidWhen9Digits() {
```
FridaTveit: Same two comments for this test :)

matthewmorgan:
@colin-mullikin @FridaTveit When I took a look at this, I noticed that this exercise does not implement all the canonical tests.  To be clear, the example _would_ correctly handle the cases presented in the missing tests, but would there be value to including them I wonder?

@colin-mullikin the intention of your PR was not originally to bring the exercise in to compliance with the canonical tests.  On the other hand, it seems like a small leap.  Would you like to build in this functionality, or do you (or any @exercism/java !) feel that additional work should be in a separate PR?  I'm fine either way.  In my professional work I am a fan of keeping the scope of a PR logically contained, but I just thought I'd mention it.

For reference, here are a few of the tests not implemented by this track:
```
      {
        "description": "invalid with letters",
        "phrase": "123-abc-7890",
        "expected": null
      },
      {
        "description": "invalid with punctuations",
        "phrase": "123-@:!-7890",
        "expected": null
      },
      {
        "description": "invalid with right number of digits but letters mixed in",
        "phrase": "1a2b3c4d5e6f7g8h9i0j",
        "expected": null
      }
```

Anyway, I'm happy to write this up as a separate issue if it seems to far afield.

colin-mullikin:
@matthewmorgan I noticed this as well. I decided not to implement the extra tests in this PR because our solution would actually fail the last test you listed. Once you strip out all the non-digits (as we do in the example solution), that becomes a perfectly valid number.

If people decide that we should include these tests, I would be more than happy to take a stab at it, in this PR or in a separate one.

matthewmorgan:
@colin-mullikin right you are.

It sounds like it makes sense to have a separate issue to bring this into compliance with the canonical data, which as you pointed out, will mean changing the example solution as well.

I'll write that up.  Thanks for the good work!

Issue #253, opened 2017-01-19
Grade-school: Structure very unclear


colin-mullikin:
The intended structure of the School class is very hard to pin down. All of the add() calls are done directly on the School, but then for the first few tests, things (size, added students) are accessed through an arbitrary middle layer, `db()`. After the first few tests, suddenly the calls change from `school.db().get({grade number})` to `school.grade({grade number})`.

Examples:
    `school.add("Chelsea", 3);`
    `assertThat(school.db().get(3).size(), is(1));`
    `assertThat(school.grade(5).size(), is(2));`

If the point of the exercise is to force adherence to an arbitrary (and frankly ridiculous) structure, then I would consider it a great success.

FridaTveit:
I completely agree that this exercise is very confusing at the moment. I agree that the <code>db()</code> layer is not needed, and that using it only some of the time is very confusing. Having looked at the exercise in Java and in C#, I would suggest we remove that <code>.db()</code> function completely and make all the tests use the <code>.grade()</code> syntax. The C# exercise doesn't have the <code>.db</code> function and the description of the exercise doesn't really seem to ask for it. As it's only removing functionality, this shouldn't break any existing solutions.

I also find the name and behaviour of the <code>.sort()</code> function confusing. I would expect a function with that name to internally sort the School object and either return a new School object whose content is now sorted, or to modify the object the function is called on. However, this doesn't quite match what the description of the exercise asks for. As it currently returns the internal database of the School object sorted, I would expect it to be called something like <code>getAllStudentsInAllGradesSorted()</code> or <code>allStudentsInAllGradesSorted()</code>. However, making either of those changes would break existing solutions to the exercise. @colin-mullikin, @stkent, @jtigger and @matthewmorgan what do you think? 

The C# and F# exercises have a <code>roster()</code> function which returns all students in all grades (like the <code>db()</code> function). A later test then checks that this dictionary is sorted. I think that implementation makes more sense, though that still means you have two ways of getting all students in a particular grade (<code>school.db().get(grade)</code> and <code>school.grade(grade)</code>).

colin-mullikin:
I agree with pretty much everything you said. 

[...]

PR #259, opened 2017-01-21
Issue #253: Tried to improve grade school structure. 

FridaTveit:
Removed .db() function, added numberOfStudents() function and renamed the sort() function to studentsByGradeAlphabetical(). Added test to check that grade() gets the students in the order they were inserted. Removed getsStudentsInAGrade() test as what it was testing was already being covered by the tests before it. This should hopefully fix issue #253 .

### DIFF
--- a/exercises/grade-school/src/example/java/School.java
+++ b/exercises/grade-school/src/example/java/School.java
@@ -13,12 +13,15 @@ public class School {
   }
 
   public void add(String student, int grade) {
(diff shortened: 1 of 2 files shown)


FridaTveit:
Also added two tests for protection against mutation of the list object

matthewmorgan:
Great semantics @FridaTveit -- well done.  Closes #253

Issue #261, opened 2017-01-21
Bring phone-number exercise in to compliance with canonical data


matthewmorgan:
As discussed in #258 , our implementation of this exercise does not conform to the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/phone-number/canonical-data.json) in x-common.

Bringing the Java track in line on this will involve both additions to the test suite and altering the input checking in the example.

stkent:
Closed by #274!

PR #274, opened 2017-01-30
Issue #261: Bring phone-number into compliance with canonical data

FridaTveit:
Added tests for cases we hadn't covered and changed example implementation to work in those cases. This should solve issue #261 .


stkent:
This looks good to me! I notice that the final three tests are not included in the canonical data - are we ok with leaving these additional tests in the suite? (Note that they also increase the API surface of the implemented class).

FridaTveit:
Yes, I did notice and wonder about that. I don't know why they were originally introduced so I didn't want to remove them without checking. I don't mind them being there, they don't add too much difficulty to the exercise and they make it a bit more interesting. @jtigger, @matthewmorgan do you think they should be there?

jtigger:
_(Sorry for the delay, I missed the @-mention.)_

While they are interesting, I agree, they seem to fall outside the intended scope of the exercise.  The description even establishes _why_ we're scrubbing the phone number: "_... so that they can be sent SMS messages._"  That one phrase makes the `.pretty()` and `.getAreaCode()` seem like they're just hanging out for no apparent reason.

I look at it from the practitioner's point of view and pretend I'm a designer (that last part takes _a lot_ of imagination).  I ask, "how's does this exercise make the practitioner _feel_?"  Cohesion of problem (and therefore effort) makes me feel like I'm solving a whole problem; the lack thereof makes me feel like I'm being asked to do busywork.

FridaTveit:
Okay, that makes sense. I'll remove those tests then :)

stkent:
Thanks for weighing in, @jtigger, and for the follow-up changes, @FridaTveit! LGTM :)

PR #280, opened 2017-02-07
hello-world: Modify tests to reject `name == ""` comparison

otraczyk:
Hello World was not supposed to be tricky, but I think I've found an uncovered case.

In some solutions people were using `name == ""` to check for empty string, as opposed to `name.equals("")`. The `==` operator in java compares object references, not values. Currently the tests pass in such case, but it's only a coincidence, since they compare to the same instance of the String constant `""` in JVM's constant pool. For newcomers, it could be a dangerous misconception. This change avoids it.

It's my first PR, so any feedback and guidance is welcome.

I've also considered making a separate test case instead of changing current one, to clarify the problem. Do you think that would be better?

### DIFF
--- a/exercises/hello-world/src/test/java/HelloWorldTest.java
+++ b/exercises/hello-world/src/test/java/HelloWorldTest.java
@@ -8,10 +8,16 @@ public class HelloWorldTest {
 
     @Test
     public void helloNoName() {
-        assertEquals("Hello, World!", HelloWorld.hello(new String("")));
+        assertEquals("Hello, World!", HelloWorld.hello(""));
         assertEquals("Hello, World!", HelloWorld.hello(null));
     }
 
+    @Test
+    @Ignore
+    public void emptyStringIsComparedByValue() {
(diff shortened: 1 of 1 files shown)


jtigger:
Hello @otraczyk!

Thank you for taking the time to make this improvement.  These are exactly the kinds of refinements — ways to improve the learning experience — we just love.

In terms of making this a separate test... yeah, I think you're onto something there.  We wouldn't want to leave someone stuck at "wow, why *does* `==` no longer work?!".

I was thinking of suggesting a comment, but a separate test is starting to feel more appropriate to me too.  You can bake into the name of the test the lesson being conveyed.

Please feel free to make that change and we'll review.

Cheers!

otraczyk:
Hi @jtigger . Thanks for your attention!
I've added a separate test. Do you think the case is clear enough?

jtigger:
I do!  Thank you for making the adjustment, Olga.

Congratulations, you've successfully completed your first pull request!

Issue #279, opened 2017-02-07
anagram: test names and order a bit confusing?


FridaTveit:
After recently having done the anagram exercise, I found some of the test were in a slightly strange order with the names not being very consistent. Of the first few tests we have `testSimpleAnagram` and `testDetectMultipleAnagrams`. After that there are tests for other things such as duplicates, and same word is not anagram and so on. But then at the end of the class we have `testDetectAnagram` and `testMutipleAnagrams`. These are not the same as the first two (they use different words) but they are very similar which makes it confusing that they come right at the end. So my question is:
- Should we reorder the tests so that these four tests come right after each other?
- Should we try and rename some of these four exercises so that it is more clear how they are different (e.g. `testDetectSimpleAnagram`, `testDetectLongerAnagram`, `testDetectMultipleAnagrams` and `testDetectMultipleAnagramsOutOfLongerList`)?
- Should we try and be consistent about whether we use `detect` in the test names or not?

Also, the tests don't use all the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/anagram/canonical-data.json). Is this something which should be fixed?

jtigger:
- yes;
- yes;
- aaaaaand, yes!

(and yes).

:)

While we're at it, since the experience is fresh in your mind, if reordering the tests help create a more natural TDD flow.  That is: that if a prior test really forces the practitioner to _also_ address a subsequent test, better to put that second test, first (did I explain that clearly?).

FridaTveit:
Yes, I think I know what you mean :) Great, I'll make a PR for this then :)

stkent:
Resolved by #285!

PR #285, opened 2017-02-10
Change order and names of anagram tests

FridaTveit:
Changed names of tests to be consistent and changed the order so that similar tests are grouped together. Added more tests using the canonical data. I didn't add tests for everything in the canonical data because some test cases appeared to be redundant. For example, it looks like the "does not detect false positives" test case checks the same thing as "anagrams must use all letters exactly once". However, I'm happy to add tests for every case if people think I should.

See issue #279 for discussion.

### DIFF
--- a/exercises/anagram/src/test/java/AnagramTest.java
+++ b/exercises/anagram/src/test/java/AnagramTest.java
@@ -18,11 +18,19 @@ public class AnagramTest {
 
     @Ignore
     @Test
-    public void testSimpleAnagram() {
+    public void testDetectSimpleAnagram() {
         Anagram detector = new Anagram("ant");
         List<String> anagram = detector.match(Arrays.asList("tan", "stand", "at"));
         assertThat(anagram, hasItem("tan"));
         assertThat(anagram.size(), is(1));
+    }
+
+    @Ignore
+    @Test
+    public void testDetectLongerAnagram() {
+        Anagram detector = new Anagram("listen");
+        List<String> anagrams = detector.match(Arrays.asList("enlists", "google", "inlets", "banana"));
+        assertThat(anagrams, hasItem("inlets"));
     }
 
     @Ignore
@@ -35,10 +43,17 @@ public class AnagramTest {
 
     @Ignore
     @Test
+    public void testDetectMultipleAnagramsForLongerWord() {
+        Anagram detector = new Anagram("allergy");
+        List<String> anagrams = detector.match(Arrays.asList("gallery", "ballerina", "regally", "clergy", "largely", "leading"));
+        assertThat(anagrams, allOf(hasItem("gallery"), hasItem("largely"), hasItem("regally")));
+    }
+
+    @Ignore
+    @Test
     public void testDoesNotConfuseDifferentDuplicates() {
         Anagram detector = new Anagram("galea");
-        List<String> anagrams = detector.match(Arrays.asList("eagle"));
-        assertTrue(anagrams.isEmpty());
+        assertTrue(detector.match(Arrays.asList("eagle")).isEmpty());
     }
 
     @Ignore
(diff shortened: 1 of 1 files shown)


exercises/anagram/src/test/java/AnagramTest.java
```diff
@@ -35,10 +43,17 @@ public void testDetectMultipleAnagrams() {
...
+        assertThat(anagrams, allOf(hasItem("gallery"), hasItem("largely"), hasItem("regally")));
+    }
+
+    @Ignore
+    @Test
     public void testDoesNotConfuseDifferentDuplicates() {
         Anagram detector = new Anagram("galea");
-        List<String> anagrams = detector.match(Arrays.asList("eagle"));
-        assertTrue(anagrams.isEmpty());
+        assertTrue(detector.match(Arrays.asList("eagle")).isEmpty());
```
stkent: In the canonical tests this check appears before the multiple anagrams tests. Should we match that order here, or propose a change to the canonical data?

exercises/anagram/src/test/java/AnagramTest.java
```diff
@@ -52,6 +67,20 @@ public void testIdenticalWordIsNotAnagram() {
 
     @Ignore
     @Test
+    public void testCapitalWordIsNotAnagram() {
```
stkent: I think naming this `testCapitalWordIsNotOwnAnagram ` would match the canonical test description ("capital word is not own anagram") slightly better.

exercises/anagram/src/test/java/AnagramTest.java
```diff
@@ -52,6 +67,20 @@ public void testIdenticalWordIsNotAnagram() {
...
+        Anagram detector = new Anagram("BANANA");
+        assertTrue(detector.match(Arrays.asList("Banana")).isEmpty());
+    }
+
+    @Ignore
+    @Test
+    public void testIdenticalWordRepeatedIsNotAnagram() {
```
stkent: Appears before "capital word is not own anagram" in the canonical data; should we match or propose a change in x-common?

stkent:
@FridaTveit a few comments in! We deviate in order quite a bit from x-common; I feel like we should either match the standard ordering _or_ propose a change to the x-common ordering with justification. Which do you think is the right choice for this exercise?

jtigger:
I have always assumed that the scope of the "contract" with `config.json` and tracks only involved the contents of it and not the ordering.  No one told me this.  I didn't read it anywhere.  I just assumed it.

I figured that a given language's features (e.g. default values for arguments, error flows, available standard libraries) would make one path or another flow better.

I have gotten the impression, over the couple years I've been involved, that Track Maintainers have a good amount of autonomy; some argue too much (e.g. there are apparently a whole set of Haskell-oriented exercises in x-common).

I think the appropriate thing is what affords the best UX for the practitioner.  Here, balancing the immediate feel of the Java variation of this exercise with the general feel that all variations of the exercise having the same scope.

Thoughts?

FridaTveit:
I don't mind too much about the ordering. I found it very confusing as it was since there were two pairs of tests that had nearly exactly the same name but were far apart. Now that their names are slightly different it's less confusing. I still think it makes sense to group the two multiple anagrams tests together but it wouldn't bother me too much if there is another test between them.

Either way it would be nice for future reference to know whether the ordering of the tests in the canonical data is something we have to stick to or whether it's just a suggestion.

exercises/anagram/src/test/java/AnagramTest.java
FridaTveit: I think it makes sense to group the two multiple anagrams tests together but I don't mind too much.

exercises/anagram/src/test/java/AnagramTest.java
FridaTveit: Sure, will change it :)

exercises/anagram/src/test/java/AnagramTest.java
FridaTveit: I can move it :)

stkent:
Good to know r.e. the ordering! I've always tried to match the canonical ordering, but the viewpoint that we are free to reorder depending on what makes sense in our language of choice seems eminently reasonable :)

exercises/anagram/src/test/java/AnagramTest.java
stkent: I'm down with that!

exercises/anagram/src/test/java/AnagramTest.java
stkent: Thanks; I think in the case where there's not a compelling reason to deviate from x-common data then that should be our default :)

stkent:
@FridaTveit replied; once the changes I confirmed as requests are made, LGTM!

stkent:
Nice work, @FridaTveit!

PR #288, opened 2017-02-12
luhn: fix example solution

stkent:
Previously this code was applying the required operations to every
second digit starting from the LEFT. The correct solution applies the
required operations to every second digit starting from the RIGHT. The
fact that this was not caught by the tests is a problem! The canonical
tests have been updated since this exercise was last overhauled, so next
steps should be to update to match the expanded canonical test suite,
then test against a broken implementation to see if it is now caught.
Even if it is, it may be worth including an explicit left vs right test
in the canonical suite if one does not already exist.


stkent:
Closing in favor of #289

jtigger:
Nice catch, @stkent!

stkent:
Thanks, @jtigger! I'm continuing to work on this problem upstream to help steer others away from my own fate 😅

PR #289, opened 2017-02-12
luhn: update to latest canonical tests

stkent:
https://github.com/exercism/x-common/blob/134e4970dfa3f1cee7e054ddcacf694609f19290/exercises/luhn/canonical-data.json

Also includes some adjustments to the example solution to fix broken edge cases.

### DIFF
--- a/exercises/luhn/src/test/java/LuhnValidatorTest.java
+++ b/exercises/luhn/src/test/java/LuhnValidatorTest.java
@@ -19,8 +19,20 @@ public class LuhnValidatorTest {
 
     @Ignore
     @Test
-    public void testThatAValidCanadianSocialInsuranceNumberIsIdentifiedAsValid() {
+    public void testThatASimpleValidCanadianSocialInsuranceNumberIsIdentifiedAsValid() {
+        assertTrue(new LuhnValidator().isValid(" 5 9 "));
+    }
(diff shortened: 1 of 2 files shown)


exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -19,12 +19,24 @@ public void testThatTheStringConsistingOfASingleZeroIsInvalid() {
 
     @Ignore
     @Test
-    public void testThatAValidCanadianSocialInsuranceNumberIsIdentifiedAsValid() {
+    public void testThatASimpleValidCanadianSocialInsuranceNumberIsIdentifiedAsValid() {
```
FridaTveit: Not that it matters very much, but I don't think this is a valid canadian SIN as it only has two digits. In the canonical data this test is called "simple valid sin" so maybe change this test name to `testThatASimpleValidSocialInsuranceNumberIsIdentifiedAsValid()`?

stkent: Good catch; will update, thanks!

stkent:
@FridaTveit updated!

FridaTveit:
Great, thanks! :)

PR #293, opened 2017-02-13
Migrate to JUnit @Rule in triangle exercise [Fix #268]

### LINKED ISSUE #268: triangle: use JUnit's @Rule feature
We've been migrating to using JUnit's [ExpectedException](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) `@Rule` feature instead of `@Test(expected = SomeException.class)`. Change the occurrences of `@Test(expected = SomeException.class)` in TriangleTest to make them use the `@Rule` feature instead.

michaelspets:


exercises/triangle/src/test/java/TriangleTest.java
```diff
@@ -1,11 +1,14 @@
...
+import org.junit.Rule;
+import org.junit.rules.ExpectedException;
 
 import static org.junit.Assert.assertEquals;
 
 public class TriangleTest {
 
-
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
```
FridaTveit: I would add a new line here to make it more readable

exercises/triangle/src/test/java/TriangleTest.java
```diff
@@ -86,32 +89,37 @@ public void verySmallTrianglesAreLegal() throws Exception {
     }
 
     @Ignore
-    @Test(expected = TriangleException.class)
+    @Test
     public void trianglesWithNoSizeAreIllegal() throws Exception {
```
FridaTveit: I know you didn't add this, but it would be great if you could change `throws Exception` to `throws TriangleException` as it's usually best to be as specific as you can :)

michaelspets:
You 're welcome!

Issue #283, opened 2017-02-09
palindrome-products: simplify construction logic


jtigger:
Currently, `palindrome-products` is using the builder pattern-style interface to create a list of palindromes.

Looks like we could get away with a simpler means of doing this.  See #232 for the relevant discussion.

FridaTveit:
Closed by PR #305.

PR #305, opened 2017-02-15
Simplify construction logic for palindrome products

FridaTveit:
Fixes issue #283

### DIFF
--- a/exercises/palindrome-products/src/example/java/Palindromes.java
+++ b/exercises/palindrome-products/src/example/java/Palindromes.java
@@ -1,77 +1,43 @@
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.*;
 
 public class Palindromes {
 
-    public static class PalindromeSortedSetBuilder {
-
-        private int minFactor = 1;
-        private int maxFactor = 1;
-
-        public PalindromeSortedSetBuilder() {
-        }
-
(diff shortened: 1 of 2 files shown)


matthewmorgan:
Hey @FridaTveit this looks great.

I have two (small) suggestions, and the first might be something that the group has decided to do without me noticing, so if that's the case, just LMK:

1. I think it's better to explicitly declare your dependencies in your imports rather than use `*`.  It looks like this change was deliberate in the example code, so if that's the preferred style, I understand.
2. Looks like the comparator used in all the tests is the same.  Maybe that could be factored out to a named function?

I don't think either of these things is worth holding up the PR over, personally.  But let me know your thoughts.

Thanks!

FridaTveit:
Thanks for the comments! :)

1. This wasn't exactly a deliberate change, I just used auto format (ctrl + alt + L) on the file in Intellij. I don't particularly mind but I tend to go with what auto format does as that makes life easier. But if you want me to change it to explicitly declare them then I'm happy to do that :)

2. Great idea! I'd be happy to make that change :)

matthewmorgan:
I don't have a **strong** feeling about it, but if there's a preferred style, we should use that, for sure.  It's nice not to see lines and lines of boilerplate, but sometimes you can end up importing the wrong class that way without realizing it.  @exercism/java ?   What's the house position on `*` imports vs specific?

stkent:
I am ambivalent; in my personal code I have no problem with `*` imports, as IntelliJ's handling of auto-importing and potential conflicts is so good.

jtigger:
@stkent: jinx!! (again!)

matthewmorgan:
OK, that's good enough for me!

Issue #306, opened 2017-02-15
nucleotide-count: fix spelling mistakes in test names


FridaTveit:
Change the test names in NucleotideTest so that they use the correct nucleotide names: adenine, cytosine, guanine, and thymine.

FridaTveit:
Fixed in PR #307. Thanks @extrika! :)

extrika:
You're welcome! My first contribution to an open source project!

stkent:
@extrika that's awesome! :)

jtigger:
Congratulations, @extrika!

Now that you're on _this_ side of the experience, we'd love to hear about your experience in getting here.  Are there parts of any instructions that were wrong or confusing?  Did you have to overcome any challenges outside of the code fix itself?

extrika:
I guess the challenge for a beginner is becoming comfortable with git and GitHub. I used the HelloWorld guide from  GitHub to become familiar with the 'fork/branch/commit/pull request" workflow. The documentation from the exercism repo is clear and I must say that I'm impressed by the human contact! I love the exercism project and I look forward to contributing in any way I can.

extrika:
Any tips or common mistakes to avoid for a beginner?

jtigger:
Thanks for your reflections @extrika!  That is valueable to know!

Personally, I think you've already avoided the big mistakes: trying too hard to make the aubmission perfect without review; conversely, not caring about the details.

extrika:
Thanks!

PR #307, opened 2017-02-15
Issue #306: fix spelling mistakes in test names

extrika:


exercises/nucleotide-count/src/test/java/NucleotideTest.java
```diff
@@ -42,7 +42,7 @@ public void testRepetitiveCytidineGetsCounted() {
```
FridaTveit: Looks like you missed `testRepetitiveCytidineGetsCounted` :)

extrika: Sorry about that! I'll fix it right away.

PR #309, opened 2017-02-16
Build jq from source.

jtigger:
- we need version 1.5 and can't get it from the official repo.
- includes adjusting the PATH so that our tools (like jq) get
  precedence.


stkent:
LGTM!

PR #311, opened 2017-02-17
isogram : add to track

Kai-Shiro:
I'll implement the isogram exercise
https://github.com/exercism/x-common/tree/master/exercises/isogram


jtigger:
Renamed to make clear the status of this PR.  @Kai-Shiro, when you're ready for a review, please rename it back to `bookstore : add to track` or the like.  Thanks!

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,29 @@
+package exemple;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class Isogram {
+	
+	public static boolean isIsogram(String word){
+		
+		Set<Character> charset= new HashSet<Character>();
```
matthewmorgan: You can just use the [diamond operator](http://stackoverflow.com/questions/4166966/what-is-the-point-of-the-diamond-operator-in-java-7) for your `HashSet` declaration like this:

`Set<Character> charset = new HashSet<>();`

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,29 @@
...
+		String[] words = word.split(" ");
+		String newword = concat(words);
```
matthewmorgan: In Java, the convention is to use `camelCase`, as in `newWord`.

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,29 @@
...
+		for(int i = 0; i < newword.length(); i++)
+			charset.add(newword.charAt(i));
+		
+		return charset.size() == newword.length();
+	}
+	
+	public static String concat(String[] words){
+		String newword = new String("");
+		for(int i = 0; i < words.length; i++)
+			newword = newword + words[i];
```
matthewmorgan: If you like, you could replace this with:
```Java
String newWord = "";
for (String word: words){
    newWord += word;
}
```
No need for the index variable.

Or, for some extra-delicious Java 8:

```Java
String newWord = stream(words).collect(joining());
``` 

To use the `Streams` version you'll need to import:

```Java
import static java.util.Arrays.stream;
import static java.util.stream.Collectors.joining;
```

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,29 @@
...
+	public static boolean isIsogram(String word){
+		
+		Set<Character> charset= new HashSet<Character>();
+		String[] words = word.split(" ");
+		String newword = concat(words);
+		words = newword.split("-");
+		newword = concat(words).toLowerCase();
+		
+		for(int i = 0; i < newword.length(); i++)
+			charset.add(newword.charAt(i));
```
matthewmorgan: Usually we like to be explicit with our curly braces, even in short blocks like this.

matthewmorgan:
@Kai-Shiro thanks for doing this!  We really appreciate the work.

in the spirit of challenging each other to produce the best possible product, I have made a couple of suggestions that are primarily stylistic.  I also threw out a couple of alternative implementations that I thought might be interesting to you.  Please feel free to use them, or not, as you prefer.

Finally, there's a merge conflict with `config.json` that was created when we merged some older PRs.  Would you mind fixing that, or letting me know if you would like help doing it?

Thanks again.

Kai-Shiro:
@matthewmorgan Thanks for the advices! I have applied some changes to the code and i think i have fixed the merge conflict with <code>config.json</code>

matthewmorgan:
Nice work @Kai-Shiro !  This implementation is looking much cleaner-- concise, yet still easy to read.

If you're up for it, I would like to suggest a couple more changes.

1. We're trying to move away from static methods for problems like this.  Please see the discussion about the `Pangram` exercise here: https://github.com/exercism/xjava/pull/344  Two things to really pay attention to in the above discussion are the conversation around the semantics of the class and method names, and the preference for instance methods over class methods.

2. It doesn't appear that your `concat` method needs to be `public`.  Since it's only used internally by the Isogram class, you can (and should) make it `private`.

3.  Small nit: the `charset` variable should be camelCased.

Please let me know if you have any questions, or if you would like help with any of the details.

Kai-Shiro:
@matthewmorgan I have applied some changes.
- There is no longer statics methods
- I change <code>charset</code> to <code>charSet</code>
- I change the code to use instance methods

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,37 @@
+package exemple;
+
+import java.util.HashSet;
+import java.util.Set;
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.joining;
+
+public class Isogram {
```
matthewmorgan: I am thinking that the semantics are very close on this class, but not quite right.  Think of it this way: what is the purpose of this class?  What do the methods do?  What if, for example:
```Java
IsogramChecker isogramChecker = new IsogramChecker();
String word = "isogram";
boolean result = isogramChecker.isIsogram(word);
```

exercises/isogram/src/test/java/IsogramTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class IsogramTest {
+
+	@Test
+	public void testIsogram() {
+		Isogram iso = new Isogram("duplicates");
+		assertTrue(iso.isogramChecker());
+	}
+	
+	@Test
```
matthewmorgan: You'll want to annotate all tests except the first one with `@Ignore`, so the users can explore their solution using each test as a goalpost.

exercises/isogram/src/example/java/Isogram.java
```diff
@@ -0,0 +1,37 @@
+package exemple;
```
Kai-Shiro: Oops, french mistake!

matthewmorgan:
Nicely done!

PR #315, opened 2017-02-20
pangram: implement missing test [Fix #314]

### LINKED ISSUE #314: pangram: include missing test
There is one test case in the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json) for pangram which is not currently implemented; the one with description "upper and lower case versions of the same character should not be counted separately". Add this test to PangramsTest. The input and expected output can be found in the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json).

michaelspets:


exercises/pangram/src/test/java/PangramsTest.java
```diff
@@ -66,4 +66,10 @@ public void pangramWithNonAsciiCharacters() {
     public void panagramInAlphabetOtherThanAscii() {
         assertFalse(Pangrams.isPangram("Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства."));
     }
+
+    @Ignore
+    @Test
+    public void upperAndLowerCaseVersionsOfTheSameCharacterShouldNotBeCountedSeparatelyIsNotPangram() {
```
FridaTveit: Sorry to be very picky and possibly confusing, but I don't think this one needs `IsNotPangram` at the end. The other tests are specifying that some kinds of phrases are or aren't recognized as pangrams, e.g. pangram in another alphabet is not recognized as a pangram whereas an all lowercase pangram is. However, this test is checking for a common error in the implementation, i.e. counting upper and lowercase letters as different letters. Therefore I think `upperAndLowerCaseVersionsOfTheSameCharacterShouldNotBeCountedSeparately` is an okay name :) I hope that makes sense! :)

stkent:
Trivia: this resulted from my [very first Exercism contribution of note](https://github.com/exercism/xkotlin/pull/10) over in xkotlin! ♻️

exercises/pangram/src/test/java/PangramsTest.java
michaelspets: Ok I can see what you meen but I can't understand what we actually test in this case. The test is supposed to check if same chars are counted seperately but I don't see how we check that in this test... For example we have "pangram with mixed case and punctuation" test case that checks if a pangram with upper and lower case chars is considered true by counting upper and lowercase letters but I think that we don't count anything here. So my question is "Are we actually test anything with this test?"

Issue #178, opened 2016-11-18
[policy] Include starter implementations for the first X exercises so that new-to-the-language practitioners are driven by test failures, not compilation errors.


rstockbridge:
I am new to Java but have prior programming experience. The problems I have completed so far on the Java track either have either come with a nearly blank starter implementation, or no starter implementation at all.

For my code to be compilable (and therefore testable), I have had to reverse engineer all the tests at once to deduce method signatures and 3rd party dependencies, even if most tests are ignored to begin with. It was quite challenging to infer some of the return values given that I am a beginner in Java.

It would be really helpful if the starter implementations contain sufficient information to be compilable, i.e. they contain required methods with appropriate signatures and dummy return values, and import statements. This way the focus can be on writing code that passes one test at a time.

jtigger:
First, thank you for taking the time to provide this feedback, @rstockbridge.

This topic has also come up in other language tracks: exercism/xrust#117, exercism/xhaskell#181.  There's been a mix of opinions.

In exercism/xrust#117:

@IanWhitney said
> For me, I like not having the signature. I'm entirely new to Rust and static typing and I've found that I like puzzling out what I have to implement by looking at the tests.

@kytrinyx said:
> I'm also on the fence. In the Go track we've provided stub files early on, and then we phase them out after the first few. For the most part reverse engineering function signatures doesn't seem to be difficult in Go ... I can't speak to how much busywork that adds here.

...


A proposal incorporating your request with those observations:

1. the first ~10 exercises, we provide complete stubs.
- the subsequent ~10 exercises, we stub a portion of the interface and comment-out the bodies of the tests that cover the other part.
- After that, include no starter code.
  - that said, if there are "interesting" interface aspects, mention these in the `HINT.md` file (gets merged into the `README.md` for the exercise).

Thoughts?

rstockbridge:
This proposal sounds excellent 😄 - a good balance of helping beginners get started while still leaving some things to be figured out. Obviously it makes most sense to implement this after #142 is complete.

jtigger:
We've been using the new GitHub Projects feature to express priorities: https://github.com/exercism/xjava/projects/2

I've added this to our "Epics" column... once we determine the (re)ordering of the track, we'll break that down into individual user stories according to the proposal.

rstockbridge:
Awesome, sounds good!

jtigger:
I believe this qualifies as a "policy" — a decision that affects future work.  Labeled accordingly.

[...]

Issue #177, opened 2016-11-17
[policy] Prefer object-based APIs (instance methods) over class-based (static methods)


stkent:
Some exercise test suites force the use of static methods exclusively (e.g. `nth-prime`); others force the use of objects (e.g. `anagram`). Is there a preference for one paradigm vs the other? Having users switch back and forth between the two while working through exercises seems unnecessarily confusing.

jtigger:
This topic (AFAIK) has not been discussed in the broad scope you're framing, @stkent.  I see one "discussion" in #90.

**tl;dr:** most (all?) of these exercises should be implemented in the form of instance methods since they contain "domain logic" and we (Exercism) want to encourage exemplary software.

An argument:

1. A programmer participating in Exercism should be exposed to our best understanding of how to design "good" software.
- "Good" software design maximizes the profit of a product over its lifetime (i.e. minimize maintenance and development costs while maximizing value delivered through its use).
- As a product increases in complexity, so does the risk of maintenance costs:
  - as complexity increases, so does the risk of defects;
  - as complexity increases, so does the difficulty in isolating defects;
  - the effort required to isolate and properly address defects is a maintenance cost.
- Such risk can be materially mitigated by co-developing a testing system alongside the product such that this testing system:
 - exercises the product in ways that is likely to expose defects in essential logic (colloquially, "domain logic");
 - is used frequently so as to expose such defects as soon as possible (read: executes expeditiously);
 - is maintained along with the product.
- For such a testing system to be feasible, the product itself must be testable:
  - individual components of the product should be easy to isolate so as to exercise them sufficiently to gain confidence that they each behave as desired in a given context.
- To achieve that isolation, the product's components must support at least two flavors of its dependencies: at least one production implementation and one test double implementation.
- In Java, the least surprising way to provide two implementations for the same behavior is to use virtual functions.
- In Java, instance methods are, by default, virtual functions; static methods are not.

jtigger:
There are a number of maintainers here.  @exercism/java, care to chime in?

@stkent, what's your point of view?

matthewmorgan:
@jtigger @stkent i agree with John in principle.  I do think it's useful if we can to provide some exposure to static methods as a feature of the language.

In the ecmascript track we had a similar issue with the use of class based solutions versus functional or object literal solutions. We ended up setting an "on boarding" pattern of requiring classes for the first dozen exercises, then mixed it up after that.

Having said all that, I think the goal of promoting good practices should take priority over exploring all language features.

[...]

PR #334, opened 2017-02-28
rna-transcription: make object-based and add starter implementation

FridaTveit:
Changed rna-transcription to have an object based implementation (see issue #177) and added starter implementation as it is currently the second exercise in the new ordering (see issue #178).

### DIFF
--- a/exercises/rna-transcription/src/example/java/RnaTranscription.java
+++ b/exercises/rna-transcription/src/example/java/RnaTranscription.java
@@ -1,6 +1,6 @@
 public class RnaTranscription {
 
-    public static String ofDna(String strand) {
+    public String ofDna(String strand) {
         StringBuilder sb = new StringBuilder();
         for (char c : strand.toCharArray()) {
             switch (c) {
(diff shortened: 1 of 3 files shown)


stkent:
Looks good! I'm not sure whether I prefer returning `null` vs throwing an unimplemented exception in the stubbed implementation - I feel like `null` could in general lead to a confusing error message when the tests are first run, and that the thrown exception route gives us more control over that messaging and experience. What do you think, Frida? I don't feel particularly strongly either way so I'm happy to go with whatever you decide, just let me know :)

FridaTveit:
I didn't think about that, thanks for mentioning it! :) An exception is probably better. Do you think `UnsupportedOperationException` with message "Method has not been implemented yet" would be okay?

stkent:
@FridaTveit that exception type sounds good to me, and to [others](http://stackoverflow.com/questions/829136/what-is-the-standard-exception-to-throw-in-java-for-not-supported-implemented-op)!

stkent:
Thanks, @FridaTveit!

PR #336, opened 2017-03-01
perfect-numbers: update to canonical tests v1.0.0

stkent:
perfect-numbers [now has canonical data](https://github.com/exercism/x-common/blob/19d0c7714ce664a1ad762af624c17f8e269fa8b2/exercises/perfect-numbers/canonical-data.json). This PR updates the xjava tests to match this new data. Note too the comment at the top of the file which specifies the version of the tests being implemented. Does this seem like a reasonable practice moving forward?

### DIFF
--- a/exercises/perfect-numbers/src/example/java/NaturalNumber.java
+++ b/exercises/perfect-numbers/src/example/java/NaturalNumber.java
@@ -5,7 +5,7 @@ final class NaturalNumber {
     private final int naturalNumber;
 
     NaturalNumber(int naturalNumber) {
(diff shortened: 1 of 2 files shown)


stkent:
I wonder if we could drop that version metadata somewhere more useful, and then eventually be able to generate a list of exercises that need updates for newer canonical data? 🤔

matthewmorgan:
@stkent did you have someplace in mind for test version data?  I like the idea in principle, seems like it might be good to have an automated way to collect and present that data.  It might be too much of a chore to do it manually.

stkent:
Maybe a `.version` file that contains the canonical test version string only, inside the test dir? I'm not sure if that would pollute participant's clones though.

PR #337, opened 2017-03-01
Optimize track icon

stkent:
69% savings! From ~7kb down to ~3kb, losslessly (preserving any color profile information too), via [ImageOptim](https://imageoptim.com/mac).


matthewmorgan:
@stkent ++

PR #339, opened 2017-03-03
allergies: Update to canonical data

rohit1104:
Test case - **allergic to eggs in addition to other stuff** is incomplete.
Checks for shellfish and strawberries have been added.


rohit1104:
😄

PR #340, opened 2017-03-03
anagram: Update to canonical data

rohit1104:
Add test cases - **does not detect a word as its own anagram** and
**anagrams must use all letters exactly once**


matthewmorgan:
Lovely.  Thanks @rohit1104 !

PR #380, opened 2017-03-08
Acronym: adjust API to use an instance method

FridaTveit:
This time I decided to go with a Value Object (see #344 for discussion) because I think it makes sense to construct an acronym object from a sentence and have a method on that object to get the acronym as a string.

But if other people disagree I could instead make the class an AcronymGenerator with a generateAcronym method :) 

Fixes issue #351.


jtigger:
Sorry for the long delay, here, @FridaTveit.

I'm comparing the situation for `acronym` against that of `pangram` (where we went the other way; #344).  The difference I can see is that in the `pangram` case we're digesting a property of a given sentence whereas in `acronym`, we're calculating an entirely new value.  Is that how you're thinking?

I'm using the feel of the API as a kind of data point:

```java
new AcronymGenerator().acronymFor("...");
```

vs.

```java
new Acronym("...").get();
```

exercises/acronym/src/example/java/Acronym.java
```diff
@@ -2,7 +2,17 @@
...
 
 public class Acronym {
-    public static String generate(String phrase){
+    private final String acronym;
+
+    public Acronym(String phrase) {
+        acronym = generateAcronym(phrase);
+    }
+
+    public String getAcronym() {
```
jtigger: 🤔  `get()` vs. `getAcronym()`?

At this point, there is exactly one value in the object.  Cherry-picking [`ThreadLocal#get`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html#get--) as an example from the JDK.

FridaTveit: I agree, `get()` is probably better :)

FridaTveit:
Yes, that was what I was thinking :)

jtigger:
Thanks for the extra thought in this, @FridaTveit!  Looks lovely! :)

PR #383, opened 2017-03-13
leap :add to track 

Zaldrick:
Based from the C# exercise, I've implemented the leap exercise for java.
https://github.com/Zaldrick/xjava/tree/master/exercises/leap


Zaldrick:
Didn't understand the meaning of "foredone". Sorry for that useless PR.

PR #384, opened 2017-03-16
poker : add to track

Zaldrick:
Inspired by Poker C# exercice, I made a Java version.


exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.stream.Collectors;
+
+public class Poker {
+	private static class Card {
+		private int rank;
+		private int Suit;
+
+		public Card(String card) {
+			rank = Parserank(card);
+			Suit = ParseSuit(card);
```
FridaTveit: If `parseRank` and `parseSuit` are only used in this class, then they should be moved inside this class. That way each class is responsible for its own behaviour :)

exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class Poker {
+	private static class Card {
+		private int rank;
+		private int Suit;
```
FridaTveit: Variables normally start with a lowercase letter in java, so `suit` instead of `Suit` :)

exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+
+	private static class Hand {
+		private String input;
+		private int score;
+
+		public Hand(String hand) {
+			this.input = hand;
+			this.score = ScoreHand(ParseCards(hand));
```
FridaTveit: Same goes for these two methods. `scoreHand` and `parseCards` are only used in this class therefore their implementations should be moved inside this class.

Zaldrick:
Thank you for your review , I made indeed some newby errors.
Change will be done this evening.

exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class Poker {
```
FridaTveit: This class is looking a lot better now! :) But it is quite long with the two nested classes, Card and Hand, in it. I would suggest extracting these into their own files. That would make each file more readable :)

exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,142 @@
...
+			private int getScore() {
+				return score;
+			}
+
+			private String getInput() {
+				return input;
+			}
+
+			private static List<Card> parseCards(String hand) {
+				ArrayList<Card> parseCards = new ArrayList<Card>();
```
FridaTveit: To avoid confusion with the method name, maybe name this variable `parsedCards` instead?

exercises/poker/src/example/java/Poker.java
```diff
@@ -0,0 +1,142 @@
...
+				String[] parseString = hand.split(" ");
+				for (int i = 0; i < parseString.length; i++) {
+					parseCards.add(new Poker.Card(parseString[i]));
+				}
+				return parseCards;
+			}
+
+			private static int scoreHand(List<Card> cards) {
+
+				List<Card> cardsByrank = new ArrayList<Card>();
```
FridaTveit: Remember to use camelCase for variable names as well, so `cardsByRank` in this case :)

exercises/poker/src/main/java/Poker.java
```diff
@@ -0,0 +1,12 @@
+import java.util.List;
+
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
+
+public class Poker {
+
+	public static List<String> bestHands(List<String> hands) {
+		{
```
FridaTveit: You don't need these extra curly brackets :)

exercises/poker/src/test/java/PokerTest.java
```diff
@@ -0,0 +1,181 @@
...
+    public void OneHand()
+    {
+        final String hand = "4S 5S 7H 8D JC";
+        assertEquals(Arrays.asList(hand),Poker.bestHands(Arrays.asList(hand)));
+    }
+
+    @Test
+    @Ignore
+    public void NothingVsOnePair()
+    {
```
FridaTveit: To be consistent, maybe move this curly bracket to after the signature? So it becomes
`public void NothingVsOnePair() {` instead?

Zaldrick:
Thanks you again for your comments ! Also I take the time to think about some improvements on my side too. 
About making Card and Hand classes, I was thinking about it at the beginning, but as far as I saw in others exercices, the solution was always on one file. So basically I tried to put everything in one file even if I know it's cleary against the Java policy haha.

Also I really think that my method Hand::scoreHand is too big. I tried to reduce it a little but now I don't really know how to reduce it more. Maybe by adding some attributs to the class making able to divide into more method ? I don't know but if you can give me some tips, it would be great :)

exercises/poker/src/example/java/Poker.java
```diff
@@ -1,139 +1,14 @@
...
-				Collections.sort(ranks, (a, b) -> b.compareTo(a));
-				Collections.reverse(ranks);
-				return ranks.get(0);
-			}
-		}
 			public static List<String> bestHands(List<String> hands) {
 				ArrayList<Hand> scoredHands = new ArrayList<Hand>();
 
-				for (String s : hands) {
+				for (String s : hands)
```
FridaTveit: It's best practice to keep the curly brackets in this case. Otherwise it's easy to add another line to the for loop and forget to add the curly brackets then :)

config.json~
```diff
@@ -0,0 +1,346 @@
+{
```
FridaTveit: You still need to remove this file :)

exercises/poker/src/test/java/PokerTest.java
```diff
@@ -0,0 +1,162 @@
+import static org.junit.Assert.assertEquals;
+import java.util.Arrays;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class PokerTest
+{
```
FridaTveit: This class is looking really good now! :) I would just move this curly bracket to be consistent, so that it's `public class PokerTest {`. If you're using Intellij you could do ctrl + alt + L (on Windows, cmd + alt + L on a mac) which will reformat your code for you. That should fix spacing and formatting issues like this, and you can tell it to optimise your imports as well. [This page](https://www.jetbrains.com/help/idea/2016.3/reformatting-source-code.html) explains it more. I use it all the time because I can never remember to do stuff like remove unused imports manually :)

exercises/poker/src/example/java/Hand.java
```diff
@@ -0,0 +1,106 @@
...
+		Map<Integer, Integer> frequencymap = getFrequencyMap(cards);
+		List<Integer> ranks = frequencymap.entrySet().stream().map(x -> x.getKey())
+				.sorted(Comparator.reverseOrder()).collect(Collectors.toList());
+
+		frequencymap = frequencymap.entrySet().stream()
+				.sorted(Map.Entry.comparingByValue(Collections.reverseOrder())).collect(Collectors
+						.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+		List<Integer> rankCounts = frequencymap.entrySet().stream().map(x -> x.getValue())
+				.collect(Collectors.toList());
```
Zaldrick: thanks for this.
I also improve the other list in the same way :
List<Integer> ranks = new ArrayList<>(frequencyMap.keySet());

exercises/poker/src/example/java/Hand.java
```diff
@@ -0,0 +1,106 @@
...
+		cardsByRank = cards.stream().sorted(Comparator.comparing(Card::getRank)).unordered()
+				.collect(Collectors.toList());
+
+		Map<Integer, Integer> frequencymap = getFrequencyMap(cards);
+		List<Integer> ranks = frequencymap.entrySet().stream().map(x -> x.getKey())
+				.sorted(Comparator.reverseOrder()).collect(Collectors.toList());
+
+		frequencymap = frequencymap.entrySet().stream()
+				.sorted(Map.Entry.comparingByValue(Collections.reverseOrder())).collect(Collectors
+						.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
```
Zaldrick: I use a LinkedHash because it maintains insertion order of elements. I know that it requires more memory, but I tried to use HashMap and TreeMap. Both didn't give me the right order.

I also tried your suggestions but i've go the same probleme that I've got with HashMap.
Sometime i've got the right order, sometime not. The twoSquares test won't work with Hash or your suggestions.
Example :
"2S 2H 2S 8D 2H";
"4S 5H 5S 5D 5H";

Before ordering the frequencyMap  :
{2=4, 8=1}
{4=1, 5=4}
After :
{2=4, 8=1}
{4=1, 5=4}

Nothing change ... It seems to use the key to order even with 
.sorted(Map.Entry.comparingByValue())
or 
.sorted(Map.Entry.comparingByValue(Collections.reverseOrder())
(i've got the same results btw, it's weird ....)

What i want is :
{2=4, 8=1}
{5=4, 4=1}

Do you have any suggestions ?

About getFrequencyMapInDescendingOrder, I had a reason for not doing this initially, but can't remembre. Anyway I change it and everything keeps working, so thanks !

exercises/poker/src/example/java/Hand.java
```diff
@@ -0,0 +1,106 @@
...
+
+		List<Integer> rankCounts = frequencymap.entrySet().stream().map(x -> x.getValue())
+				.collect(Collectors.toList());
+
+		List<Integer> suits = cards.stream().map(s -> s.getSuit()).collect(Collectors.toList());
+
+		if (ranks.equals(Arrays.asList(14, 5, 4, 3, 2))) {
+			ranks = Arrays.asList(5, 4, 3, 2, 1);
+		}
+
```
Zaldrick: working on it. But I also tried to make more instance method, so I need time to figure out how to implements that.

exercises/poker/src/test/java/PokerTest.java
Zaldrick: Thank you again for your suggestions. Sorry for taking so long time to answer, I wasn't available and  It's a little bit harder to fix so I need more time haha. I started but I think it's not enough for making a commit. I guess I will finish that this weekend.

exercises/poker/src/example/java/Hand.java
```diff
@@ -0,0 +1,106 @@
...
+		boolean flush = suits.stream().distinct().count() == 1;
+		boolean straight = ranks.stream().distinct().count() == 5 && ranks.get(0) - ranks.get(4) == 4;
+		Iterator<Integer> itr = frequencymap.keySet().iterator();
+		int pairA = itr.next();
```
Zaldrick: Damn why I created a iterator and didn't use it after.
I fix it by remplacing every "frequencymap.keySet().iterator().next()" with pairA.
I renamed pairA, it wasn't a explicite name.
Not sure about using a another int for stocking secondHighestFrequency because I will use only if I got two pairs. Can't I just use itr.next() ?
I understand that using two variables will make the code more clear anyway.

exercises/poker/src/example/java/Hand.java
FridaTveit: If there's a reason you're doing it then that's fine, I just wanted to check :)

Great! :)

exercises/poker/src/example/java/Hand.java
FridaTveit: That's fine, take your time, there's no rush :)

exercises/poker/src/example/java/Hand.java
FridaTveit: Sure, using `itr.next()` would also work fine :) The reason I suggested you use would be that it's clearer what it is. But I agree that if you're only using it in one case then it might not be worth it. Maybe renaming `itr` to `iteratorOverFrequencies` would have the same effect and make it more clear? What do you think? :)

exercises/poker/src/test/java/PokerTest.java
FridaTveit: No problem, that's completely understandable! We all have other committments as well so it's fine to need some time before answering comments and updating reviews. This is meant to be fun to do so don't feel pressured to do everything really fast if you're busy. It's better to do it well and have fun while doing it, than doing it super fast :)

According to [this page](http://stackoverflow.com/questions/15655126/how-to-auto-format-code-in-eclipse) the shortcut on eclipse is ctrl + shift + F :)

Zaldrick:
Hi !
Yeah sorry for putting aside this PR.
I think it will be possible for me to finish the job next week. So if it's okay for you, I'll find the time for it :)

FridaTveit:
Sure, sounds great! :)

Zaldrick:
Hey, I tried to take your old suggestions (about instance functions and some other little things) into account. I guess it's pretty decent now. What do you think ? Is there anything else that could improve the code ?

config.json
```diff
@@ -404,6 +404,15 @@
       ]
     },
     {
+      "slug": "poker",
+      "difficulty": 7,
+      "topics": [
```
FridaTveit: At the moment, none of the other exercises have topics (see issue #523). But letting people add topics to new exercises if they want to might be a good idea. Thoughts @exercism/java?

exercises/poker/src/main/java/Poker.java
```diff
@@ -0,0 +1,12 @@
+import java.util.List;
```
FridaTveit: According to the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations), since this exercise is after the first 20 exercises it shouldn't have any starter implementation. Instead there should just be a `.keep` file here :)

exercises/poker/src/test/java/PokerTest.java
```diff
@@ -0,0 +1,164 @@
+import static org.junit.Assert.assertEquals;
+
+import java.util.Arrays;
+
+import org.junit.Test;
+
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
FridaTveit: This should be `import org.junit.Ignore;` :)

exercises/poker/src/test/java/PokerTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+public class PokerTest {
+    @Test
+    public void oneHand() {
+        final String hand = "4S 5S 7H 8D JC";
+        assertEquals(Arrays.asList(hand),new Poker(Arrays.asList(hand)).getBestHands());
+    }
+
+    @Test
+    @Ignore
```
FridaTveit: Sorry, I just noticied one more thing from the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests). The `@Ignore`s should have a hint: `@Ignore("Remove to run test")`. This is to make it more clear to the users that they have to remove that line to get the test to run :)

exercises/poker/src/test/java/PokerTest.java
Zaldrick: Of course ! My bad for this careless mistake !

exercises/poker/src/test/java/PokerTest.java
Zaldrick: You don't have to be sorry, it's my fault for not paying attention to POLICIES doc. Thanks for the comment :)

exercises/poker/src/main/java/Poker.java
Zaldrick: I remove function declarations and just left class declaration as I saw in another exercice that I take for example. Is this okay ?

Zaldrick:
Changes done :)

PR #388, opened 2017-03-18
flatten-array: add @Ignore to tests [Fix #379]

### LINKED ISSUE #379: flatten-array: add @Ignore to tests
Add `@Ignore` annotation to all tests except the first one in FlattenerTest.

michaelspets:


stkent:
I always forget to add this! Is there no import statement required here?

jtigger:
Oh, this is hilarious.  So, because in both the unit test and CI builds we _delete_ all `@Ignore`s, this specific case is not tested for.

exercises/flatten-array/src/test/java/FlattenerTest.java
```diff
@@ -32,6 +32,7 @@ public void testASingleLevelOfNestingWithNoNulls() {
               8)));
     }
 
+    @Ignore
```
jtigger: As @stkent noted, going to need an `import org.junit.Ignore;` at the top to prevent a compilation error for practitioners when they download the exercise.

stkent:
Hah, well-deduced @jtigger! I was most confused when the build passed. Look great now with the import added; LGTM! Thanks, @michaelspets 💯

jtigger:
Yeah, how about that?!?! :)

PR #390, opened 2017-03-20
grade-school: remove potential test pollution [Fix #371]

### LINKED ISSUE #371: grade-school: remove potential test pollution
The `grade-school` exercise creates a single instance of `School` and uses it through the entire test suite.  The suite has deftly avoided test pollution by leaning on deterministic test ordering and using different grades in each test.  This sets a bad example of how to organize tests: each test should be isolated from the others so as to avoid false negatives (test failures when there is no corresponding defect in the code) and false positives (tests passing when a defect exists but is covered up by side-effects of earlier tests).

Please move the construction of `School` into a [`setUp()`](http://junit.sourceforge.net/javadoc/org/junit/Before.html) method.

superklamer:
Changed school object instantiation to be in a setUp method instead of the class body. And added import statement for @Before.

### DIFF
--- a/exercises/grade-school/src/test/java/SchoolTest.java
+++ b/exercises/grade-school/src/test/java/SchoolTest.java
@@ -1,3 +1,4 @@
+import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


jtigger:
There it is!  Thanks for your persistence, @superklamer.

PR #398, opened 2017-03-21
difference-of-squares: adjust API to use instance method [Fix #253]

### LINKED ISSUE #253: Grade-school: Structure very unclear

FridaTveit:
Fixes issue #253.

### DIFF
--- a/exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
+++ b/exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
@@ -13,7 +13,7 @@ public final class DifferenceOfSquaresCalculator {
                 .sum();
     }
 
-    public int betweenSquareOfSumAndSumOfSquaresTo(final int input) {
+    public int computeDifferenceOfSquares(final int input) {
         return computeSquareOfSumTo(input) - computeSumOfSquaresTo(input);
     }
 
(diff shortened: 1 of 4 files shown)


exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
```diff
@@ -1,19 +1,19 @@
...
 
-    public static int computeSumOfSquaresTo(final int input) {
+    public int computeSumOfSquaresTo(final int input) {
         return IntStream.rangeClosed(1, input)
                 .map(i -> (int) Math.pow(i, 2))
                 .sum();
     }
 
-    public static int betweenSquareOfSumAndSumOfSquaresTo(final int input) {
+    public int betweenSquareOfSumAndSumOfSquaresTo(final int input) {
```
stkent: Given the rename of the test class, maybe updating this to `computeDifferenceOfSquares` would be clearer and shorter now?

stkent:
@FridaTveit one tiny suggestion, looks good to me otherwise!

exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
FridaTveit: True, will change it! :)

stkent:
Great, thanks @FridaTveit!

PR #403, opened 2017-03-27
sum-of-multiples: use instance method instead of static method [Fix #355]

### LINKED ISSUE #355: sum-of-multiples: adjust API to use an instance method over class methods

FridaTveit:
Fixes issue #355.


stkent:
Looks good to me! :)

PR #404, opened 2017-03-28
Beer song instance methods

vasouv:
Oops, because I moved a couple of curly brackets, the diff is weird now. Should I fix in case it's confusing?

Also, question: why do we have ignore annotations if gradle runs all the tests anyway?

### DIFF
--- a/exercises/beer-song/src/test/java/BeerSongTest.java
+++ b/exercises/beer-song/src/test/java/BeerSongTest.java
@@ -9,7 +9,7 @@ public class BeerSongTest {
     public void singFirstVerse() {
         assertEquals("99 bottles of beer on the wall, 99 bottles of beer.\n" +
                         "Take one down and pass it around, 98 bottles of beer on the wall.\n\n",
-                BeerSong.verse(99));
+                new BeerSong().verse(99));
     }
 
     @Ignore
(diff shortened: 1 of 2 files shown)


FridaTveit:
That's fine, I'm glad you moved the curly brackets, it looks better now :) Doing `gradle build` runs all the tests but doing `gradle test` (which is what we tell the people doing the exercises to do) only runs the ones without `@Ignore` :)

vasouv:
You're absolutely right! Now that you've mentioned it, I recall reading in the docs that the build runs all tests for devs, and ignores some of them for the users. I get it.

Thanks for the clarification :)

exercises/beer-song/src/test/java/BeerSongTest.java
```diff
@@ -9,39 +9,39 @@
     public void singFirstVerse() {
         assertEquals("99 bottles of beer on the wall, 99 bottles of beer.\n" +
                         "Take one down and pass it around, 98 bottles of beer on the wall.\n\n",
-                BeerSong.verse(99));
+                new BeerSong().verse(99));
```
FridaTveit: If you make a private BeerSong variable, then you can initialise it in a setup method and just do `beerSong.verse(99)` here :) That way you don't have to duplicate the initialisation in every test :)

vasouv:
Fixed. And to think my previous pull request was doing exactly that at the wordy exercise :P

PR #405, opened 2017-03-28
twelve-days: use instance methods [Fix #357]

### LINKED ISSUE #357: twelve-days: adjust API to use an instance method over class methods

FridaTveit:
Fixes issue #357.

### DIFF
--- a/exercises/twelve-days/src/example/java/TwelveDays.java
+++ b/exercises/twelve-days/src/example/java/TwelveDays.java
@@ -16,11 +16,11 @@ public class TwelveDays {
             "twelve Drummers Drumming, "
     };
 
-    public static String verse(int verseNumber) {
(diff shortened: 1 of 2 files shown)


exercises/twelve-days/src/example/java/TwelveDays.java
```diff
@@ -16,11 +16,11 @@
             "twelve Drummers Drumming, "
     };
 
-    public static String verse(int verseNumber) {
+    public String verse(int verseNumber) {
         return constructVerse(verseNumber);
```
jtigger: Any reason we wouldn't just inline `constructVerse()` here?

FridaTveit: No, I can change it :)

PR #412, opened 2017-04-05
Add probot configuration file

### LINKED ISSUE #411: Override probot/stale defaults, if necessary
Per the discussion in https://github.com/exercism/discussions/issues/128 we
will be installing the probot/stale integration on the Exercism organization on
April 10th, 2017.

By default, probot will comment on issues that are older than 60 days, warning
that they are stale. If there is no movement in 7 days, the bot will close the issue.
By default, anything with the labels `security` or `pinned` will not be closed by
probot.

If you wish to override these settings, create a .github/stale.yml file as described
in https://github.com/probot/stale#usage, and make sure that it is merged
before April 10th.

If the defaults are fine for this repository, then there is nothing further to do.
You may close this issue.

stkent:
Closes #411.

Initial configuration specifies that "first-timers only" and "policy" issues should be exempt from cleanup, and that issues with no activity in the last 60 days should have the (new) [stale label](https://github.com/exercism/xjava/labels/stale) applied. I _think_ the quoted string will work here!

Defaults for non-specified configuration can be found here:

https://github.com/probot/stale#usage

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,8 @@
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - "first-timers only"
+  - "good first patch"
+  - "policy"
+
+# Label to use when marking as stale
+staleLabel: stale


.github/stale.yml
```diff
@@ -0,0 +1,7 @@
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  "first-timers only"
```
jtigger: I'm pretty sure that in YAML, elements in a list are bullets:

```yaml
exemptLabels:
  - "first-timers only"
  - "policy"

jtigger: Also, should we include "good first patch"?  Or are you considering folding those into the "first-timers only" label?

stkent:
Good catch r.e. YAML formatting. And I'd say, let's protect good first patches as well! Changes incoming...

stkent:
@jtigger updated!

jtigger:
Looks like a fine start to me.  Merging!

PR #414, opened 2017-04-05
pig-latin: convert to use instance methods

### LINKED ISSUE #360: pig-latin: adjust API to use an instance method over class methods

stkent:
Closes #360.

If I understand correctly, this is now an example of a "Function Object" rather than a "Value Object"? Yay, functions as almost-first-class citizens.

### DIFF
--- a/exercises/pig-latin/src/example/java/PigLatinTranslator.java
+++ b/exercises/pig-latin/src/example/java/PigLatinTranslator.java
@@ -2,7 +2,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class PigLatin {
+public class PigLatinTranslator {
 
     public static final String AY = "ay";
     public static final String THR = "thr";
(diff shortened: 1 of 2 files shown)


exercises/pig-latin/src/example/java/PigLatinTranslator.java
```diff
@@ -44,16 +44,16 @@ private static String translateWord(String word) {
...
+    private boolean wordStartsWithVowelLike(String word) {
         return word.startsWith(YT) || word.startsWith(XR) || word.substring(0, 1).matches(VOWELS_REGEX);
     }
 
-    private static boolean wordStartsWithPrefixes(String word, String... prefixes) {
-        return Arrays.asList(prefixes).stream()
-                .anyMatch(x -> word.startsWith(x));
+    private boolean wordStartsWithPrefixes(String word, String... prefixes) {
+        return Arrays.stream(prefixes)
+                .anyMatch(word::startsWith);
```
stkent: Both these changes suggested by IntelliJ.

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -1,5 +1,4 @@
 import org.junit.Test;
-import org.junit.Ignore;
```
stkent: Unused because of the parametrization.

PR #418, opened 2017-04-10
Add script to prune extraneous .keep files

stkent:
**Goal**

Automate removal of extraneous `.keep` files (used to retain practitioner source directories in Git when no explicit `.java` starter files are provided).

**Usage**

From the root of the xjava repository, execute `./scripts/prune-extra-keep-files.sh`

### DIFF
--- a/scripts/prune-extra-keep-files.sh
+++ b/scripts/prune-extra-keep-files.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+for EXERCISE_DIRECTORY in $(find exercises -mindepth 2 -maxdepth 2 -type d -name "src"); do
+  STARTER_DIRECTORY="${EXERCISE_DIRECTORY}/main/java"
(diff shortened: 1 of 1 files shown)


scripts/prune-extra-keep-files.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+for EXERCISE_DIRECTORY in $(find exercises -mindepth 2 -maxdepth 2 -type d -name "src"); do
```
jtigger: 👍 using "modern" command substitution.  ([why is it preferred?](http://mywiki.wooledge.org/BashFAQ/082)).

scripts/prune-extra-keep-files.sh
```diff
@@ -0,0 +1,12 @@
...
+  STARTER_DIRECTORY="${EXERCISE_DIRECTORY}/main/java"
+  STARTER_FILE_COUNT=$(find "${STARTER_DIRECTORY}" -mindepth 1 -maxdepth 1 -type f -name "*.java" | wc -l)
```
jtigger: 👍 appreciate the precision you're using, here; it's easy for declarative statements to "say more" than we intend.

PR #421, opened 2017-04-11
atbash-cipher: make class use instance method [Fix #366]

### LINKED ISSUE #366: atbash-cipher: adjust API to use an instance method over class methods

FridaTveit:
Fixes issue #366.


stkent:
✅  😄

PR #422, opened 2017-04-12
bookstore: fix test suite equality checks

### LINKED ISSUE #416: book-store: Example implementation is incorrect
The description of the exercise describes a a scenario that is also a [test case](https://github.com/exercism/xjava/blob/master/exercises/book-store/src/test/java/BookstoreTest.java#L70-L74) where two groups of 4 are cheaper than two groups of 5 and 3. The example implementation always starts with the biggest possible group and does not try all other combinations to really find the cheapest price.

The problem why the tests don't find this is the incorrect use of the delta parameter in the assert method.
```java
List<Integer> books = new ArrayList<>(Arrays.asList(1,1,2,2,3,3,4,5));
Bookstore bookstore = new Bookstore(books);
assertEquals(51.20,bookstore.calculateTotalCost(),2);
```

While `calculateTotalCost()` actually returns the wrong price of `51.6` because of the `delta=2` even `53.2` or `49.2` would be considered equal.

> delta - the maximum delta between expected and actual for which both numbers are still considered equal. 

Maybe `delta` should better be something like `0.001` so people really implement the correct algorithm.

stkent:
Closes #416.

Example implementation was updated to match the [xcsharp](https://github.com/exercism/xcsharp/blob/master/exercises/book-store/Example.cs) implementation.

I also ran the classes through my formatter to make them more consistent with other files in the repo.

### DIFF
--- a/exercises/book-store/src/example/java/Bookstore.java
+++ b/exercises/book-store/src/example/java/Bookstore.java
@@ -1,61 +1,57 @@
-import java.awt.print.Printable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class Bookstore{
+class Bookstore {
 
-	private static int BOOK_PRICE = 8, MAX_GROUP_SIZE = 5;
-	private List<Integer> books;
-	private static double[] DISCOUNT_TIERS = {0,5,10,20,25};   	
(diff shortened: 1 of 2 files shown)


stkent:
Comparison view ignoring whitespace-only changes: https://github.com/exercism/xjava/pull/422/files?w=1

PR #423, opened 2017-04-13
palindrome-products: make class use instance methods [Fix #377]

### LINKED ISSUE #377: palindrome-products: adjust API to use an instance method over class methods

FridaTveit:
Fixes issue #377.


stkent:
Hooray! Almost there, once I get around to the final task!

PR #424, opened 2017-04-14
pascals-triangle: update to v1.0.0 tests

### LINKED ISSUE #365: pascals-triangle: adjust API to use an instance method over class methods

stkent:
Closes #365.

Updates the exercise to match the [most current canonical data](https://github.com/exercism/x-common/blob/e9393ff749095d30e6386b4ebec8238ee53454c0/exercises/pascals-triangle/canonical-data.json) at the time of implementation.


exercises/pascals-triangle/src/example/java/TriangleGenerator.java
```diff
@@ -0,0 +1,21 @@
+class TriangleGenerator {
```
FridaTveit: Maybe `PascalsTriangleGenerator` would make it more obvious what the class is for?

stkent:
@FridaTveit updated!

Issue #426, opened 2017-04-17
rna-transcription: make a lot of noise about deleting @Ignores


jtigger:
I have not encountered at least two solutions that pass the first test case swimmingly, but clearly wouldn't pass the _whole_ suite:

- http://exercism.io/submissions/1794866917ef4e79beeb6cd40c469c81
- http://exercism.io/submissions/757e83f9001e4e6699f45df89346477a

This should be _very_ rare.

I'm taking this as a sign we're not making enough noise about deleting `@Ignore`s.

Please update `rna-transcription` so that it is hard to miss the fact that once the first test case succeeds the practitioner is reminded to delete the next `@Ignore`.

Options (include as many as you think would be sufficient):
- include a `HINT.md`
- put comments in the test suite.
- add some console output that prints after the first test succeeds.
- other?

stkent:
We could adopt the xcsharp-style [hints in `@Ignore` annotations](http://junit.sourceforge.net/javadoc/org/junit/Ignore.html):

> @Ignore takes an optional default parameter if you want to record why a test is being ignored

I'm not sure that's enough by itself, but it can't hurt!

ajayboseac:
@jtigger what is the use of the` @Ignore  ` tags  ? Why cant we just delete it ?

jtigger:
Great question @ajayboseac.  The historical discussion / policy is #101.  tl;dr — to make it crystal clear what failing test should be addressed at any given moment.

stkent:
Thanks to @FridaTveit, we've now added hints to the ignore statements in rna-transcription to prompt users to un-ignore noninitial tests. What else should we, or do we need to, do? My current thoughts are as follows:

- add the same hints to _every_ test suite ignore statement, which I believe is how e.g. xcsharp operates;
- add explicit guidance via the HINTS.md file to the _first_ exercise that contains ignored tests. This might be best placed in the new exercise that's going to replace the current (overcomplicated) hello-world.
- capture the decisions we make in a policy, somewhere. (ref. https://github.com/exercism/xjava/issues/425)

FridaTveit:
I think that sounds like a good plan :)

jtigger:
Perfect!

FridaTveit:
For adding the Ignore hints to every test suite, should we make "good first patch" issues for that or should we just do it?

jtigger:
[...]

PR #429, opened 2017-04-19
rna-transcription: add hint to Ignore annotation

FridaTveit:
See issue #426 for discussion.


stkent:
I'm guessing this is failing because we very literally strip `@Ignore` while running the journey test; we should probably update the regex to handle `@Ignore\(.*\)` or similar.

FridaTveit:
Wops, I didn't notice it failed. I'll look into it, thanks! :)

stkent:
I think this is an excellent improvement at the right site, so I'm going to merge this PR now! However, I'd like to leave #426 open as I feel we should probably also use additional mechanisms to help implementors avoid this pitfall. Let's continue any discussion there :)

PR #430, opened 2017-04-20
Fixes #420

ajayboseac:
nth-prime: simplify example implementation

PR #431, opened 2017-04-20
Base patched Trackler gem off actual version tag.

jtigger:
- was: hand-rolling a version of Trackler based of master and then
  applying changes from the PR/commit/local change to that trackler gem.
- now: using the exact tagged version of Trackler as the basis for the
  hand-rolled gem.

This fixes the issue where a change make to Trackler's master is
incompatible with the current version of x-api.  In short, we were
building the Trackler gem wrong.

### DIFF
--- a/bin/journey-test.sh
+++ b/bin/journey-test.sh
@@ -143,6 +143,11 @@ make_local_trackler() {
 
   local xjava=$( pwd )
   pushd ${trackler}
+
+  # Get the version of Trackler x-api is currently using
+  local version=$( grep -m 1 'trackler' ${xapi_home}/Gemfile.lock | sed 's/.*(//' | sed 's/)//' )
+
+  git checkout v${version}
   git submodule init -- common
   git submodule update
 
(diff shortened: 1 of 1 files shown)


jtigger:
This come up by inadvertently breaking PR #430.

Our journey test puts a complete install of Exercism through its paces, ensuring that each and every exercise in the Java track can be fetched and solved.

In preparation for that journey, we stand-up two key components of Exercism: x-api (the endpoint which clients obtain exercises) and trackler (the versioned library of those exercises).

In order to test the patch of Java exercise code, we need to hand-roll a version of the trackler gem to a) be compatible with the latest version of x-api and b) include the changes we're currently testing.

To date, we were grabbing the `master` branch of trackler and applying the Java track changes.  At the time, I didn't appreciate that this ran the risk that if there were changes to Trackler code itself they may be incompatible with the latest x-api.  Today, they did.

The right way to do this is to first check-out the local check-out of Trackler to the tagged version.  That's what this patch does.

jtigger:
@exercism/java could I get a second pair of eyes on this?

Two reasons:
1. I'd rather we be right than fast.
2. I'd rather _we_ have some insight into how our CI works, generally.  Here's a good/meaningful peek into the journey test.

Thank you!

jtigger:
Thank you, @FridaTveit!

Issue #455, opened 2017-04-28
rna-transcription: rename "ofDna" with "transcribe"


jtigger:
Had [a conversation with @chuckwondo](http://exercism.io/submissions/bdb1052cde20421f9396fdebdf412f31) on the Java track about the name of the method in `rna-transcription`; through it, it became clear that `ofDna()` lacks a certain clarity that a verb like `transcribe()` would include more punch.

Any objections to renaming that method?

FridaTveit:
Sounds like a good idea :)

chuckwondo:
What about also making the method static and updating the test suite to match? It seems odd to me to be forced to create a stateless instance just so that the method can be invoked.

FridaTveit:
We've made it a policy to prefer instance methods over static methods. See issue #177 for discussion of why :)

jtigger:
To be clear, @chuckwondo, the things that we're finding that "need" a policy are exactly those stylistic grey areas.  As I think clear in the policy issue itself, there's good reasons for either approach.

There's definitely "inertia" behind the instance-based approach, but there's always room for discussion.  Always.

chuckwondo:
Thanks @jtigger. I'm glad to see that there are documented policies, and I'm glad there's room for discussion.

For this particular policy, I'm not sure I agree with always using instance methods instead of static methods is the most effective stance. There are good reasons for using one or the other in various scenarios.

At the very least, if you want to stick with instance methods, then I suggest designing the usage to better align with that choice. In particular, using an instance method is generally used in conjunction with a stateful object. Therefore, if you're going to invoke an instance method, it would appear to be better aligned with passing state to a constructor, not to the method.

Passing everything required for a computation to the method performing that computation implies there's no need to have an instance, because there's no need for it to hold state information.

Therefore, I propose that for instance methods, we should pass state information to the constructor and pass nothing to the method (unless there's a need to do so beyond the state information held by the instance). For static methods, obviously pass all necessary inputs directly to the method.

What I am not fond of in this case with rna-transcription (and all of the other problems that follow suit) is the creation of an instance that holds no state, only to invoke a method on that instance to which all necessary input is passed. Either make the method static, or pass the state to the constructor (and pass nothing to the method).

For the less experienced programmers, mixing things up like this has the potential to instill arguably poor coding practices.

FridaTveit:
[...]

PR #521, opened 2017-05-09
rna-transcription: rename "ofDna" with "transcribe"

Smarticles101:
Ref. #455 
Also renamed input string to `dnaStrand`


Smarticles101:
@jtigger I think that the issue you mentioned is well related to this PR.  I think it would be great to rename it to `dnaStrand`.  I'll go ahead and make those changes. :+1:

PR #522, opened 2017-05-09
run-length-encoding: add to track

Smarticles101:
Lot's of free time with the last couple weeks of school on my hands and summer coming up, plan to implement a bunch of exercises. :grin:


config.json
```diff
@@ -136,6 +136,11 @@
       "topics": []
     },
     {
+      "slug": "run-time-encoding",
```
Smarticles101: Lol oops

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
```diff
@@ -0,0 +1,47 @@
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class RunLengthEncoding {
+    public String encode(String data) {
+        String encodedData = "";
+
+        for (int i = 0; i < data.length(); i++) {
```
Smarticles101: I think that's a great idea!  I was actually thinking of writing it this way after I had finished it.  Implementing it this way would increase readability a lot.

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
```diff
@@ -0,0 +1,47 @@
...
+            char c = data.charAt(i);
+
+            if (i != 0 && c == data.charAt(i - 1)) continue;
```
Smarticles101: Good idea.  I try to keep the body of code on the same line as the header if it's only one line and doesn't require curly braces.  This way, if I add more code later, it's less likely for me to have problems because I will see that there are no curly braces.  I suppose though, that it's better just to insert the curly braces anyways.  I think that it's also more common to see it that way.

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
```diff
@@ -0,0 +1,47 @@
...
+
+            int amt = 1;
+
+            for (int j = i + 1; j < data.length() && c == data.charAt(j); j++) {
+                amt++;
+            }
+
+            if (amt == 1) encodedData += c;
+            else encodedData += amt + Character.toString(c);
```
Smarticles101: Cool!  I've never heard about StringBuilder before, it seems like it would be a lot better though considering that concatenating a string creates a `new` string every time

exercises/run-length-encoding/src/test/java/RunLengthEncodingTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public void encodeWithOnlySingleValues() {
+        Assert.assertEquals("XYZ", runLengthEncoding.encode("XYZ"));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void decodeWithOnlySingleValues() {
+        Assert.assertEquals("XYZ", runLengthEncoding.decode("XYZ"));
+    }
+
```
Smarticles101: Guess I missed that.  Oops

exercises/run-length-encoding/src/test/java/RunLengthEncodingTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Before
+    public void setUp() {
+        runLengthEncoding = new RunLengthEncoding();
+    }
+
+    @Test
+    public void encodeEmpty() {
+        Assert.assertEquals("", runLengthEncoding.encode(""));
+    }
+
```
Smarticles101: Are you sure?  Other tests I have looked at leave the `@Ignore` off of the first/simplest test.

FridaTveit: It should only be the first test which doesn't have an `@Ignore`. Youv'e left it off the first two (my comment was meant for the second test, sorry if that wasn't clear) :)

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
FridaTveit: Yes, I agree, it's not so bad when it's on the same line as the header. But it's probably still a bit clearer to put it on it's own line with curly brackets. That way it's easier to see that it's an if statement :)

Smarticles101:
Take a look at my new encode method.  It's all in one loop, but it feels slightly 'messier' to me.  What's your opinion @FridaTveit 

Also, I think I somehow screwed up the files somehow because its saying I changed every single file when I run `git status`.  I ran `git diff` and saw that `^M` was at the end of every line, which is apparently some sort of newline character.  Anyone know how to get rid of all these?

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
```diff
@@ -0,0 +1,65 @@
...
+        if (data.length() == 0) {
+            return data;
+        }
+
+        int charCount = 0;
+        char beforeChar = data.charAt(0);
+
+        for (int i = 0; i < data.length(); i++) {
+            char currentChar = data.charAt(i);
+
```
FridaTveit: I think there are too many ifs here. From what I understand, you only really need to do two things; if `beforeChar == currentChar` then you need to increment the count. If `beforeChar != currentChar` or this is the last char in the string, then we need to add the latest char with count to the string. So something like (in pseudocode):
```
if (beforeChar == currentChar) {
  increment count
}

if (beforeChar != currentChar || reached end of string) {
  add beforeChar with count
  if beforeChar != currentChar && reached end of string then add currentChar with count 1
  beforeChar = currentChar
  charCount = 1
}
```

Extracting the adding char with count to its own method would also simplify it :)

What do you think? :)

FridaTveit:
Oh, and of course if you disagree with any of my comments then feel free to say so or to leave them, they're just suggestions :)

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
Smarticles101: I think this is great.  This was definitely the help I was looking for :+1:

Smarticles101:
Yep @FridaTveit I think I've almost fixed that.  I might add another pull request some time to add a .gitattributes file to ensure that the right eol characters are used so that no one else has this problem.

exercises/run-length-encoding/src/example/java/RunLengthEncoding.java
```diff
@@ -4,41 +4,43 @@
...
 
-            beforeChar = currentChar;
-            charCount = 1;
+    public void addChars(StringBuilder toAddTo, char toAdd, int countOfChar) {
+        if(countOfChar != 1) {
+            toAddTo.append(countOfChar);
         }
 
-        return encodedData.toString();
+        countOfChar.append(toAdd);
```
FridaTveit: This is why the build is failing. It should be `toAddTo.append(toAdd)`. Sorry I didn't spot it before!

PR #560, opened 2017-05-14
gigasecond: standardize starter implementation

### LINKED ISSUE #538: gigasecond: create and/or validate starter implementations

This exercise is currently one of the first ten exercises in the track. Please review the exercise files and create and/or update the starter implementation according to the guidelines provided in the link above.

stkent:
Fixes #538.

---

Reviewer Resources:

### DIFF
--- a/exercises/gigasecond/src/example/java/Gigasecond.java
+++ b/exercises/gigasecond/src/example/java/Gigasecond.java
@@ -3,18 +3,18 @@ import java.time.LocalDate;
 
 class Gigasecond {
 
-    private LocalDateTime initialDateTime;
+    private LocalDateTime birthDateTime;
 
-    Gigasecond(LocalDate initialDate) {
-        this.initialDateTime = initialDate.atTime(0, 0);
+    Gigasecond(LocalDate birthDate) {
+        this.birthDateTime = birthDate.atTime(0, 0);
     }
 
-    Gigasecond(LocalDateTime initialDateTime) {
-        this.initialDateTime = initialDateTime;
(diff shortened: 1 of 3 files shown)


exercises/gigasecond/src/example/java/Gigasecond.java
```diff
@@ -1,19 +1,20 @@
 import java.time.LocalDateTime;
 import java.time.LocalDate;
 
-public class Gigasecond {
+class Gigasecond {
 
-    private LocalDateTime birthDateTime;
+    private LocalDateTime initialDateTime;
```
FridaTveit: I agree that calling it `initialDateTime` makes the class more general and might be nicer. But calling it `birthDateTime` lines up more with the description of the exercise and might make it easier for people solving it to understand what they need to do? I don't particularly mind though :)

stkent: I feel like we've had a very similar discussion before, so apologies, but: I was referring to the [canonical description](https://github.com/exercism/x-common/blob/master/exercises/gigasecond/description.md) for this problem and it doesn't mention birthdates (which is why I made the adjustment I did). Please remind me where I can find the reference you're thinking of? Is the [blurb](https://github.com/exercism/x-common/blob/master/exercises/gigasecond/metadata.yml#L2) always combined with the `description.md` file? Thanks!

FridaTveit: Yes, that was what I was thinking of :)

stkent:
@FridaTveit updated!

PR #574, opened 2017-05-16
hello-world: explain UnsupportedOperationException in HINT.md file [Fix #350]

### LINKED ISSUE #350: rna-transcription: what to do with the UnsupportedOperationException
In the second exercise, `rna-transcription`, we introduce the practitioner to the idiomatic `UnsupportedOperationException`.  This is lovely in many ways.

Let's soften the introduction with a terse mention of what that exception is, what it does (i.e. causes the test to fail with that error message) and what to do about it.

Do this by creating a `HINT.md` in the root of the exercise directory; the contents of that file are automatically merged into the `README.md` file delivered with the exercise.

FridaTveit:
Fixes issue #350.

---

Reviewer Resources:


exercises/hello-world/HINT.md
```diff
@@ -0,0 +1,4 @@
+The starter implementation for this exercise uses [UnsupportedOperationException](http://docs.oracle.com/javase/8/docs/api/?java/lang/UnsupportedOperationException.html).
```
stkent: Is 'starter implementation' terminology that practitioners are familiar with? Maybe 'default' implementation would be easier to understand here if not?

exercises/hello-world/HINT.md
```diff
@@ -0,0 +1,4 @@
+The starter implementation for this exercise uses [UnsupportedOperationException](http://docs.oracle.com/javase/8/docs/api/?java/lang/UnsupportedOperationException.html). 
+This [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/) will cause the test to fail with the provided error message.
```
stkent: I'd go a little further and explain why we do this: "Throwing exceptions in this manner means the tests will compile but not pass right away." or similar?

Separate point: it seems a little back-to-front to be linking out the the general Exceptions docs after linking to the Javadoc for a specific implementation; is there a way to reorder this hint to avoid that?

exercises/hello-world/HINT.md
stkent: Maybe we can add a little tidbit here on `UnsupportedOperationException` specifically, since any exception would cause a test failure with a given method. Perhaps something like "This exception is commonly thrown to alert developers that a method has not yet been implemented."

FridaTveit: Both great suggestions, will change it! :)

exercises/hello-world/HINT.md
FridaTveit: Sure, I'll add that and try to see if I can reorder it :)

stkent:
@FridaTveit great tweaks, thanks! LGTM! :shipit:

PR #585, opened 2017-05-21
secret-handshake: update starter implementation

### LINKED ISSUE #544: secret-handshake: create and/or validate starter implementations

stkent:
Fixes #544.

---

Reviewer Resources:

### DIFF
--- a/exercises/secret-handshake/src/main/java/HandshakeCalculator.java
+++ b/exercises/secret-handshake/src/main/java/HandshakeCalculator.java
@@ -1,5 +1,9 @@
+import java.util.List;
+
 final class HandshakeCalculator {
 
-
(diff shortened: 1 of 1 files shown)


exercises/secret-handshake/src/main/java/HandshakeCalculator.java
```diff
@@ -1,5 +1,9 @@
-final class HandshakeCalculator {
+import java.util.List;
 
+final class HandshakeCalculator {
 
+    List<Signal> calculateHandshake(int number) {
```
chuckwondo: I suggest adding the `public` modifier. It's not necessary for the tests to pass, but it's not generally left out, unless you specifically want something to be package private.

PR #586, opened 2017-05-21
sum-of-multiples: update starter implementation

### LINKED ISSUE #545: sum-of-multiples: create and/or validate starter implementations

stkent:
Fixes #545.

---

Reviewer Resources:

### DIFF
--- a/exercises/sum-of-multiples/src/main/java/SumOfMultiples.java
+++ b/exercises/sum-of-multiples/src/main/java/SumOfMultiples.java
@@ -1,3 +1,11 @@
-public class SumOfMultiples {
-    
+class SumOfMultiples {
+
+    SumOfMultiples(int number, int[] set) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    int getSum() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
 }


exercises/sum-of-multiples/src/main/java/SumOfMultiples.java
```diff
@@ -1,3 +1,11 @@
...
+class SumOfMultiples {
+
+    SumOfMultiples(int number, int[] set) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    int getSum() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
```
chuckwondo: I suggest making the constructor and the method `public`. Although leaving them out won't cause problems for running the tests (since they are in the same default package), that's not generally how they would be written, and I be

stkent: Tricky one. See https://github.com/exercism/xjava/issues/571 for some prior discussion. I tend to think of each exercise's tests + implementations as the entire system, and write code with that in mind. I'm not sure @exercism/java has reached a consensus on this yet, however.

chuckwondo: Fair enough. Thanks for the follow up.

Issue #571, opened 2017-05-16
Remove final keyword from all test class declarations


stkent:
Discussion and consensus: https://github.com/exercism/xjava/issues/373

Marking our test classes as `final` is (1) more syntax for a learner to parse, and (2) not increasing the clarity of the writer's intent given the context of an individual problem's codebase. We should remove it from all test class declarations.

morrme:
Totally abusing this issue to ask: I understand that there is a goal to have a significant number of `good first patch` issues. However, I'm always looking to pitch in and there are very few issues *not* reserved for first timers.  Are these tracks generally supposed to be seen as a "first stop" towards OSS contribution, and not really made for "repeat" contributors ?

stkent:
No, you are absolutely welcome and actively encourage to tackle any other issues you feel comfortable with! The good first patch backlog is maintained to allow newcomers to ease into open source contribution, but once you've got some of those under your belt, moving on to more complex tasks is a natural step. There are always [new exercises to build](http://exercism.io/languages/java/todo) (just make sure you check that no-one has called dibs yet, and that the exercise is not marked deprecated or foregone in the config.yaml file at the root of the repo) as well as the open issues. If you're unsure about jumping in to an issue, just comment requesting help, and a maintainer will be more than happy to help you figure out anything you need to get started. ❤️

stkent:
@morrme also, just as a heads up, here are the guidelines we try to stick to with respect to good first patches: https://github.com/exercism/xjava/blob/master/POLICIES.md#good-first-patches. As you'll see, they aren't always off-limits to experienced contributors - right now we're running a surplus, I believe!

stkent:
Whoops, accidental closure, fixed!

FridaTveit:
What about new starter implementations? I think some of those classes are also marked as final. Should we remove the final keyword from those as well?

FridaTveit:
The test classes which are marked as `final`:
- [ ] WordProblemSolverTest
- [ ] BracketCheckerTest
- [ ] BaseConverterTest
- [ ] FlattenerTest
- [ ] OpticalCharacterReaderTest
- [ ] HandshakeCalculatorTest
- [ ] RobotTest
- [ ] RectangleCounterTest
- [ ] QueenAttackCalculatorTest
- [ ] NaturalNumberTest
- [ ] MinesweeperBoardTest
- [ ] LargestSeriesProductCalculatorTest
- [ ] DifferenceOfSquaresTest
[...]

PR #595, opened 2017-05-22
difference-of-squares: remove "final" from test class

morrme:
see #571 

---

Reviewer Resources:


exercises/difference-of-squares/src/test/java/DifferenceOfSquaresCalculatorTest.java
```diff
@@ -4,10 +4,7 @@
 
 import static org.junit.Assert.assertEquals;
 
-/*
- * version: 1.1.0
- */
-public final class DifferenceOfSquaresCalculatorTest {
+public class DifferenceOfSquaresCalculatorTest {
```
stkent: Looks like we lost the version comment here; could you please restore it? Thanks!

morrme: Yes, there was a conflict that needed to be resolved and it wasn't in the original file so I think I got rid of it.

stkent:
👍 awesome, thanks!

PR #597, opened 2017-05-22
simple-linked-list: simplify example implementation [Fix #568]

### LINKED ISSUE #568: simple-linked-list: unnecessary generic method
I believe that the example solution unnecessarily uses a generic method when entire class is generic. Specifically it seems to me that `private <T> T[] newArray(Class<T> clazz, int size)` can be replaced by `private T[] newArray(Class<T> clazz, int size)`.

FridaTveit:
Should fix issue #568.

---

Reviewer Resources:

### DIFF
--- a/exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
+++ b/exercises/simple-linked-list/src/example/java/SimpleLinkedList.java
@@ -3,12 +3,12 @@ import java.lang.reflect.Array;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 
-public class SimpleLinkedList<T> {
+class SimpleLinkedList<T> {
 
     private static class Element<T> {
 
         final T value;
-        Element next;
+        Element<T> next;
 
         Element(T value) {
             this.value = value;
@@ -18,16 +18,16 @@ public class SimpleLinkedList<T> {
     private Element<T> head;
     private int size;
 
-    public SimpleLinkedList() {
+    SimpleLinkedList() {
     }
 
-    public SimpleLinkedList(T[] values) {
-        for (int ii = values.length - 1; ii >= 0; ii--) {
-            push(values[ii]);
+    SimpleLinkedList(T[] values) {
+        for (int i = values.length - 1; i >= 0; i--) {
+            push(values[i]);
         }
     }
 
-    public final void push(T value) {
+    final void push(T value) {
         Element<T> newElement = new Element<>(value);
         this.size++;
         if (Objects.isNull(head)) {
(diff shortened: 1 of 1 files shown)


stkent:
I've been musing on this exercise... bear with me...

Since we have both `simple-linked-list` and `linked-list`, and the capabilities demanded by `linked-list` are a superset of the capabilities demanded by `simple-linked-list` (AFAICT), it seems like the tests for `simple-linked-list` should be a strict subset of the tests for `linked-list`. Right now, that's not the case, and there are tests that demand _extra_ functionality in `simple-linked-list` that is not part of `linked-list`.

I think this PR improves the situation captured by #568 for sure; however, I think it could go a little further based on the reasoning above. @FridaTveit what do you think? And if you agree, would you like to make a separate card for that work, or address it right now by modifying this PR?

FridaTveit:
I completely agree! And I'd be happy to make those changes as part of this PR :)

stkent:
Awesome! ❤️

FridaTveit:
Actually, having looked more closely at the [description for simple-linked-list](https://github.com/exercism/x-common/blob/master/exercises/simple-linked-list/description.md), it doesn't look like there's very much, if anything, we can remove. The description says that you need to be able to reverse the list and convert to and from an array, and those are the main features that `simple-linked-list` has but `linked-list` doesn't, as far as I can tell. But I still think it's strange that, in my opinion, `simple-linked-list` is more difficult to implement than `linked-list`... @stkent, what do you think?

stkent:
In that case, I'd say the issue may lie with the [description for `linked-list`](https://github.com/exercism/x-common/blob/master/exercises/linked-list/description.md), which (IMO) should include all those requirements from `simple-linked-list` plus whatever tests are specific for the doubly-linked version. It might be worth tackling the open x-common issues and standardizing the canonical data for these exercises before we launch down a path that might be separate from other tracks.

https://github.com/exercism/x-common/issues/587
https://github.com/exercism/x-common/issues/568

FridaTveit:
Standardising the canonical data sounds like a good plan :) I think the description for `linked-list` is okay, that's what I would expect from a linked list. I more think the problem is the description for `simple-linked-list`. I think either the "to and from array" bit (and possibly the reverse bit as well) should be removed, or it should be renamed `single-linked-list` instead of `simple`. Because at the moment I don't think it really is simple...

stkent:
Agreed, I think ideally we'd have

- `simple-linked-list` renamed to `singly-linked-list`;
- `linked-list` renamed to `doubly-linked-list`;
- `singly-linked-list` made simpler than `doubly-linked-list`;

but I'm not sure how difficult it is to rename an exercise slug. (I think it's been done once before, so it is possible...) I'm going to merge this PR now, since the discussions above are vastly expanding the scope, and then we can work on those upstream issues before making any further changes. Sound good?

FridaTveit:
Sounds good :)

stkent:
See: https://github.com/exercism/discussions/issues/156 for discussion around exercise renames.

PR #601, opened 2017-05-23
Use idiomatic string interpolation in build file

stkent:
http://docs.groovy-lang.org/latest/html/documentation/#_string_interpolation

---

Reviewer Resources:

### DIFF
--- a/exercises/build.gradle
+++ b/exercises/build.gradle
@@ -9,14 +9,14 @@ subprojects { project ->
       java.srcDirs = ["src/example/java"]
     }
     project["compileJava"].doFirst { compileTask ->
-      println "  (source = " + compileTask.source.asPath + ")"
+      println "  (source = ${compileTask.source.asPath})"
     }
 
     starterSource {
       java.srcDirs = ["src/main/java"]
     }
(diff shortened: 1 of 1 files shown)


exercises/build.gradle
```diff
@@ -9,14 +9,14 @@ subprojects { project ->
...
-      println "  (source = " + compileTask.source.asPath + ")"
+      println "  (source = ${compileTask.source.asPath})"
     }
 
     starterSource {
       java.srcDirs = ["src/main/java"]
     }
     project["compileStarterSourceJava"].doFirst { compileTask ->
-      println "  (source = " + compileTask.source.asPath + ")"
+      println "  (source = ${compileTask.source.asPath})"
```
FridaTveit: From the documentation it looks like you can also use just `$compileTask.source.asPath`? But I'm happy to keep the curly brackets if leaving them out doesn't work or you think it's clearer to keep them :)

stkent: For something short like `$compileTask` I normally omit the braces, but with a longer composite invocation such as this one, I prefer them for clarity of scoping, as you suggested :)

PR #604, opened 2017-05-24
transpose: add to track

Smarticles101:
Implementing [transpose](https://github.com/exercism/x-common/blob/master/exercises/transpose/description.md)
([test cases](https://github.com/exercism/x-common/blob/master/exercises/transpose/canonical-data.json))

---

Reviewer Resources:


Smarticles101:
Finally got around to doing this after I took a long break. I had some fun making the example solution for this. :smile:

exercises/transpose/src/example/java/Transpose.java
```diff
@@ -0,0 +1,41 @@
...
+                transposedForm.append("\n");
+            }
+        }
+
+        return transposedForm.toString();
+    }
+
+    private void padRows(String[] rows) {
+        for (int i = 0; i < rows.length; i++) {
+            for (int j = i; j < rows.length; j++) {
```
FridaTveit: Could you pass in `maxRowSize` to this method and just do `if (rows[i].length < maxRowSize` instead of the second loop?

exercises/transpose/src/test/java/TransposeTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+    @Before
+    public void setup() {
+        transpose = new Transpose();
+    }
+
+    @Test
+    public void emptyString() {
+        String input = "";
+        String expected = "";
```
FridaTveit: You could maybe name them `inputMatrix` and `transposedMatrix`? But that's up to you, `input` and `expected is fine as well :)

exercises/transpose/src/test/java/TransposeTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+        assertEquals(expected, transpose.transpose(input));
+    }
+
+    @Test
+    @Ignore("Remove to run test")
+    public void twoCharsInARow() {
+        String input = "A1";
+        String expected = "A" +
+                "\n1";
```
FridaTveit: The indentation here makes this a bit difficult to read. Could you maybe try and align the rows of the matrices so it looks more like a matrix?

exercises/transpose/src/test/java/TransposeTest.java
```diff
@@ -0,0 +1,259 @@
...
+        String input = "A\n" +
+                "1";
+        String expected = "A1";
+
+        assertEquals(expected, transpose.transpose(input));
+    }
+
+    @Test
+    @Ignore("Remove to run test")
+    public void simple() {
```
FridaTveit: I know you've got these names from the canonical data, so it's not your fault, but `simple` is not a very descriptive test name. Maybe `transposeSimpleMatrix` or `simpleMatrix`? While we should use the test data in the canonical data, we don't have to stick to the exact test names given in the description :)

exercises/transpose/src/example/java/Transpose.java
```diff
@@ -0,0 +1,41 @@
...
+            if (rows[i].length() > maxRowSize) {
+                maxRowSize = rows[i].length();
+            }
+        }
+
+        padRows(rows);
+
+        for (int newRowOldCol = 0; newRowOldCol < maxRowSize; newRowOldCol++) {
+            for (int newColOldRow = 0; newColOldRow < rows.length; newColOldRow++) {
+                if (newRowOldCol < rows[newColOldRow].length()) {
```
FridaTveit: If `newRowOldCol` is less than `maxRowSize` (by the conditions of the `for` loop), would this condition ever not be true?

exercises/transpose/src/example/java/Transpose.java
Smarticles101: No, and this is a bit tricky to see. I struggled with seeing this while making the solution. The problem description says the following:

>Pad to the left with spaces.
>Don't pad to the right.

You should only add padding in the case where there is a row to the right that is larger than it. Using the `maxRowSize` as you suggested works fine on all test cases except the last one, where the largest row is in the middle, but the last row is smaller, this causes the problem of extra spaces being inserted at which point you're technically padding to the right.

exercises/transpose/src/example/java/Transpose.java
Smarticles101: Yes. The way you're supposed to pad, you end up having rows on one or two test cases that are not the length of the longest row

exercises/transpose/src/example/java/Transpose.java
Smarticles101: I think whoever did this in the csharp track had the same idea, so when padding they just use the `maxRowSize`. However, their last test case is messed up, it has two extra spaces at the end where there shouldn't be. Ref [this](https://github.com/exercism/csharp/issues/305)

exercises/transpose/src/test/java/TransposeTest.java
Smarticles101: Agreed. IntelliJ keeps moving the indents back. :man_facepalming:

exercises/transpose/src/example/java/Transpose.java
FridaTveit: I see. Thanks for explaining that! :)

exercises/transpose/src/example/java/Transpose.java
```diff
@@ -0,0 +1,41 @@
...
+                    transposedForm.append(rows[newColOldRow].charAt(newRowOldCol));
+                }
+            }
+
+            if (newRowOldCol != maxRowSize - 1) {
```
FridaTveit: From what I can see then `maxRowSize` is only being used to check whether you've reached the last new row? To simplify that you could store each new row in an array and then at the end do `String.join("\n". newRows)`. You definitely don't have to do that though, it's just a suggestion :)

exercises/transpose/src/main/java/Transpose.java
```diff
@@ -0,0 +1,5 @@
+public class Transpose {
```
FridaTveit: According to the [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md) doc, since this exercise is now after the first 20 exercises, it shouldn't have any starter implementation. Instead there should just be a `.keep` file in this directory :)

exercises/transpose/src/test/java/TransposeTest.java
```diff
@@ -0,0 +1,264 @@
...
+        String expected = "A1";
+
+        assertEquals(expected, transpose.transpose(input));
+    }
+
+    @Test
+    @Ignore("Remove to run test")
+    public void simpleMatrix() {
+        String input = "ABC\n" +
+                       "123";
```
FridaTveit: These are very easy to read now. Great! :)

PR #606, opened 2017-05-26
change: remove "final" from test class declaration

morrme:
Closes: #571

---

Reviewer Resources:


stkent:
❤️  I think we're done with this cleanup! I'll open a new PR soon to note this policy regarding use of the `final` keyword (I'll try to capture some more general underlying principal of which this is a specific example).

PR #609, opened 2017-05-28
clock: prefer assert(Not)Equals to assert(True|False)

stkent:
e.g. https://stackoverflow.com/questions/20990524/asserttrue-vs-assertequals-for-ints

---

Reviewer Resources:

### DIFF
--- a/exercises/clock/src/test/java/ClockEqualTest.java
+++ b/exercises/clock/src/test/java/ClockEqualTest.java
@@ -1,98 +1,99 @@
 import org.junit.Ignore;
 import org.junit.Test;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertEquals;
 
 public class ClockEqualTest {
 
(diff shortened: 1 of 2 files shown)


exercises/clock/src/test/java/ClockEqualTest.java
```diff
@@ -1,98 +1,98 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertEquals;
 
 public class ClockEqualTest {
 
     @Ignore("Remove to run test")
     @Test
     public void clocksWithSameTimeAreEqual() {
-        assertTrue(new Clock(15, 37).equals(new Clock(15, 37)));
+        assertEquals(new Clock(15, 37), (new Clock(15, 37)));
```
FridaTveit: I agree that this is a good change :) But it might make it harder for people to see what they need to implement to make the test pass. Maybe it's worth adding a `HINT.md` file explaining a bit about overriding the `equals` method?

stkent: Good call! I added a hint file; let me know what you think of the content :)

PR #617, opened 2017-05-30
word-search: add to track

stkent:
[x-common directory](https://github.com/exercism/x-common/tree/73ebd13032219888ee29374b648b9a769d536549/exercises/word-search)

---

Reviewer Resources:


exercises/word-search/src/test/java/WordSearcherTest.java
```diff
@@ -0,0 +1,260 @@
...
+        wordSearcher = new WordSearcher();
+    }
+
+    @Test
+    public void testLocatesWordsWrittenLeftToRight() {
+        Map<String, Optional<WordLocation>> expectedLocations = new HashMap<>();
+        expectedLocations.put("clojure", Optional.of(new WordLocation(new Pair(1, 10), new Pair(7, 10))));
+
+        final Map<String, Optional<WordLocation>> actualLocations = wordSearcher.search(
+                expectedLocations.keySet(),
```
stkent: Not sure how I feel about this... inviting comments!

stkent:
I helped write the canonical data for this exercise, and at the time I advocated for this "building up" of the search list as the tests progress; however, after implementing, I'm not convinced. In a real system, if e.g. search left to right behavior was to suffer from a regression, every single test would break at once. Other tracks (pre- canonical data) had passed in a single-item list for all but the final test, and then had a final check that all words could be found simultaneously. There's still some crossover there, but less (max 2 tests fail if any search direction suffers from a regression). 🤷‍♂️

config.json
```diff
@@ -241,6 +241,11 @@
       "topics": []
     },
     {
+      "slug": "word-search",
+      "difficulty": 1,
```
stkent: Should be a 7

exercises/word-search/src/test/java/WordSearcherTest.java
FridaTveit: I think keeping it like this is okay :) It might be a bit clearer to declare the set separately and name it `wordsToSearchFor` or something. But it's fine as it is :)

exercises/word-search/src/example/java/Pair.java
```diff
@@ -0,0 +1,39 @@
...
+    @SuppressWarnings("SimplifiableIfStatement")
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Pair pair = (Pair) o;
+
+        if (x != pair.x) return false;
+        return y == pair.y;
```
stkent: It can, nice spot! (IntelliJ generated this implementation for me 😇.)

stkent:
@FridaTveit updated!

exercises/word-search/src/example/java/Pair.java
FridaTveit: That's strange that it would generate it like that... So do you want to simplify it? :)

stkent: Whoops, yes, thanks!

stkent:
Updated, building now!

PR #628, opened 2017-06-04
ocr-numbers: estimate difficulty, position

### LINKED ISSUE #341: ocr-numbers: estimate difficulty and move to appropriate position within track
By default, new exercises are scored with difficulty 1 and located at the end of the track. We should use this issue to discuss and determine an estimated difficulty for `ocr-numbers`, then submit an associated PR that moves it to the appropriate position within the track based on that difficulty.

stkent:
Closes #341.

@FridaTveit I think this could be a 7 or an 8 but it is definitely easier than `rectangles`, which involves searching a grid for rectangles of arbitrary sizes and is currently also rated a 7. That might need to bump to an 8 or 9. The main difference is that in `ocr-numbers` we have an input grid of known-ish size and we're comparing against known digit representations; in `rectangles` we have arbitrarily-sized input (IIRC) and arbitrarily-sized rectangles.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -271,11 +271,6 @@
       "topics": []
     },
     {
-      "slug": "rectangles",
-      "difficulty": 7,
-      "topics": []
-    },
-    {
       "slug": "meetup",
       "difficulty": 7,
       "topics": []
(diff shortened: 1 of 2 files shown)


FridaTveit:
I think this should be an 8, so maybe `rectangles` should be moved to 8 or 9 then? Do you agree? Would you be happy making that change as part of this review? :)

stkent:
Yup, happy to make that adjustment right now!

Issue #629, opened 2017-06-07
isogram: test fails on Windows laptop


FridaTveit:
`testWorksWithGermanLetters` fails on my Windows laptop when run against the example implementation.

stkent:
Closed by #630.

PR #630, opened 2017-06-07
isogram: add hint for german umlauts encoding

FridaTveit:
Fixes issue #629. Setting an environment variable fixed the problem for me so I thought it was worth documenting this in a HINT.md file in case anyone else has this problem.

---

Reviewer Resources:

### DIFF
--- a/exercises/isogram/HINT.md
+++ b/exercises/isogram/HINT.md
@@ -1,0 +1,10 @@
+If you find that `testWorksWithGermanLetters` fails even though you think you've implemented it correctly,
+this could be because of the [JVM encoding](https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_61/rzaha/international.htm). 
+Try setting the `JAVA_TOOL_OPTIONS` environment variable to `-Dfile.encoding=UTF8` to fix this.
+- To set an environment variable on Windows:
+    1. Go to Computer -> Properties -> Advanced System Settings
+    2. Add a new environment variable
+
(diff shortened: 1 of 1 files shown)


exercises/isogram/HINT.md
```diff
@@ -0,0 +1,7 @@
+If you find that `testWorksWithGermanLetters` fails even though you think you've implemented it correctly,
+this could be because of the [JVM](https://en.wikipedia.org/wiki/Java_virtual_machine) encoding. Try setting this encoding in the environment variables on
+your machine:
+1. Go to Computer -> Properties -> Advanced System Settings
```
stkent: If I understand correctly, this could also be an issue on non-Windows machines, where environment variables are set differently. Should we (a) give per-platform instructions here, (b) give no implementation instructions here, or (c) do something else that I haven't thought of?

exercises/isogram/HINT.md
```diff
@@ -0,0 +1,7 @@
+If you find that `testWorksWithGermanLetters` fails even though you think you've implemented it correctly,
+this could be because of the [JVM](https://en.wikipedia.org/wiki/Java_virtual_machine) encoding. Try setting this encoding in the environment variables on
```
stkent: At this point, I find myself wondering "what does it mean for the JVM to have an encoding set?" Is it possible to succinctly answer that question in this hint, do you think?

exercises/isogram/HINT.md
FridaTveit: I'll try to find per-platform instructions :)

PR #634, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.


stkent:
@kytrinyx if you could please rebase this against master, the CI build should now pass!

PR #644, opened 2017-06-14
Clean up root build script

stkent:
@FridaTveit I spent some time poking around this file to make sure I understood it well. Let me know if you think the re-organization and extra commentary add value or just noise!

---

Reviewer Resources:

### DIFF
--- a/exercises/build.gradle
+++ b/exercises/build.gradle
@@ -1,44 +1,67 @@
-subprojects { project ->
-  apply plugin: "java"
+// This root Gradle project is used for:
+//
+// - Checking that every test suite compiles against its corresponding example implementation;
+// - Checking that every example implementation passes its corresponding test suite;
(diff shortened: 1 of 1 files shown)


exercises/build.gradle
```diff
@@ -1,44 +1,67 @@
...
   // Note that journey-test.sh also strips @Ignore annotations using sed. The
   // stripping implementations here and in journey-test.sh should be kept
   // consistent.
   task copyTestsFilteringIgnores(type: Copy) {
     from "src/test/java"
-    into "build/gen/test/java"
+    into generatedTestSourceDir
     filter { line ->
-      line.contains("@Ignore") ? "" : line
+      line.contains("@Ignore") ? null : line
```
stkent: `null` here just skips the line, rather than including an unnecessary blank line.

Issue #642, opened 2017-06-14
Remove final from starter implementations


FridaTveit:
We have decided to remove the `final` keyword from class declarations in starter implementations as it could be confusing to people not familiar with Java. See issue #640 and issue #571 for discussion.

The exercises with `final` in the starter implementation are:
- [ ] acronym
- [ ] all-your-base
- [ ] bracket-push
- [ ] change
- [ ] diamond
- [ ] flatten-array
- [ ] luhn
- [ ] minesweeper
- [ ] ocr-numbers
- [ ] perfect-numbers
- [ ] queen-attack (both `QueenAttackCalculator` and `BoardCoordinate`)
- [ ] rectangles
- [ ] robot-simulator (both `Robot` and `GridPosition`)
- [ ] saddle-points (both `Matrix` and `MatrixCoordinate`)
- [ ] scrabble-score
- [ ] secret-handshake
- [ ] sublist
- [ ] wordy

vivshaw:
Should final be removed from just the starter class in `main`, or also removed from the example implementation in `example`?

stkent:
Removing from the starter class only is perfect! Thanks for taking this on :)

The example implementations are for track maintenance purposes only, so the authors of these solutions are free to write code as they normally would :)

stkent:
Closed by #650 💪

PR #650, opened 2017-06-17
Removing final from starter classes

vivshaw:
Here's some work toward #642. Wasn't sure whether you wanted `final` removed from the `example` implementations also so I didn't do that yet, but I will gladly zip through those ones too if you want.

---

Reviewer Resources:


stkent:
This looks great, @vivshaw! Thanks for clarifying the scope of the issue too :) Merging! 💯

PR #652, opened 2017-06-17
difference-of-squares: return doubles

### LINKED ISSUE #331: difference-of-squares: should methods return double instead of int?
when doing this exercise, I noticed that the test fixture expects that the return value of the 
```computeSquareOfSumTo```, ```computeSumOfSquaresTo```,  and ```betweenSquareOfSumAndSumOfSquaresTo``` to be an int. Should these return a ```double``` instead?

For example, running ```Difference.computeSumOfSquaresTo(Interger.MAX_VALUE)``` will cause an overflow.

Would be happy to submit a patch as well as an additional test for this.

stkent:
Closes #331.

Matches [v1.1.0 of the canonical tests](https://github.com/exercism/x-common/blob/7a1108b24e06142495e135a15317a3b76769026e/exercises/difference-of-squares/canonical-data.json), and adds test cases with input 2000 (which exercises the `double` return type of the adjusted methods).

I will suggest this change upstream, but since it exercises something that may be type system specific, I wouldn't be surprised if it isn't accepted there.

I think this is a valuable change since it follows the pattern established by other Java math functions.

---

Reviewer Resources:

### DIFF
--- a/exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
+++ b/exercises/difference-of-squares/src/example/java/DifferenceOfSquaresCalculator.java
@@ -2,18 +2,19 @@ import java.util.stream.IntStream;
 
 final class DifferenceOfSquaresCalculator {
 
-    int computeSquareOfSumTo(final int input) {
-        final int sum = input * (input + 1) / 2;
-        return (int) Math.pow(sum, 2);
+    double computeSquareOfSumTo(final int input) {
+        final double sum = input * (input + 1) / 2;
+        return Math.pow(sum, 2);
     }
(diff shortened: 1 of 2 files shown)


stkent:
Please do not merge just yet!

stkent:
So based on upstream [feedback](https://github.com/exercism/problem-specifications/pull/831), in which it was highlighted that other exercises exist specifically to introduce big number handling, I'm going to close this PR and the corresponding issue without merging. If we want to implement `grains`, let's dig around the change that marked it foregone (if that's still accessible) to see what the original thought process was there, and maybe reevaluate.

PR #662, opened 2017-06-22
bank-account: add to track

FridaTveit:
[Description](https://github.com/exercism/problem-specifications/blob/master/exercises/bank-account/description.md)

---

Reviewer Resources:


stkent:
[FYI: Kotlin has [an implementation of this exercise](https://github.com/exercism/kotlin/tree/master/exercises/bank-account) that might be inspirational, but I'm not sure if it's compliant with any canonical data!]

config.json
```diff
@@ -374,6 +374,13 @@
       ]
     },
     {
+    	"slug": "bank-account",
+    	"difficulty": 6,
+    	"topics": [
+
+    	]
+    }
```
stkent: Missing comma here!

exercises/bank-account/src/test/java/BankAccountTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @Test
+    public void newlyOpenedAccountHasEmptyBalance() throws BankAccountActionInvalidException {
+        bankAccount.open();
+
+        assertEquals(0, bankAccount.getBalance());
+    }
+
+    @Ignore
```
stkent: All `@Ignore` annotations should include the standard hint: `"Remove to run test"` :)

exercises/bank-account/src/test/java/BankAccountTest.java
```diff
@@ -0,0 +1,201 @@
...
+    @Test
+    public void canAdjustBalanceConcurrently() throws BankAccountActionInvalidException, InterruptedException {
+        bankAccount.open();
+        bankAccount.deposit(1000);
+
+        for (int i = 0; i < 10; i++) {
+            adjustBalanceConcurrently();
```
stkent: Does this test fail 100% of the time? If so, would you mind explaining how that works? I'm curious!

stkent: Ah, I see from the hint it is not guaranteed to fail. Can we provide any recommendation as to how many times a user should run this test to be "reasonably certain" their implementation is threadsafe? Alternatively, can we inject a little more thread control into solutions to be able to pull the appropriate levers ourselves and guarantee a concurrent modification attempt?

exercises/bank-account/src/test/java/BankAccountTest.java
FridaTveit: Sorry, forgot! Will fix :)

exercises/bank-account/src/test/java/BankAccountTest.java
FridaTveit: It failed every time I ran it (I think I ran it about 10 times in a row) with a non-threadsafe implementation. Is there any way we can be absolutely guarantee a concurrent modification attempt? I would expect the test to fail every time if their implementation isn't thread safe, but I can't be 100% certain. I could put in the hint that they should run the test a couple of times just to be safe?

stkent: Cool, I see the same behavior here. Let's add to the hint just to be safe :)

FridaTveit: Will do :)

stkent:
Love it! Great add.

PR #663, opened 2017-06-22
two-fer: add to track

Smarticles101:
---

Reviewer Resources:


Smarticles101:
I was thinking since this is meant to be right after hello world, should we add any readme files that the user might need to read? The description of this exercise introduces test driven development - would it be appropriate to remove `@Ignore`s from the hello-world exercise and introduce them here instead?

stkent:
@Smarticles101 you are right that the purpose of adding this exercise was to allow Hello, World to become as simple as possible, thereby testing user environment, etc. Removing Ignores from Hello, World, and introducing them in this exercise instead supports that goal, so I say 👍

Smarticles101:
How does this look @stkent?

config.json
```diff
@@ -12,6 +12,11 @@
       ]
     },
     {
+      "slug": "two-fer",
+      "difficulty": 1,
+      "topics": []
```
stkent: This topics entry should be formatting like the others (with a blank line between the brackets).

stkent:
I think this update to the documentation is good for now, though looking at the content of the two exercises it is clear that we should separately split that content between the two exercises.

Ref. https://github.com/exercism/problem-specifications/issues/520, https://github.com/exercism/problem-specifications/blob/master/exercises/hello-world/description.md, https://github.com/exercism/problem-specifications/issues/548

stkent:
I'm happy to take on the separate work I mention after this PR is merged.

stkent:
@Smarticles101 just one change to make in the config formatting, then LGTM!

Smarticles101:
@stkent ah, so two-fer is a replacement for what hello-world was, and hello-world should just become the traditional hello world.

stkent:
Yep, exactly :)

stkent:
Ok, this is ready to merge, I'm just going to wait on hitting the button until I have 30 mins right after to update the documentation around this and hello-word :)

stkent:
I'll be taking care of this later this afternoon.

PR #666, opened 2017-06-26
Updates now that two-fer is part of the track!

stkent:
Overview:

- Renamed HINT.md -> HINTS.md track wide so that hints are actually included with the README as they are supposed to be!
- Updated `hello-world` to be as dumb as possible, which was why `two-fer` was introduced in the first place.
- Removed short getting started guide from `hello-world` - I felt that it was unclear where one should actually start with that exercise (README? TUTORIAL? GETTING_STARTED?) and that maintaining the duplicated-but-different information between the tutorial and the getting started page was unnecessary overhead.
- Updated the TUTORIAL.md for `hello-world` for the new simpler exercise; removed all the steps related to TDDing this exercise since there's now only one test!
- Moved TDD-related discussion to the HINTS file for `two-fer`, and made it more of a high-level discussion of how to TDD than previously.

It's quite likely that I made some mistakes here and there in this conversion. Please review closely! A good strategy may be to run the journey test locally, which will fetch copies of these exercises as they will appear to users.

---

Reviewer Resources:


exercises/hello-world/TUTORIAL.md
```diff
@@ -503,39 +282,11 @@ Here's an up-to-date list of submissions on the Java track:
...
-
 ## Contribute to Exercism
 
-The entire of Exercism is Open Source and is the labor of love for over
+The entire of Exercism is Open Source and is a labor of love for over
 100 maintainers and many more contributors.
 
 A starting point to jumping in can be found here:
 
 https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md
```
FridaTveit: This is a different link to the one I've used in the general read me: https://github.com/exercism/docs/tree/master/contributing-to-language-tracks. We should probably be consistent and stick to one of them :P Yours looks easier to read since it's one file instead of many. But the [porting an exercise](https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md#porting-an-exercise-to-another-language-track) link is broken, and that's one of the most useful bits of the documentation in my opinion.

stkent: Good question! I asked upstream (https://github.com/exercism/docs/issues/61) and will wait and see what response I get there before committing one way or the other :)

stkent:
@FridaTveit updated, building now!

PR #668, opened 2017-06-27
JAVA -> Java

wolf99:
Java is not an acronym or an abbreviation, it should not be written in all-caps (doing so encourages the misunderstanding further, especially as exerscism.io is aimed at newbies!).

---

Reviewer Resources:


stkent:
Absolutely agree with the reasoning; thanks for taking the time to patch this up! LGTM.

Issue #325, opened 2017-02-24
list-ops: estimate difficulty and move to appropriate position within track


stkent:
The exercise `list-ops` was added after we'd already re-ordered the track. By default, new exercises are scored with difficulty 1 and located at the end of the track. We should use this issue to discuss and determine an estimated difficulty for `list-ops`, then submit an associated PR that moves `list-ops` to the appropriate position within the track based on that difficulty.

FridaTveit:
Without any starter implementation I found list-ops very hard because it was a challenge trying to reverse engineer the signatures. So as it is I would give it a difficulty of 9 or 10. I don't think figuring out what the signatures are should be the challenging part of the exercise though. So I think the exercise could be improved by adding a starter implementation and possibly explaining what the different functions are meant to do in the README (or adding links to explanations). If we did that then I think the exercise wouldn't be so difficult, maybe 7?

stkent:
Closed by #670.

PR #670, opened 2017-06-30
list-ops: estimate difficulty 9

FridaTveit:
See issue #325.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -495,17 +495,17 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


stkent:
I think this is right for now, based on your reasoning in #325; I plan to port the updated `list-ops` implementation from Kotlin over to Java soon, and will propose a new estimated difficulty and position when I do so!

PR #671, opened 2017-06-30
list-ops: update to match Kotlin track

stkent:
Based on v1.0.0 of the canonical data:

https://github.com/exercism/problem-specifications/tree/28a32203b96ebaecac6a6968831bd66718d8ba30/exercises/list-ops

Plus modifications that are being discussed in

https://github.com/exercism/problem-specifications/issues/826

---

Reviewer Resources:

### DIFF
--- a/exercises/list-ops/HINTS.md
+++ b/exercises/list-ops/HINTS.md
@@ -1,0 +1,3 @@
+## Hints
+
+The `foldLeft` and `foldRight` methods are "fold" functions, which is a concept well-known in the functional programming world, but less so in the object-oriented one. See the Wikipedia page on folding for [general background](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) and [signature/implementation hints](https://en.wikipedia.org/wiki/Fold_(higher-order_function)#Linear_folds).
--- a/exercises/list-ops/src/main/java/ListOps.java
+++ b/exercises/list-ops/src/main/java/ListOps.java
@@ -1,0 +1,44 @@
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+class ListOps {
+
+    static <T> List<T> append(List<T> list1, List<T> list2) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T> List<T> concat(List<List<T>> listOfLists) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T> List<T> filter(List<T> list, Predicate<T> predicate) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T> int size(List<T> list) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T, U> List<U> map(List<T> list, Function<T, U> transform) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T> List<T> reverse(List<T> list) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T, U> U foldLeft(List<T> list, U initial, BiFunction<U, T, U> f) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T, U> U foldRight(List<T> list, U initial, BiFunction<T, U, U> f) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    private ListOps() {
+        // No instances.
+    }
+
+}
--- a/exercises/list-ops/src/test/java/ListOpsTest.java
+++ b/exercises/list-ops/src/test/java/ListOpsTest.java
@@ -1,316 +1,186 @@
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.function.BiFunction;
-import java.util.function.BinaryOperator;
-import java.util.function.Predicate;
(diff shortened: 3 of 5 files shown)


exercises/list-ops/HINTS.md
```diff
@@ -0,0 +1,3 @@
+## Hints
```
stkent: Not sure if we do this in other HINTS files, but I like this!

exercises/list-ops/src/main/java/ListOps.java
```diff
@@ -0,0 +1,44 @@
...
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static <T, U> U foldRight(List<T> list, U initial, BiFunction<T, U, U> f) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    private ListOps() {
+        // No instances.
+    }
```
stkent: @FridaTveit would this have provided enough of a starter for you when working on this exercise, do you think? I decided not to add in explicit descriptions of the operations because I think the tests + the type signatures make it possible to deduce the purpose of each operation now?

exercises/list-ops/src/test/java/ListOpsTest.java
```diff
@@ -1,316 +1,186 @@
...
 import java.util.List;
-import java.util.function.BiFunction;
-import java.util.function.BinaryOperator;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-import static junit.framework.TestCase.assertEquals;
-import static junit.framework.TestCase.assertFalse;
-import static junit.framework.TestCase.assertNotNull;
-import static junit.framework.TestCase.assertTrue;
```
stkent: JUnit 3 ‼️

exercises/list-ops/src/test/java/ListOpsTest.java
```diff
@@ -1,316 +1,186 @@
...
-        );
-        final List<Integer> expected
-                = Arrays.asList(0, 1, 2, 3, 4, 1, 2, 3, 4, 5, 6);
-        final List<Integer> actual = ListOps.concat(list1, list2);
-
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
+    @Test
+    public void testFoldLeftWithDirectionDependentOperationOnNonEmptyList() {
```
FridaTveit: How is this direction dependent? I understand that the operation in itself is but  with these numbers you would get the same answer no matter which direction you traverse the list wouldn't you? Wouldn't it be a better test if that wasn't the case?

exercises/list-ops/src/main/java/ListOps.java
stkent: The keep file was deleted in this PR already, I think?

exercises/list-ops/src/test/java/ListOpsTest.java
stkent: If folding from the left:

Step 0: accumulation = 5
Step 1: accumulation = 2 / 5 = 0
Step 2: accumulation = 0 / 5 = 0

If folding from the right:

Step 0: accumulation = 5
Step 1: accumulation = 5 / 5 = 1
Step 2: accumulation = 1 / 2 = 0

Hmm, yep, something funky here. Will look again.

stkent: Ok, so the difference between these two cases is due to some subtleties of how the fold function parameter is typed. In particular, note that it is of type

```
BiFunction<U, T, U>
```

for the left fold, and of type

```
BiFunction<T, U, U>
```

for the right fold. Since `T` is the type of the list entries, and `U` is the type of the accumulated value, this means that if we pass the anonymous function `(x, y) -> x / y` to both folds, we're actually performing slightly different calculations. See [this upstream comment and the ones following](https://github.com/exercism/problem-specifications/issues/826#issuecomment-307626155) for some relevant chatter. So what we're actually doing is the following:

---

If folding from the left:

Step 0: accumulation = 5
Step 1: accumulation = a / c = 5 / 2 = 2
Step 2: accumulation = a / c = 2 / 5 = 0

If folding from the right:

Step 0: accumulation = 5
Step 1: accumulation = c / a = 5 / 5 = 1
Step 2: accumulation = c / a = 2 / 1 = 2

I updated the function signatures in the test suite to use generic parameter labels `x` and `y` rather than `accumulation` and `current` to make it clearer that we're using the same anonymous function in both cases. It's up to the user to infer from the (provided) type signature how those parameters should be populated. I update the HINTS file to provide a direct link to the section of the fold Wikipedia page relevant to interpretation of these signatures.

stkent:
What do you think the new exercise difficulty should be, @FridaTveit? I'm thinking it might not need a huge adjustment since we introduce generics quite late, and this touches generics + functional programming concepts! 😱

exercises/list-ops/src/main/java/ListOps.java
FridaTveit: Yes, sorry, my mistake :)

exercises/list-ops/src/test/java/ListOpsTest.java
FridaTveit: Okay. So someone solving the exercise definitely can't traverse the list the wrong way and still pass this test?

FridaTveit:
That's true. Maybe change it to 8?

stkent:
Sounds good; incoming!

exercises/list-ops/src/test/java/ListOpsTest.java
FridaTveit: ?

stkent: I believe that is correct, based on how the anonymous function + method signature combines, yes :)

FridaTveit: Great :)

Issue #554, opened 2017-05-14
Remove starter implementations for exercises with positions 21 or greater


stkent:
Ref.: https://github.com/exercism/xjava/blob/master/POLICIES.md#starter-implementations

redshirt4:
Are the exercise positions (>= 21) referring to this ordered list? http://exercism.io/languages/java/exercises

Smarticles101:
No, they refer to the [config](https://github.com/exercism/java/blob/master/config.json) file

stkent:
(Those two lists are generally in sync; the website might lag behind the config file by a few days is all :))

Smarticles101:
Lol yeah I just realized that the site follows the order in the config file.

Smarticles101:
Closed by #681

PR #677, opened 2017-07-03
spiral-matrix: add to track

stkent:
[problem specification directory](https://github.com/exercism/problem-specifications/tree/57aa5f082262096d2bc5c983388ed5a00dc73822/exercises/spiral-matrix)

---

Reviewer Resources:


config.json
```diff
@@ -285,6 +285,13 @@
       ]
     },
     {
+      "slug": "spiral-matrix",
+      "difficulty": 6,
```
stkent: Best guess.

exercises/spiral-matrix/src/main/java/SpiralMatrixBuilder.java
```diff
@@ -0,0 +1,5 @@
+class SpiralMatrixBuilder {
+
+
+
+}
```
stkent: Check: is this needed per policy?

exercises/spiral-matrix/src/example/java/SpiralMatrixBuilder.java
```diff
@@ -0,0 +1,27 @@
+class SpiralMatrixBuilder {
+
+    int[][] ofSize(int size) {
+        if (size == 0) return new int[][]{};
```
FridaTveit: I would add curly brackets and put the `if` body on a new line, just to be safe :)

exercises/spiral-matrix/src/main/java/SpiralMatrixBuilder.java
FridaTveit: No, according to the policy after the first 20 exercises no starter implementation needs to be provided. But quite a few exercises after the first 20 have an empty starter implementation like this. Do you think that's okay or should we make them more consistent?

exercises/spiral-matrix/src/example/java/SpiralMatrixBuilder.java
```diff
@@ -0,0 +1,27 @@
+class SpiralMatrixBuilder {
+
+    int[][] ofSize(int size) {
```
FridaTveit: Maybe `buildMatrixOfSize()` would be a more descriptive name?

exercises/spiral-matrix/src/main/java/SpiralMatrixBuilder.java
stkent: I think there's an open issue for cleanup; let's decide there whether we want no implementation or empty class(es)! https://github.com/exercism/java/issues/554

exercises/spiral-matrix/src/example/java/SpiralMatrixBuilder.java
stkent: Kotlin habits sneaking in ;)

PR #681, opened 2017-07-04
Remove starter implementations

redshirt4:
Began implementing changes for Issue #554, currently incomplete.

I would appreciate an in-progress review to ensure implementation matches coding standard. I tried to follow the spirit of the starter implementation policy, but am unsure if empty classes are considered a stub.

---

Reviewer Resources:


exercises/kindergarten-garden/src/main/java/Plant.java
```diff
@@ -1,21 +0,0 @@
-public enum Plant {
```
Smarticles101: I'm pretty sure this file needs to stay here as it is needed to complete the exercise.

FridaTveit: Yes, this should still be here. When we use enums we usually include this in the starter implementation regardless of the difficulty of the exercise. This is because they're not really part of the solution to the problem, they're more part of the problem specification if that makes sense :) Sorry for the confusion @redshirt4! :)

exercises/collatz-conjecture/src/test/java/CollatzCalculatorTest.java
```diff
@@ -23,43 +23,43 @@ public void setUp() {
 
     @Test
     public void testZeroStepsRequiredWhenStartingFrom1() {
-        assertEquals(0, collatzCalculator.computeStepCount(1));
+//        assertEquals(0, collatzCalculator.computeStepCount(1));
```
FridaTveit: As per the discussion in PR #683, we decided not to comment out tests after all as this might be confusing to people solving the exercise. The [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md) has been updated to reflect this. So all you have to do is to remove the `CollatzCalculator.java` file completely and replace it with a `.keep` file (the `.keep` file basically tells git not to ignore that directory even though it's empty). Sorry for the confusion! :)

redshirt4:
Thank you @Smarticles101 and @FridaTveit for the clarifications. I'll take a look at the new spec and push.

redshirt4:
Thank you for the review @Smarticles101 and @FridaTveit . Is it best practice to ask for a re-review of current, or to open a new pull request?

Smarticles101:
@redshirt4 I think it is best practice to ask for a re-review or to wait for another review.

exercises/etl/src/main/java/Etl.java
```diff
@@ -3,6 +3,6 @@
 
 public class Etl {
     public Map<String, Integer> transform(Map<Integer, List<String>> old) {
-        return null;
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
```
Smarticles101: The fact that you didn't delete this starter implementation because it had a complex stub shows that you read and paid attention to the policies doc which means alot

exercises/saddle-points/src/main/java/MatrixCoordinate.java
```diff
@@ -1,31 +0,0 @@
-class MatrixCoordinate {
```
FridaTveit: Please don't remove this file. Like the occasional enum definitions and exception classes in other exercises, this has been provided as part of the solution so that the exercise isn't overly complicated to implement :) You also don't need a `.keep` file if you keep this in.

FridaTveit:
Sorry @Smarticles101, I didn't notice you were reviewing this at the same time! I was trying to do the review on a train which only occasionally has internet :P I don't think it matters very much about `MatrixCoordinate`. But since it requires overriding `equals` and `hashCode` it might be nice to leave it in to reduce complexity, especially since saddle-point only has difficulty 4. What do you think?

FridaTveit:
Sorry @Smarticles101, I didn't notice you were reviewing this at the same time! I was trying to do the review on a train which only occasionally has internet :P I don't think it matters very much about `MatrixCoordinate`. But since it requires overriding `equals` and `hashCode` it might be nice to leave it in to reduce complexity, especially since saddle-point only has difficulty 4. What do you think?

Smarticles101:
@FridaTveit I agree. I missed that file on accident. But yes, it should be kept

redshirt4:
@FridaTveit changes pushed to address review kickbacks

Issue #523, opened 2017-05-10
Add exercise topics to config


Smarticles101:
I noticed today that none of the exercises in the config have any topics listed.  Not an urgent issue, but it might be good to add some.  

~~Should make for a super easy fix because the [xcsharp](https://github.com/exercism/xcsharp/blob/master/config.json) track already has valid topics listed in their config.  IMO this would be a good `good first patch` issue.~~

Upon further reflection, I have realized that it may not be good to just copy topics from other tracks.  Rather, I think it would be better for contributors to come up with and agree on topics for exercises.

For reference, valid topics are listed [here](https://github.com/exercism/x-common/blob/master/TOPICS.txt)

jtigger:
Yes, @Smarticles101... will love to incorporate topics.  Yeah, and I agree that each language may trigger different topics depending on their features (that said, C# and Java are _very_ similar in feature set).  Also, I've noticed that all that can _really_ be said about an **exercise** is what kind of problem it is.  It's the **solutions** that invoke different language features — how people choose to solve them.

I am not aware of a feature that draws on this data, either.  I suspect that the redesign will directly address the gap between a series of exercises and an actual curriculum. 

For more context, digest the discussion: "Problem ordering, topics covered, and understanding of concepts required" — exercism/discussions#60.

Smarticles101:
This is going to become relevant as soon as #704 is merged.

FridaTveit:
We should start working on this now PR #704 is merged :)

stkent:
Is there upstream consensus yet on how we should represent topics, @kytrinyx? https://github.com/exercism/discussions/issues/167 seems to suggest not?

kytrinyx:
No upstream consensus yet. I'd appreciate it if you weigh in on that discussion if you have opinions.

stkent:
Thanks for confirming; I don't have strong opinions on the mechanism used to apply labels, just wondering if we should wait on adding labels here in case the mechanism changes substantially.

kytrinyx:
Yeah, that seems reasonable.

stkent:
Putting this back on hold while upstream decisions get made.

FridaTveit:
Do you think we can start working on this now @exercism/java?

stkent:
I have seen some activity around the topics but I've not been following closely; if they are closer to canonicalized upstream and we know what the strategy is regarding tagging (tag with all topics that are unavoidable or tag with all topics that may be encountered based on implementation choices) then 👍

[...]

PR #682, opened 2017-07-04
circular-buffer: add to track

vivshaw:
[test spec directory](https://github.com/exercism/problem-specifications/tree/master/exercises/circular-buffer)

---

Reviewer Resources:


FridaTveit:
Thanks for contributing @vivshaw! :D Sorry for messing with the PR name, it's just to make it consistent so that it's easy to see when someone is working on a new exercise :) Once you're ready for it to be reviewed, feel free to change it to "circular-buffer: add to track". That way it's easy for us to tell when you want a review so that we don't start leaving comments while you're still working on it :)

vivshaw:
Alright, finished implementing all canonical test cases & the build passes.

The one big question I have is, for the exceptions that we want the CircularBuffer to throw when reading an empty buffer, should we use an existing exception type (and if so, which), or expect the user to implement a custom BufferEmptyException and BufferFullException? I went with the latter, because learning how to implement your own exceptions is useful, but I'm worried it'll make the exercise overly complex to newcomers who are inexperienced with exception handling.

FridaTveit:
What I've done before is make those exception classes and including them in both the example implementation and the starter implementation (see the bank account exercise for an example). That way the users get experience with custom exceptions but, like you said, the exercise doesn't get too complex :)

config.json
```diff
@@ -467,6 +467,15 @@
       ]
     },
     {
+      "slug": "circular-buffer",
+      "difficulty": 8,
+      "topics": [
+        "Queues",
+        "Classes",
+        "Exceptions"
```
Smarticles101: None of the other exercises have topics at the moment. This has come up recently in #384 and #523 thoughts @exercism/java?

vivshaw:
@FridaTveit Alright, I've just done that & updated the test suite accordingly. Let me know if anything else needs changed.

exercises/circular-buffer/src/test/java/CircularBufferTest.java
```diff
@@ -0,0 +1,161 @@
+import static junit.framework.TestCase.assertEquals;
```
Smarticles101: All the other exercises use `import static org.junit.Assert.assertEquals;`

exercises/circular-buffer/src/test/java/CircularBufferTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class CircularBufferTest {
+    @Rule
```
Smarticles101: Usually there is a blank line after the class declaration.

config.json
FridaTveit: I think it should be fine for people to add topics when they implement a new exercise if they want to, as long as they're valid topics as specified [here](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) :) What do you think @stkent?

exercises/circular-buffer/src/example/java/BufferIOException.java
```diff
@@ -0,0 +1,5 @@
+public class BufferIOException extends RuntimeException {
```
FridaTveit: This should extend `Exception` instead of `RuntimeException`. `RuntimeException` should be for errors  you can't fix, generally programmer errors where the best thing to do is for the program to crash. If you want the exception to be handled then extend `Exception` because the compiler will force you to handle the exception. See [this article](http://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/) for some more information about it :)

config.json
stkent: It looks like we're [going to be adding topics for every exercise soon](https://github.com/exercism/discussions/issues/159), so I'm a +1 to what @FridaTveit suggested :)

>In order to prepare your track for the new track progression model, you'll need to do the following:
>
>- **add topics to each exercise**. We have a list of topics that you can use for inspiration, but this list is not exhaustive. The topics you choose should make sense for your language.
>- add a rough guess at the difficulty of each exercise.
>- choose 10-20 exercises to mark as core
>- give the core exercises an order
>- for all other exercises, mark them as `unlocked_by` one of the core exercises (identified by the slug)

While we shouldn't embark on the above en masse just yet, adding topics is therefore good future-proofing!

vivshaw:
Thanks for the feedback! I've just finished implementing all the changes @Smarticles101 and @FridaTveit requested.

@FridaTveit, thanks in particular for jogging my memory on `Exception` vs. `RuntimeException`. I was having trouble getting `Exception` to do what I expected, until I took a look at your bank-account exercise & remembered I have to `throw` in all the tests too & not just the implementation!

exercises/circular-buffer/src/test/java/CircularBufferTest.java
```diff
@@ -0,0 +1,162 @@
...
+        buffer.read();
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void canReadItemJustWritten() throws BufferIOException {
+        CircularBuffer<Integer> buffer = new CircularBuffer<>(1);
+
+        buffer.write(1);
+        assertEquals((int) buffer.read(), 1);
```
FridaTveit: Also these are the wrong way around. It's `assertEquals(expected, actual)`. I get that wrong all the time too :P

Smarticles101: I thought the same thing about the cast the first time I looked at it so I ran it and found out why. 

```
error: reference to assertEquals is ambiguous
        assertEquals(buffer.read(), 1);
        ^
  both method assertEquals(long,long) in Assert and method assertEquals(Object,Object) in Assert match
```

vivshaw: @FridaTveit, the int cast was indeed due to the error that @Smarticles101 posted. I wrestled with that int cast for a couple hours & just now figured out a) why it does that and b) how to avoid it! It seems to be due to some wackiness with how Java autoboxing & unboxing interacts with target type inference.

The problem goes like this: Since Java `Collections` can't store raw types, any method we use to pull values out of our data structure will return `Integer` rather than `int`. So, my line `return buffer.remove()` returns an `Integer`. This then becomes an argument to `assertEquals()`, but the other argument is an `int`. `assertEquals()` is overloaded, but doesn't have any implementation with the type signature `Integer, int`, so it's stuck deciding between its two best guesses for the type signature: `long, long` and `Object, Object`.

So if we stick with `assertEquals()`, we can either: 1) remove the cast, in which case any users trying to implement the exercise with a `Collection` may run into this confusing bug, or 2) kludge both the arguments to the same type with a cast so they're both unboxed or both boxed.

But, `assertThat()` to the rescue! Turns out that `asssertThat(Integer i, is(int j))` is unambiguously typed & neatly sidesteps the problem. So I redid the tests with that & we should be golden, I think.

PR #685, opened 2017-07-04
Update policy around language constructs

stkent:
@Smarticles101 @FridaTveit please check!

---

Reviewer Resources:

### DIFF
--- a/POLICIES.md
+++ b/POLICIES.md
@@ -40,9 +40,11 @@ References: [[1](https://github.com/exercism/java/issues/177#issuecomment-261291
 
 References: [[1](https://github.com/exercism/java/issues/178)], [[2](https://github.com/exercism/java/pull/683#discussion_r125506930)]
 
-### Do not make user-facing classes `final`
+### Minimize initial complexity in user-facing code
 
-> Starter implementations and test classes should not include the `final` keyword in class declarations as it isn't strictly necessary and could be confusing to people unfamiliar with Java.
+> Minimize (within reason) the number of Java language constructs a user is exposed to simultaneously at the beginning of the track. Instead, introduce those constructs with examples and hints in an appropriate later exercise. Specific applications of this policy include:
+> - avoiding use of the `public` keyword in user-facing class declarations, method signatures, and parameter signatures;
+> - avoiding use of the `final` keyword in user-facing class declarations, method signatures, and parameter signatures.
 
 ### Ignore noninitial tests
 
--- a/POLICIES.md
+++ b/POLICIES.md
@@ -17,9 +17,8 @@ Our policies are not set-in-stone. They represent directions chosen at a point i
 
 | Track Event | Policies to review |
 |:------------|:-----------------|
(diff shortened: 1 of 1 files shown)


POLICIES.md
```diff
@@ -40,9 +40,11 @@ References: [[1](https://github.com/exercism/java/issues/177#issuecomment-261291
 
 References: [[1](https://github.com/exercism/java/issues/178)], [[2](https://github.com/exercism/java/pull/683#discussion_r125506930)]
 
-### Do not make user-facing classes `final`
+### Minimize initial complexity in user-facing code
 
-> Starter implementations and test classes should not include the `final` keyword in class declarations as it isn't strictly necessary and could be confusing to people unfamiliar with Java.
+> Minimize (within reason) the number of Java language constructs a user is exposed to simultaneously at the beginning of the track. Instead, introduce those constructs with examples and hints in an appropriate later exercise. Specific applications of this policy include:
+> - avoiding use of the `public` keyword in user-facing class declarations, method signatures, and parameter signatures;
```
FridaTveit: In my opinion we're leaving out `public` not just because of minimising complexity (especially since `public` is more familiar to most people and therefore adds less complexity than `final`) but also because it's best practice to limit access to classes, methods and variables where possible. So if something can be made package private or private, then it should be. Do you agree? Is this worth mentioning here?

stkent: Good point; I think that's worth calling out in a separate policy entry (that we aim to follow best practices); the minimization of visibility would then be an instance of _that_ policy. If you concur, I'll make the change later today.

FridaTveit: Sounds good :)

stkent:
Pushed with proposed changes; let me know what you think! Writing this is hard :P

POLICIES.md
```diff
@@ -40,9 +39,14 @@ References: [[1](https://github.com/exercism/java/issues/177#issuecomment-261291
...
-### Do not make user-facing classes `final`
+### Avoid using final
 
-> Starter implementations and test classes should not include the `final` keyword in class declarations as it isn't strictly necessary and could be confusing to people unfamiliar with Java.
+> Avoid using `final` in user-facing code. Liberal use of the `final` keyword remains controversial. It adds little value to test suites, and if included in starter implementations, places implicit restrictions on user solutions.
+
+### Adhere to best practices
+
+> Ensure that all Java code adheres to the best practices listed below:
+> - minimize the accessibility of classes and members (Effective Java, item 13)
```
Smarticles101: Perhaps a link to an explanation of this would be nice.

stkent: Unfortunately there's no way to link to content of the book itself, as far as I know. I could link to some kind of proxy like http://jtechies.blogspot.com/2012/07/item-13-minimize-accessibility-of.html?

Smarticles101: Sounds good to me

stkent:
Updated and rebased!

stkent:
👍 thanks for the eyes, @Smarticles101!

Issue #410, opened 2017-04-04
Speed up test suite(s)


jtigger:
The CI build is approaching 20 minutes.  The long pole on the build is (understandably) the journey suite.  However, the exercises are guaranteed to be independent: the test suites can be executed in parallel without fear of test pollution.  The situation is ripe for parallel execute of tests.

Please update the CI build to minimize both the duration of the total test run _and_ minimize additional added complexity (i.e. readability of the build scripts themselves is paramount).

jtigger:
As a part of this please include the concerns surfaced in #326.

Smarticles101:
Is there some way of only running tests on the modified exercise(s)?  This would significantly improve the build time as usually only simple things are changed.

stkent:
[Currently](https://travis-ci.org/exercism/xjava/builds/232104547) the build is executing using Gradle 2.2.1. We should investigate upgrading to 3.5 (or whatever the latest stable version is).

stkent:
> [Currently](https://travis-ci.org/exercism/xjava/builds/232104547) the build is executing using Gradle 2.2.1. We should investigate upgrading to 3.5 (or whatever the latest stable version is).

See https://github.com/exercism/xjava/pull/599

stkent:
See https://github.com/exercism/xjava/pull/600 for partial parallelization of the build (unit tests).

stkent:
Things I think we could still investigate here:

- parallelizing journey test execution
- running unit tests and journey tests in parallel on separate builds nodes ([reference](https://docs.travis-ci.com/user/speeding-up-the-build/), I think this is available to Travis OSS repos?)

Smarticles101:
I'm looking into the things you mentioned @stkent :)

stkent:
I had a play with the journey tests this evening, attempting to run them in parallel. However, I think I hit a wall because they don't share a common root Gradle project. Here's the patch I tried to apply:

```diff
diff --git a/bin/journey-test.sh b/bin/journey-test.sh
index b4223f7..6b34670 100755
--- a/bin/journey-test.sh
+++ b/bin/journey-test.sh
@@ -224,35 +224,36 @@ solve_all_exercises() {
   local track_root=$( pwd )
   local exercism_cli="./exercism --config ${exercism_configfile}"
   local exercises=`cat config.json | jq '.exercises[].slug + " "' --join-output`
+  local exercises_array=( $exercises )
   local total_exercises=`cat config.json | jq '.exercises | length'`
[...]

PR #688, opened 2017-07-07
Optimize CI Tests

Smarticles101:
Reference #410 

This pull request attempts to optimize the speed of the travis build's some more.  Unit tests and journey tests are run on separate build nodes, this trims a few minutes.  In the journey test it now downloads everything in parallel to save some time. And then `gradle` builds now use `--configure-on-demand` and `bundle install` uses the `-j4` tag.

After this get's reviewed and it all looks good, I'll make a pull request to mirror the changes in the kotlin track.

---

Reviewer Resources:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 language: java
 jdk:
   - oraclejdk8
+env:
+  - SCRIPT=bin/unit-tests.sh
+  - SCRIPT=bin/journey-test.sh
 
 # http://docs.travis-ci.com/user/migrating-from-legacy
 sudo: false
@@ -23,8 +26,7 @@ before_script:
   - export PATH=$TRAVIS_BUILD_DIR/bin:$PATH  # ensure our tools are prefered over included ones.
 
 script:
-  - bin/unit-tests.sh
-  - bin/journey-test.sh
+  - $SCRIPT
 
 # configure caching (https://docs.travis-ci.com/user/languages/java/#Caching)
 before_cache:
--- a/bin/journey-test.sh
+++ b/bin/journey-test.sh
@@ -164,7 +164,7 @@ make_local_trackler() {
   cp -r ${track_root}/exercises tracks/${TRACK}
 
   gem install bundler
-  bundle install
+  bundle install -j4
   gem build trackler.gemspec
 
   # Make *this* the gem that x-api uses when we build x-api.
@@ -181,7 +181,7 @@ start_x_api() {
   pushd $xapi_home
 
   gem install bundler
-  bundle install
+  bundle install -j4
   RACK_ENV=development rackup &
   xapi_pid=$!
   sleep 5
@@ -233,7 +233,7 @@ solve_all_exercises() {
 
     pushd ${exercism_exercises_dir}/${TRACK}/${exercise}
     # Check that tests compile before we strip @Ignore annotations
-    "$EXECPATH"/gradlew compileTestJava
+    "$EXECPATH"/gradlew compileTestJava --configure-on-demand --parallel
     # Ensure we run all the tests (as delivered, all but the first is @Ignore'd)
     for testfile in `find . -name "*Test.${TRACK_SRC_EXT}"`; do
       # Strip @Ignore annotations to ensure we run the tests (as delivered, all but the first is @Ignore'd).
@@ -241,7 +241,7 @@ solve_all_exercises() {
       # The stripping implementations here and in copyTestsFilteringIgnores should be kept consistent.
       sed 's/@Ignore\(\(.*\)\)\{0,1\}//' ${testfile} > "${tempfile}" && mv "${tempfile}" "${testfile}"
     done
-    "$EXECPATH"/gradlew test
+    "$EXECPATH"/gradlew test --configure-on-demand --parallel
     popd
 
     current_exercise_number=$((current_exercise_number + 1))
--- a/bin/journey-test.sh
+++ b/bin/journey-test.sh
@@ -241,7 +241,7 @@ solve_all_exercises() {
       # The stripping implementations here and in copyTestsFilteringIgnores should be kept consistent.
       sed 's/@Ignore\(\(.*\)\)\{0,1\}//' ${testfile} > "${tempfile}" && mv "${tempfile}" "${testfile}"
     done
-    "$EXECPATH"/gradlew test --configure-on-demand --parallel
+    "$EXECPATH"/gradlew test --configure-on-demand --parallel &
     popd
 
     current_exercise_number=$((current_exercise_number + 1))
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -2,6 +2,9 @@
 language: java
 jdk:
   - oraclejdk8
+env:
+  - SCRIPT=bin/unit-tests.sh
+  - SCRIPT=bin/journey-test.sh
```
stkent: Cool! This seems to be working great! https://travis-ci.org/exercism/java/builds/251297232

bin/journey-test.sh
```diff
@@ -181,7 +181,7 @@ start_x_api() {
   pushd $xapi_home
 
   gem install bundler
-  bundle install
+  bundle install -j4
```
stkent: How did you decide on a value of 4 here?

Also, let's use `--jobs` if we can; makes it easier to parse later on :)

bin/journey-test.sh
```diff
@@ -233,15 +233,15 @@ solve_all_exercises() {
 
     pushd ${exercism_exercises_dir}/${TRACK}/${exercise}
     # Check that tests compile before we strip @Ignore annotations
-    "$EXECPATH"/gradlew compileTestJava
+    "$EXECPATH"/gradlew compileTestJava --configure-on-demand --parallel
```
stkent: Configuration on demand is incubating when used alongside parallel runs, according to the build output. Can you explain a little about why it might be expected to be robust here?

stkent: As far as I can tell, this step being run in parallel does not imply all the journey tests are run in parallel; see the console output of the journey tests here: https://travis-ci.org/exercism/java/jobs/251297234

bin/journey-test.sh
```diff
@@ -233,15 +233,15 @@ solve_all_exercises() {
...
     # Ensure we run all the tests (as delivered, all but the first is @Ignore'd)
     for testfile in `find . -name "*Test.${TRACK_SRC_EXT}"`; do
       # Strip @Ignore annotations to ensure we run the tests (as delivered, all but the first is @Ignore'd).
       # Note that unit-test.sh also strips @Ignore annotations via the Gradle task copyTestsFilteringIgnores.
       # The stripping implementations here and in copyTestsFilteringIgnores should be kept consistent.
       sed 's/@Ignore\(\(.*\)\)\{0,1\}//' ${testfile} > "${tempfile}" && mv "${tempfile}" "${testfile}"
     done
-    "$EXECPATH"/gradlew test
+    "$EXECPATH"/gradlew test --configure-on-demand --parallel &
```
stkent: Same questions as before r.e. this combination of flags and the fact that the console output still appears serial.

bin/journey-test.sh
```diff
@@ -269,18 +269,24 @@ main() {
...
 
-  # Make a local version of trackler that includes the source from this repo.
-  git_clone "x-api" "${xapi_home}"
-  git_clone "trackler" "${trackler_home}"
-  assert_ruby_installed "${trackler_home}"
+  # Download everything we need in parallel
+  git_clone "x-api" "${xapi_home}" &
+  git_clone "trackler" "${trackler_home}" &
+  download_exercism_cli $(get_operating_system) $(get_cpu_architecture) "${exercism_home}" &
+  wait
```
stkent: I'm not familiar with the `wait` keyword as used here; could you provide a link to a good explanation please? Thanks!

bin/journey-test.sh
```diff
@@ -269,18 +269,24 @@ main() {
...
+
+  # Check and install the ruby stuff we need in parallel
+  assert_ruby_installed "${trackler_home}" &
+  assert_ruby_installed "${xapi_home}" &
+  wait
```
stkent: Does this stuff need to happen after `download_exercism_cli`?

bin/unit-tests.sh
```diff
@@ -10,6 +10,6 @@ bin/configlet .
 pushd exercises
 echo ""
 echo ">>> Running tests..."
-TERM=dumb ../gradlew check compileStarterSourceJava --parallel --continue
+TERM=dumb ../gradlew check compileStarterSourceJava --parallel --configure-on-demand --continue
```
stkent: Same question as before r.e. this combination of flags.

bin/journey-test.sh
Smarticles101: Not necessarily after `download_exercism_cli` but they need to happen after `git_clone "x-api" "${xapi_home}"` and `git_clone "trackler" "${trackler_home}"` so that they can verify the version of ruby that is required to build the tracker and x-api

bin/journey-test.sh
Smarticles101: https://en.wikipedia.org/wiki/Wait_(command)
Basically it *waits* for the commands running in the background to finish before continuing.

bin/journey-test.sh
Smarticles101: It seemed to be a common option, but I guess I should have been more thoughtful about it :stuck_out_tongue:. [This blog](http://archlever.blogspot.com/2013/09/lies-damned-lies-and-truths-backed-by.html) indicates that the most effective amount of jobs is one less than the amount of cores, so I might add a function to the script to get the amount of cores and set the amount of jobs to one less

bin/journey-test.sh
Smarticles101: [This](http://logans.me/profile/profile-2017-07-07-21-49-23.html) is the profile generated of when I run `../gradlew test --profile` while in the exercises directory. Configuring the projects took ~13 seconds and the entire length it ran was 3:31.
[This](http://logans.me/profile/profile-2017-07-07-21-54-09.html) is the profile generated from the same command, but I added `--configure-on-demand`. Configuration took about 6 seconds instead here and the total build time was about 2:55 :)

stkent: Was there any caching in play (looking at the timestamps)? I'm surprised that pretty much every part of the task is faster!

bin/journey-test.sh
stkent: Sounds like a good plan!

bin/journey-test.sh
Smarticles101: I don't think there was, I ran a gradle clean in between them. I'm rerunning them right now

Smarticles101: So rerunning them seemed to not show much difference...
[With configure-on-demand](http://logans.me/profile/profile-2017-07-07-22-19-31.html)
[Without](http://logans.me/profile/profile-2017-07-07-22-24-43.html)

Smarticles101: I re ran them again and it didn't seem to make much of a change. I guess something might have been up with that first build. Should I remove the --configure-on-demand tags?

Smarticles101: The [gradle docs] say this(https://github.com/gradle/gradle/blob/master/design-docs/configure-on-demand.md).

> Gradle evaluates all projects even though it might not be necessary. Imagine a 300-module build and someone just wants to run a single task from a single subproject. It would be nice if Gradle only evaluated the projects that are necessary for the selected task to run.

stkent: My hunch is that since tasks are being run for every subproject, then every subproject has to be configured _eventually_ within each build. That + natural variance could explain why in your second test runs, the configuration phase(s) of the version _with_ configuration on demand take longer in total!

My vote would be to remove this flag everywhere since the balance of extra complexity vs performance improvement seems unfavorable.

bin/journey-test.sh
```diff
@@ -164,7 +164,7 @@ make_local_trackler() {
   cp -r ${track_root}/exercises tracks/${TRACK}
 
   gem install bundler
-  bundle install
+  bundle install --jobs $((`nproc --all`-1))
```
stkent: When `nproc` is not available on a system (e.g. on macOS), produces the following log:

```
bin/journey-test.sh: line 184: nproc: command not found
+ bundle install --jobs -1
```

My guess is that this defaults back to a single job, since the build keeps running, but it might be nice to see if there are alternative ways to make this work on all the major platforms (much as we do in other parts of the test code). If that's not possible, I'd prefer to remove this. (I also wonder what happens if running on a single core; my guess is that `bundle install --jobs 0` would behave the same as `bundle install --jobs -1`, but maybe not.)

stkent: Whatever we do here should also be extracted into a variable so that it can be shared between the two `bundle install` invocations.

stkent:
May be relevant: https://stackoverflow.com/a/1715612/2911458

stkent:
This looks good on Mac now! @FridaTveit when you have a chance, please test these changes on Windows! :)

PR #691, opened 2017-07-09
Support a subcommand on configlet in CI

kytrinyx:
We will be adding subcommands to configlet.

For now, we've released a version of configlet which handles both the
old command:

    configlet path/to/track

as well as the new command:

    configlet lint path/to/track

This will let us update all the travis files to include the
subcommand before we release the version of configlet that requires
the subcommand.

https://github.com/exercism/configlet/pull/23


stkent:
Derp, I thought this was an _issue_, not a pull request; thanks, @kytrinyx, LGTM!

PR #579, opened 2017-05-18
Rotational Cipher : Changes made | Add to track

LeonDsouza:
Made all changes as asked by @FridaTveit in the previous pull request.

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/xjava/blob/master/POLICIES.md#event-checklist)

PR #695, opened 2017-07-10
rotational-cipher addd to track

LeonDsouza:
Made all the changes as requested. Sorry for the delay, just got my first job so was a bit busy. 

---

Reviewer Resources:


stkent:
No worries; congratulations on the job! 🎉

stkent:
[See #579 for previous discussion :)]

config.json
```diff
@@ -19,6 +19,13 @@
       ]
     },
     {
+      "slug": "rotational-cipher",
+      "difficulty": 1,
```
FridaTveit: Since you implemented this exercise you would probably be in a good position to estimate the difficulty if you want to :) Just change the difficulty number and move this entry that it's with the other exercises with the same difficulty. 

But don't worry if you don't want to estimate it, you can just leave it with difficulty 1 and we'll estimate it later :) It would be great if you could move it to the end of this list in that case though. That way it'll be positioned at the end of the track so at least users don't get it too early :) With it's current position this would be the third exercise users get which is probably too early.

exercises/rotational-cipher/src/test/java/RotationalCipherTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class RotationalCipherTest {
+
+    private RotationalCipher rotationalCipher;
+
+    /*@Before
+    public void setUp() {
+        rotationalCipher = new RotationalCipher();
+    }
+    */
```
FridaTveit: The commented out code should be removed :)

exercises/rotational-cipher/src/example/java/RotationalCipher.java
```diff
@@ -0,0 +1,44 @@
...
+        return dataStringBuilder.toString();
+    }
+
+    /**
+     * 
+     * @param characterToReplace
+     * @param alphabetCaseStart
+     * @param alphabetCaseEnd
+     * @param shiftKey
+     * @return ReplacementCharacter
```
FridaTveit: `shiftKey` is no longer a parameter of this method :) It's also not very useful to have `@param` annotations that just repeat the name of the parameter. So I would remove the `@param` and `@return` annotations and just leave the last part of the comment :) It's just a suggestion though, it doesn't matter too much :)

LeonDsouza: Did all of the changes @FridaTveit in my repository!

config.json
LeonDsouza: Moved it to the bottom of the json file. 👍

config.json
```diff
@@ -539,7 +539,14 @@
...
       ]
-    }
+    },
+    {
+      "slug": "rotational-cipher",
+      "difficulty": ,
+      "topics": [
+
+      ]
+    },
```
Smarticles101: The comma on this line should be left out. This is what is causing the build to fail

LeonDsouza:
Yea.. done that. Got the build on Travis to pass! :)

FridaTveit:
Thank you for fixing that @LeonDsouza and well done for getting the build to pass! :) You've now moved it back to the top of the file though. What I meant was that it can remain at the end of the file, but still have difficulty 1. That way the build doesn't complain because it doesn't have a difficulty but it also doesn't get given to users as the third exercise in the track. Does that make sense? Sorry that I didn't make that clear!

FridaTveit:
Since changes to the track are coming up I'll go ahead and merge this and submit a separate PR to move the exercise in config.json. Thank you for all your hard work @LeonDsouza! :)

PR #708, opened 2017-07-16
Replace exercise templates with track-level default

kytrinyx:
You can still override the track-level default template by adding a .meta/readme.go.tmpl for the exercise in question.

See https://github.com/exercism/meta/issues/15


Smarticles101:
Thanks @kytrinyx :)

PR #709, opened 2017-07-17
house: add to track

Smarticles101:
---

Reviewer Resources:


exercises/house/src/example/java/House.java
```diff
@@ -0,0 +1,51 @@
+class House {
+    static final String[] CHARACTERS = {
```
FridaTveit: This could be `private` I think?

exercises/house/src/example/java/House.java
```diff
@@ -0,0 +1,51 @@
...
+            "cow with the crumpled horn",
+            "maiden all forlorn",
+            "man all tattered and torn",
+            "priest all shaven and shorn",
+            "rooster that crowed in the morn",
+            "farmer sowing his corn",
+            "horse and the hound and the horn"
+    };
+
+    static final String[] ACTIONS = {
```
FridaTveit: This could also be `private`?

exercises/house/src/test/java/HouseTest.java
```diff
@@ -0,0 +1,344 @@
...
+                "that tossed the dog\n" +
+                "that worried the cat\n" +
+                "that killed the rat\n" +
+                "that ate the malt\n" +
+                "that lay in the house that Jack built.";
+
+        int startVerse = 1;
+        int endVerse = 12;
+
+        assertEquals(expected, house.verses(startVerse, endVerse));
```
FridaTveit: Maybe having a method called `sing` or `singWholeSong` or something might be useful? So that the caller doesn't have to know that there are 12 verses?

PR #726, opened 2017-08-02
Use Collections.singletonList in custom-set tests

### LINKED ISSUE #649: custom-set: use Collections.singletonList
In `CustomSetTest` lists are being constructed using [`Arrays.asList`](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-T...-). However, when there's only one element in the list [`Collections.singletonList`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singletonList-T-) should be used instead.

zwaltman:
Resolves #649.

---

Reviewer Resources:

### DIFF
--- a/exercises/custom-set/src/test/java/CustomSetTest.java
+++ b/exercises/custom-set/src/test/java/CustomSetTest.java
@@ -21,7 +21,7 @@ public class CustomSetTest {
     @Test
     public void setsWithElementsAreNotEmpty() {
         final boolean actual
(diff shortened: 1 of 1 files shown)


zwaltman:
Read the Git Basics and the Java Contributing Guide but I'm mostly still trying to get the hang of Git, so let me know if/what I screwed up so I can do it better in the future. Thanks!

Issue #723, opened 2017-07-29
series: simplify example implementation


FridaTveit:
The example implementation for `Series` can be simplified by:
- `Arrays.asList(string.split((""))).stream()` can just be `Arrays.stream(string.split(("")))`
- `.map(digit -> Integer.parseInt(digit))` can be simplified by using a method reference:
`.map(Integer::parseInt)`
- The class and its methods can be made [package private](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) (by removing the `public` access modifier)

m-a-ge:
@FridaTveit, 
PR #727 has been created.

Please add your feedback and suggestions for my future contributions

FridaTveit:
Fixed in PR #727. Thanks @m-a-ge! :)

PR #727, opened 2017-08-02
Series example refactoring

m-a-ge:
This resolves the following issue: [series: simplify example implementation](https://github.com/exercism/java/issues/723)

All tests passed

---

Reviewer Resources:

### DIFF
--- a/exercises/series/src/example/java/Series.java
+++ b/exercises/series/src/example/java/Series.java
@@ -4,12 +4,12 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class Series {
+class Series {
 
     private final int digitsSize;
     private List<Integer> digits;
 
-    public Series(String string) {
+    Series(String string) {
         this.digits =
                 Arrays.stream(string.split(("")))
                         .map(Integer::parseInt)
@@ -17,7 +17,7 @@ public class Series {
         this.digitsSize = this.digits.size();
     }
 
-    public List<List<Integer>> slices(int num) {
+    List<List<Integer>> slices(int num) {
         if (num > this.digitsSize) {
             throw new IllegalArgumentException("Slice size is too big.");
         }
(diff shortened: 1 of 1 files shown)


exercises/series/src/example/java/Series.java
```diff
@@ -4,21 +4,20 @@
...
 
-    public Series(String string) {
-        this.digits = Arrays
-                .asList(string.split(("")))
-                .stream()
-                .map(digit -> Integer.parseInt(digit))
-                .collect(Collectors.toList());
+    Series(String string) {
+        this.digits =
+                Arrays.stream(string.split(("")))
```
FridaTveit: I would put this on the line above so that it's:
```
this.digits = Arrays.stream(string.split(""))
            .map(Integer::parseInt)
            .collect(Collectors.toList());
```
You might know this already but to make sure formatting is correct you can use ctrl + alt + L if you're using intellij on Windows (it's cmd + alt + L on mac). On eclipse it's ctrl + shift + F (cmd + shift + F on mac). I've found that very useful :)

Also `string.split((""))` can be `string.split("")`. Not your fault at all, I just didn't notice it until now :)

m-a-ge: @FridaTveit 
thanks for your suggestions, I'll update the PR as soon as get back home.

Is there any industry accepted best practices for code formatting (as you mentioned) or it's usually specific to the project (if so, please point me to the page describing this rules)?

m-a-ge:
@FridaTveit 
updated the PR based on your comments in [#discussion_r131078864](https://github.com/exercism/java/pull/727#discussion_r131078864)

exercises/series/src/example/java/Series.java
FridaTveit: The defaults in IntelliJ (and probably also Eclipse) should be fine for formatting :) The [google style guide formatting section](https://google.github.io/styleguide/javaguide.html#s4-formatting) would probably be useful if you want something for reference. I've mostly learnt best practices from experience and code reviews, so I haven't read the google style guide cover to cover, but it looks sensible and I think many people use it :)

m-a-ge: thanks!

PR #731, opened 2017-08-06
Add tests to catch invalid input edge cases

McEileen:
I added version number 2.0.0 to track the changes in tests.  Let me know if there are any questions!

---

Reviewer Resources:

### DIFF
--- a/exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java
+++ b/exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java
@@ -2,8 +2,17 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.Rule;
 
 public class RnaTranscriptionTest {
+
+    /*
+    version: 2.0.0
+     */
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
 
     private RnaTranscription rnaTranscription;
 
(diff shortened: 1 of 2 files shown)


McEileen:
Agh, the travis build failed.  I will look into why it failed and make fixes.

stkent:
It looks like missing semicolons is causing compilation to fail :)

exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java
```diff
@@ -5,6 +5,13 @@
 
 public class RnaTranscriptionTest {
 
+    /*
+    version: 2.0.0
+     */
+
+    @Rule
```
Smarticles101: Rule wasn't imported :)

exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java
```diff
@@ -5,6 +5,13 @@
...
+    public ExpectedException thrown = ExpectedException.none();
```
Smarticles101: ExpectedException wasn't imported either :)

McEileen:
Thanks for the pointers!  Intellij wasn't doing red underwave marks for my compilation errors, like the missing semicolons, and it also didn't highlight `ExpectedException` or `Rule` red when I forgot to import them.  I don't think this is an Intellij configuration issue, because Intellij highlights similar errors in other projects on my machine.  

I'm wondering if something about how I forked and pulled down the repo is causing the errors to not show up.  I forked and pulled down the repo a second time, but still no luck.
Do either of you know of any issues that could be causing this problem?  (I am also googling furiously as we speak....)  Thanks.

stkent:
When developing I usually edit the top-level `build.gradle` file to point to `src/test/java` rather than `build/gen/test/java`; that might help!

PR #732, opened 2017-08-07
Choose core exercises

### LINKED ISSUE #707: Choose core exercises
We need to choose at most 20 exercises to be core exercises (set core: true, and delete the unlocked_by key) since PR #704 has been merged.
For each exercise that is not core we need to decide which exercise is the prerequisite (max 1).
We should also leave 3 or 4 simple exercises as (core: false, unlocked_by: null), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

FridaTveit:
Update config.json to set which exercises are core and which ones are
unlocked by other exercises.

Resolves #707.

---

Reviewer Resources:


FridaTveit:
I need to include `house` in the changes, which has just been merged. Will do that! :)

PR #733, opened 2017-08-07
[WIP] Adds additional test for two-fer exercise

### LINKED ISSUE #728: Commented tests in TwoferTest Class and possibly one case missing
While doing  `two-fer` I've noticed that two positive test cases were commented (not sure why).
```java
    @Ignore("Remove to run test")
    @Test
    public void aNameGiven() {
        String input = "Alice";
        String expected = "One for Alice, one for me.";

        assertEquals(expected, twofer.twofer(input));
    }

    @Ignore("Remove to run test")
    @Test
    public void anotherNameGiven() {
        String input = "Bob";
        String expected = "One for Bob, one for me.";

        assertEquals(expected, twofer.twofer(input));
    }
```

I guess they shouldn't be ignored by default.

Moreover, I suggest adding additional one (but not sure if it complies with task description):
```java
    @Test
    public void emptyStringGiven() {
        String input = "";
        String expected = "One for you, one for me.";

        assertEquals(expected, twofer.twofer(input) );
    }
```

m-a-ge:
This should close #728 

I've added additional test (as discussed) somewhat specific to `java` track and put it in the end of the file  (even though this test doesn't present in [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/two-fer/canonical-data.json) for this exercise)

---

Reviewer Resources:

### DIFF
--- a/exercises/two-fer/src/test/java/TwoferTest.java
+++ b/exercises/two-fer/src/test/java/TwoferTest.java
@@ -38,4 +38,13 @@ public class TwoferTest {
 
         assertEquals(expected, twofer.twofer(input));
     }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void emptyStringGiven() {
+        String input = "";
+        String expected = "One for , one for me.";
+
+        assertEquals(expected, twofer.twofer(input) );
+    }
 }


Smarticles101:
Ah this is failing because the example implementation needs to include the code needed to pass the new test

m-a-ge:
Give me a sec, I'll try to fix it

stkent:
Per https://github.com/exercism/java/issues/728#issuecomment-320671871, wouldn't the expected output here be

> "One for , one for you"

?

Smarticles101:
@stkent I think there was a bit of a confusion. Yes, If we are going by the fact that an empty string is a valid name then the expected output you gave would be correct

m-a-ge:
@stkent, so, you want to see following cases:

1. `null` --> `"One for you, one for me."`
2. `""` --> `"One for , one for me."` (it looks strange to me)
3. `"Bob"` --> `"One for Bob, one for me."`
4. `"Alice"` --> `"One for Alice, one for me."`

right?

stkent:
@m-a-ge correct, according to Katrina, that would be correct :) I would still place the empty string test last since it's a boundary case.

m-a-ge:
in that case, I have to update the code accordingly and clear the history for the last two commits.
It'll take some time

m-a-ge:
@Smarticles101, thanks!

PR #738, opened 2017-08-13
Fixes #737

### LINKED ISSUE #737: minesweeper: use StringBuilder
In `getRowWithNumbers` in `MinesweeperBoard.java` we should use [`StringBuilder`](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html) instead of `+=` to append to a string.

It's better to use `StringBuilder` because it's [mutable](https://www.mkyong.com/java/java-mutable-and-immutable-objects/) and therefore doesn't create a new string every time you add to it. See [here](https://stackoverflow.com/questions/5234147/why-stringbuilder-when-there-is-string) for a more thorough explanation.

matthewstyler:
Use string builder over looped string concatenation.

---

Reviewer Resources:

### DIFF
--- a/exercises/minesweeper/src/example/java/MinesweeperBoard.java
+++ b/exercises/minesweeper/src/example/java/MinesweeperBoard.java
@@ -30,13 +30,13 @@ final class MinesweeperBoard {
     }
 
     private String getRowWithNumbers(final int rowNumber) {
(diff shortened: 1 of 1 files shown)


matthewstyler:
Great, more on the way!

PR #739, opened 2017-08-13
Fixes #696

### LINKED ISSUE #696: series: use Collections.singletonList instead of Arrays.asList
`Arrays.asList` is used to create lists with only one element in test `canSliceByOne` in `SeriesTest.java` . When the list contains only one element it's better to use `Collections.singletonList`.

matthewstyler:
Micro improvement that replaces Arrays.asList with Collections.singletonList

---

Reviewer Resources:

### DIFF
--- a/exercises/series/src/test/java/SeriesTest.java
+++ b/exercises/series/src/test/java/SeriesTest.java
@@ -64,7 +64,7 @@ public class SeriesTest {
     public void canSliceByOne() {
         Series series = new Series("01234");
         List<List<Integer>> expected = Arrays.asList(
-		        Collections.singletonList(0),
+		Collections.singletonList(0),
                 Collections.singletonList(1),
                 Collections.singletonList(2),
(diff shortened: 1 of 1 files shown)


exercises/series/src/test/java/SeriesTest.java
```diff
@@ -63,11 +64,11 @@ public void keepsArbitraryDigitOrder() {
     public void canSliceByOne() {
         Series series = new Series("01234");
         List<List<Integer>> expected = Arrays.asList(
-                Arrays.asList(0),
-                Arrays.asList(1),
-                Arrays.asList(2),
-                Arrays.asList(3),
-                Arrays.asList(4)
+		Collections.singletonList(0),
```
stkent: Spacing looks off by a single char in this line only, otherwise this PR looks great!

matthewstyler: woops, you're right!

PR #740, opened 2017-08-16
Fixes #716 remove redundant type params

### LINKED ISSUE #716: word-count: remove redundant type parameters
In `WordCountTest` a map is initialised using: 
`expectedWordCount = new HashMap<String, Integer>();` 
In this case since the map is already declared as `Map<String, Integer> expectedWordCount` earlier in the file, the type parameters aren't necessary. So this can be simplified to: 
`expectedWordCount = new HashMap<>();`

matthewstyler:
Remove redundant generic type params in map initialization in WordCount and WordCountTest

---

Reviewer Resources:


stkent:
Awesome, thanks again! ❤️

Issue #745, opened 2017-08-18
largest-series-product: fix spelling mistake in exception message


FridaTveit:
In `largest-series-product` the tests expect an exception to be thrown with the message `"String to search may only contains digits."`. This should be `"String to search may only contain digits."`.

PR #748, opened 2017-08-20
fix spelling mistake in exception message #745

riktimmondal:
---

Reviewer Resources:

### DIFF
--- a/exercises/binary-search-tree/src/test/java/BSTTest.java
+++ b/exercises/binary-search-tree/src/test/java/BSTTest.java
@@ -11,7 +11,7 @@ public class BSTTest {
 
     @Test
     public void dataIsRetained() {
-        BST<Integer> bst = new BST();
+        BST<Integer> bst = new BST<>();
         final int actual = 4;
         bst.insert(actual);
         final BST.Node<Integer> root = bst.getRoot();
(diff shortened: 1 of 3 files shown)


stkent:
No idea why that failed; I'll retry the build and 🤞!

stkent:
Ah, whoops, I see why the tests failed now! The change you correctly made was in a test assertion, but the example solution used to validate these tests is still using the _old_ (incorrect) string. If you look in the file

```
exercises/largest-series-product/src/example/java/LargestSeriesProductCalculator.java
```

you should find another copy of the incorrect string `"String to search may only contains digits."`. If you fix that one too, and push up a new or modified commit, then I think we should be good :)

riktimmondal:
Ok

riktimmondal:
I push a new commit correcting the other file then also the build has failed.Please have a look at it

stkent:
@riktimmondal you appear to have pushed an unrelated commit to the branch upon which this pull request is based:

```
82e48fe Added angular bracket when calling constructor
```

This new commit is unrelated and so would ideally not be part of this PR. It's also what is _now_ causing the build to fail; there's a stray `s` character in the change that means the binary search tree tests are now failing to compile. The ideal path forward would be to:

1) Make a new branch based on master
2) Cherry-pick the two commits relevant to issue #745 (cd07493 and d5966bd)
3) Push that new branch to your fork of this repository
4) Make a new pull request based on that newly-pushed branch

I see that you have now closed this pull request, so if you are no longer interested in resolving #745, that's no problem and we can take it from here. Totally your call :)

PR #750, opened 2017-08-21
Fixes #714 assertTrue -> assertEquals

### LINKED ISSUE #714: robot-simulator: use assertEquals instead of assertTrue
In RobotTests, `assertTrue(expected.equals(actual))` is used for all assertions. This should be changed to `assertEquals(expected, actual)` because it is more readable and gives better error messages when the tests fail.

matthewstyler:
---

Reviewer Resources:

### DIFF
--- a/exercises/robot-simulator/src/example/java/GridPosition.java
+++ b/exercises/robot-simulator/src/example/java/GridPosition.java
@@ -9,21 +9,26 @@ final class GridPosition {
         this.y = y;
     }
 
-    /*
-     * This equals method is of deliberately narrow scope (only allows comparison with another GridPosition) to increase
-     * readability. In general, one should provide a full implementation of Object.equals(Object obj) and a
-     * corresponding implementation of Object.hashCode(). See
-     *
-     * https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)
-     *
-     * and
-     *
-     * https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()
-     *
-     * for more information.
-     */
-    boolean equals(final GridPosition gridPosition) {
-        return this.x == gridPosition.x && this.y == gridPosition.y;
-    }
+    @Override
+   public int hashCode() {
+      final int prime = 31;
+      int result = 1;
+      result = prime * result + x;
+      result = prime * result + y;
+      return result;
+   }
+
+   @Override
+   public boolean equals(Object obj) {
+      if (this == obj) {
+         return true;
+      } else if (obj == null || getClass() != obj.getClass()) {
+         return false;
+      } else if (x != ((GridPosition) obj).x || y != ((GridPosition)obj).y) {
+         return false;
+      } else {
+         return true;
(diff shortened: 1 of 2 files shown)


matthewstyler:
tests failing, will investigate

stkent:
I think #714 is invalid; there's a deliberate use of a custom `equals` method there. From the `GridPosition` starter class:

```java
class GridPosition {

    final int x;

    final int y;

    GridPosition(final int x, final int y) {
        this.x = x;
        this.y = y;
    }

    /*
     * This equals method is of deliberately narrow scope (only allows comparison with another GridPosition) to increase
     * readability. In general, one should provide a full implementation of Object.equals(Object obj) and a
     * corresponding implementation of Object.hashCode(). See
     *
     * https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)
     *
     * and
     *
     * https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()
     *
     * for more information.
     */
    boolean equals(final GridPosition gridPosition) {
        return this.x == gridPosition.x && this.y == gridPosition.y;
    }

}
```

stkent:
We could instead provide a full equals implementation, just need to be wary of the fact that it'll be visible to implementors.

Smarticles101:
So then is the assertEquals method using the generic `equals` method from `java.lang.Object`?

stkent:
Yep, exactly, hence the failures!

matthewstyler:
My vote would be for an overridden equals, the class only exists to report the position

FridaTveit:
Sorry, that's my bad for not checking throroughly enough when creating the issue! I did check that the change worked locally, but only for some of the tests when the grid position didn't actually change. 

I agree, I think overriding equals would be the best plan :) Would you be happy doing that as part of this PR @matthewstyler? You absolutely don't have to, I'm happy to open a separate PR for it :)

matthewstyler:
ya, no problem @FridaTveit

exercises/robot-simulator/src/example/java/GridPosition.java
```diff
@@ -9,21 +9,26 @@
...
-     * and
-     *
-     * https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()
-     *
-     * for more information.
-     */
-    boolean equals(final GridPosition gridPosition) {
-        return this.x == gridPosition.x && this.y == gridPosition.y;
-    }
+    @Override
```
FridaTveit: The spacing seems to have gone a bit wrong here and in the test file. It should be indented with four spaces :) If you're using IntelliJ you can use `ctrl + alt + L` to automatically format the code (`cmd + alt + L` on a mac). That should fix spacing issues in the current file and will even remove unused imports if you tell it to in the reformat code dialog box (see https://www.jetbrains.com/help/idea/reformatting-source-code.html). On eclipse it's `ctrl + shift + F` :)

Smarticles101:
Thanks for taking the time to do all this @matthewstyler I think it looks good now

PR #751, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

See https://github.com/exercism/meta/issues/89


stkent:
LGTM

Issue #326, opened 2017-02-24
Fail fast if configlet detects that exercise is missing from config.json?


stkent:
Right now, the build continues to execute for 12-13 minutes. Would it be possible to fail faster in this case of forgetfulness? 👼

jtigger:
Agreed!

In fact, I think there's a larger problem of where the _intent_ is to rely on the `set -e` to fail the build, but that's not always working (e.g. when an error occurs running a subshell that assigns its output to a variable in the main script).

This _should_ be happening in `unit-test.sh` (i.e. `set -e` is in the preamble), but somehow this doesn't fail the build... 💣

jtigger:
This relates to #410.  Gonna note this point in that issue and consider this question closed: yes we want fast-failing builds!

Smarticles101:
If anyone has any specific examples of set -e not failing the build when it should have, could you post them here? I'm going to try fixing this and play around with it a bit @exercism/java and @jtigger if you have the time

stkent:
https://travis-ci.org/exercism/java/builds/205128042; unit tests failed but journey tests ran anyway. This was before we split them out.

PR #752, opened 2017-08-24
Potential solve for #326 and journey-test passing when they fail

Smarticles101:
Not been tested much, but I believe that this should work

---

Reviewer Resources:

### DIFF
--- a/bin/run-journey-test-from-ci.sh
+++ b/bin/run-journey-test-from-ci.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+trap 'exit 1' ERR
 bin/build-jq.sh
 
 pr_files_json=`curl -s https://api.github.com/repos/exercism/java/pulls/${TRAVIS_PULL_REQUEST}/files`
--- a/bin/unit-tests.sh
+++ b/bin/unit-tests.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-set -e
+trap 'exit 1' ERR
+
 ./gradlew --version
 
 echo ""


stkent:
Could you push a broken commit to test this, then revert once we see a failure as expected? That'd be an awesome test in real conditions if so! ❤️

Smarticles101:
Tried setting the `.travis.yml` to set the build status to failed if one of the test suites fail, but it seems that the status stays at pending

Smarticles101:
So I wrestled with getting the build to fail as soon as the first build fails. I found out this isn't possible after emailing support. 

> Thanks for writing in with your feature request. As things are right now, it is not possible to fail a build on the first job failure. It would, however, make sense, and we have reported it internally. This is no guarantee that this will be built.

I'll an issue about it on the travis-ci github about it.

stkent:
Thanks for plugging away at this!

If I understand correctly, you're saying it's not currently possible, when using the Travis matrix stuff, to 'fail fast' as soon as one of the dimensions (e.g. unit tests) fails? Given that, what does the `trap 'exit 1' ERR` change buy us? Not asking because I don't believe it helps; I just don't know how and would appreciate the education :)

Smarticles101:
> If I understand correctly, you're saying it's not currently possible, when using the Travis matrix stuff, to 'fail fast' as soon as one of the dimensions (e.g. unit tests) fails?

That is exactly what I'm saying.

> Given that, what does the trap 'exit 1' ERR change buy us? Not asking because I don't believe it helps; I just don't know how and would appreciate the education :)

It was added to the unit-tests because that allows for errors in subshells to be caught where as `set -e` doesn't. I added it to the run-journey-test-in-ci file because before I had forgotten to add any error handling. Bash defaults to ignore errors inside scripts, so this extra line will catch the error and exit with the exit code 1. Before, if the journey-tests failed, the run-journey-test-in-ci would exit with code 0 because bash was ignoring the error returned by the journey-tests

stkent:
Awesome, thanks! So in the case of the journey tests, those now fail fast in the same way as the unit tests should? 💯 Anything preventing me from merging this as-is? :)

Smarticles101:
Nope! I believe this is good right now. I'll continue to watch the builds over the next week or two

stkent:
Sweet! :shipit:

PR #757, opened 2017-08-26
Add echo's to run-journey-test-from-ci for debug info

Smarticles101:
So I noticed that when trying to get files modified when running the journey-tests, the builds are not able to get the required information, so it continues and runs the entire journey-test suite. This pr is to figure out where the problem is, and fix it

---

Reviewer Resources:


Smarticles101:
Ran a test against [this patch](https://patch-diff.githubusercontent.com/raw/exercism/java/pull/756.patch) that had the problem I described, but I didn't seem to have an issue with it. I'll push the current commit that I have to get some info in the case this issue does happen some more

Issue #406, opened 2017-03-30
Clarification of the sentence on implementation dependencies in the Java "about" page


kotp:
@travisfw commented on [Mon Mar 06 2017](https://github.com/exercism/exercism.io/issues/3396)

The [about page for Java](http://exercism.io/languages/java/about ) says "It has been specifically designed to have as few implementation dependencies as possible" and links to [a page about "implementation dependencies" with regards to Netbeans](https://www.java-tips.org/other-api-tips-100035/60-netbeans/1025-what-is-an-implementation-dependency-and-whathowwhen-should-i-use-one.html ). I have been programming with Java since 2003 and for a year of that used Netbeans, and I have never encountered the feature described on java-tips.org. Java 9 will introduce a *module* concept which seems similar design-wise to the *OpenIDE* modules described there, but I'm not sure what the relationship between the two is (I suspect this Netbeans feature influenced the design of Java 9 modules, and may have even served to prototype that design, but I'm speculating). In any case, it's an inappropriate link for an introduction to what Java is about.

Regardless of the confusing link, I don't know what the sentence about implementation dependencies is trying to communicate. Can it be clarified?

---

@kotp commented on [Mon Mar 06 2017](https://github.com/exercism/exercism.io/issues/3396#issuecomment-284312429)

Thank you @travisfw for opening this issue.  Do you have the time to make an improvement of this paragraph?  Either removal, or clarification if that can be determined?

---

@tejasbubane commented on [Mon Mar 06 2017](https://github.com/exercism/exercism.io/issues/3396#issuecomment-284312879)

@travisfw [This](https://github.com/exercism/xjava/blob/master/docs/ABOUT.md) is the file you want to change.

---

@kytrinyx commented on [Tue Mar 07 2017](https://github.com/exercism/exercism.io/issues/3396#issuecomment-284929382)

/cc @exercism/java 

---

@kytrinyx commented on [Thu Mar 30 2017](https://github.com/exercism/exercism.io/issues/3396#issuecomment-290481039)

This issue should be in the https://github.com/exercism/xjava repository to get tracked by the right folks.

jtigger:
Thanks @travisfw for bringing this up.  Thanks, @kotp for moving this in our space.

This is an excellent additional point to #347.  I'll add the reference.

FridaTveit:
I have the following suggestions:
- We remove the link but keep the sentence as it is.
[...]

PR #761, opened 2017-08-29
Add blurb to config.json

### LINKED ISSUE #759: Add blurb to config.json
In `config.json` we need to have a blurb which explains a bit about Java in ~30 words. See [the Go track's `config.json`](https://github.com/exercism/go/blob/master/config.json) for an example.

FridaTveit:
This is probably not the best description but it's better than nothing. Please feel free to offer suggestions for a better blurb/better grammar/better wording or anything! :)

Fixes #759.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "language": "Java",
   "active": true,
-  "blurb": "Java is a very widely used Object Oriented programming language. Having few low level features makes it safer and easier to use than languages such as C while still being very powerful.",
+  "blurb": "Java is a very widely used Object Oriented programming language. It's safe, simple to use and portable so that you can \"write once, run anywhere\".",
   "exercises": [
     {
       "uuid": "f77dc7e3-35a8-4300-a7c8-2c1765e9644d",
--- a/config.json
+++ b/config.json
@@ -1,6 +1,7 @@
 {
   "language": "Java",
   "active": true,
+  "blurb": "Java is a very widely used Object Oriented programming language. Having few low level features makes it safer and easier to use than languages such as C while still being very powerful.",
   "exercises": [
     {
       "uuid": "f77dc7e3-35a8-4300-a7c8-2c1765e9644d",


michaelspets:
Hey @FridaTveit,  I do not agree with this description. I would prefer instead what James Gosling would say: The Java programming language is a general-purpose concurrent class-based object-oriented programming language, specifically designed to have as few implementation dependencies as possible. It lets application developers "write once, run anywhere".

FridaTveit:
Hi @michaelspets :) Like I said, I'm very happy to receive different suggestions for the blurb, thanks for offering your help! :)

As it is though, we've already had a complaint about the sentence "specifically designed to have as few implementation dependencies as possible." (see issue #406). Also bear in mind that the first thing the users will read after this blurb is [ABOUT.md](https://github.com/exercism/java/blob/master/docs/ABOUT.md) which starts with:
>Java is a general-purpose computer programming language that is:
>- Concurrent
>- Class-Based
>- Object-Oriented

But I guess whatever we write it will probably be repeated somewhere in `ABOUT.md`. Later in `ABOUT.md` it mentions low-level features and being simple as well. So repetition might be unavoidable/not a problem?

Also from what I can tell this is taken word for word from Wikipedia. I think @kytrinyx wanted us not to directly quote something like Wikipedia but instead use our own phrasing to make it more understandable, but I may have got that wrong?

What do you think @exercism/java?

stkent:
I think the implementation dependencies line is confusing, especially for newcomers. The "write once, run anywhere" portability is probably something worth calling out, however. I'd be in favor of adapting that and including it in the blurb somehow.

FridaTveit:
I tried to incorporate the portability in the blurb. What do you think @stkent, @michaelspets? :)

michaelspets:
It is very nice @FridaTveit.

stkent:
Perfect mashup!

Smarticles101:
LGTM!

kytrinyx:
>  I think @kytrinyx wanted us not to directly quote something like Wikipedia but instead use our own phrasing to make it more understandable

Sorry I'm so late to this.

Yes—I don't want us to quote Wikipedia or other sources directly, unless it's a really great colloquial, friendly quote, and then only if we have it in the ABOUT with proper attribution. The blurb is not the place for this.

Thanks @FridaTveit for taking the lead on getting this done, and thank you @michaelspets for your input. It's only by actually raising the issue that we can discuss these things and figure out what works for our situation.

PR #763, opened 2017-09-02
Double indent code blocks to resolve list numbering

Stargator:
[Markdown syntax](https://daringfireball.net/projects/markdown/syntax#list) states code blocks should be double indented in order for the numbering in the lists to continue.

Seeing as how the Java Installation and Tests page had numbered lists that were broken. I opted to fix them. [The issue is discussed here for v2 of exercism](https://github.com/exercism/v2-feedback/issues/105). I am checking the v1 pages to see what are affected by this.

As you can see the current rendering of the Installation page has the numbering thrown off. This should fix that.

java-numbering

---

Reviewer Resources:

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -24,23 +24,24 @@ Open an administrative command prompt.  (If you need assistance opening an admin
 
 1. If you have not installed Chocolatey, do so now:
 
-   ```batchfile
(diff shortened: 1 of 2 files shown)


stkent:
Thanks for the PR!

The double indent is normally required if you are _relying_ on the second indent to trigger code-like formatting; however, these examples are already wrapped in GFM-style code fences, so I'm surprised this change is required.

However, as long as it looks good when rendered, then 🤷‍♂️ ⛴

Stargator:
@stkent Yeah, it's really because of how red-carpet does it. While GitHub flavored is more lenient.

PR #767, opened 2017-09-03
Fixes #364 construct bob in setUp

### LINKED ISSUE #364: bob: extract common setup
In the `bob` exercise test suite, a single instance of `Bob` is created for all test methods.  This presumes there is no state; there is no need to make that assumption.

Please extract the construction of `Bob` to a [`setUp()`](http://junit.sourceforge.net/javadoc/org/junit/Before.html) method.

matthewstyler:
---

Reviewer Resources:


matthewstyler:
looks like the build timed out

FridaTveit:
I've restarted it :)

PR #770, opened 2017-09-05
Restructure exercise directories

stkent:
Moves reference solutions to the new, automatically-ignored `.meta` directory.

---

Reviewer Resources:


stkent:
Keeping an eye on https://github.com/exercism/discussions/issues/192 before I go any further with this.

stkent:
Per https://github.com/exercism/discussions/issues/192#issuecomment-327353555:

> You could certainly just define a reference solution pattern instead of moving it to `.meta`. I have a somewhat more than slight preference for putting everything that doesn't get delivered to the user into `.meta`, since that is pretty easy to understand/explain.

stkent:
This is now ready for review :)

stkent:
Good catches; I'll fix both in this PR!

PR #773, opened 2017-09-05
leap: add to track

Balance-Breaker:


Smarticles101:
> An implementation for 'leap' was found, but config.json specifies that it should be foregone (not implemented).

Sorry about that, but leap isn't supposed to be implemented on the java track. See #3

Smarticles101:
Please don't let this dissuade you. You can implement any exercises listed [here](http://exercism.io/languages/java/todo) :)

PR #774, opened 2017-09-13
Add topics for hello-world

FridaTveit:
[List of topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt)

The only other topic which might be relevant is `classes` since you do have to modify a class to do the exercise. But that's true for all exercises. So we could add `classes` as a topic to all exercises? Or none of them? Or just the ones without starter implementation?

See issue #523 for general discussion about adding topics.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,7 +6,9 @@
       "core": true,
       "difficulty": 1,
       "slug": "hello-world",
-      "topics": null,
+      "topics": [
+        "strings"
+      ],
       "unlocked_by": null,
       "uuid": "f77dc7e3-35a8-4300-a7c8-2c1765e9644d"
     },


FridaTveit:
I'm not sure why the build failed. The error seems to be "Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain". I tried restarting it but it just failed again. Any ideas @stkent @Smarticles101 ?

stkent:
Huh, weird; I bumped Gradle a few days ago (https://github.com/exercism/java/commit/a54c18e52ed0e37b2e808249c61c1329d39dd517) but the master build is fine as far as I can tell. Can you pass this locally?

FridaTveit:
As in, can I run journey-test locally? If that's what you mean, then no it fails with the same error... And it also fails locally on master... :(

stkent:
Ruh roh... looking into this now!

stkent:
Hmm, journey-test works for me locally on master. Still digging...

stkent:
@FridaTveit please rebase on master and re-push; I think this should now be fixed!

FridaTveit:
Great, will do! :) Do you know what was going wrong?

stkent:
I believe the issue was the executable mode set on the Gradle wrapper JAR. Here's the change I made:

https://github.com/exercism/java/commit/89bfa107f4863fe4edc08988fcc350611187068e

Note 644 -> 755.

FridaTveit:
Oh, okay. Thanks! :)

stkent:
Huh, weird. Back to the drawing board!

stkent:
@FridaTveit did the change in master allow you to successfully run it locally?

FridaTveit:
Will try it now :)

FridaTveit:
No, I still get the same error :(

stkent:
Spooky... pushed another commit to master, please retry again. I don't *think* this will fix things, but it might. (Same permissions fix on gradlew.bat).

FridaTveit:
Sure, will try again :)

stkent:
Not sure why https://travis-ci.org/exercism/java/builds doesn't show more recent builds than 2 months ago either...

stkent:
Sorry for all this back and forth; very confusing that everything works on my machine! :S

FridaTveit:
No problem, thanks for helping me look into it! :) But yes, it is all very strange...

stkent:
If you roll back to Gradle 4.0.1 locally, does that fix master?

```
./gradlew wrapper --gradle-version=4.0.1
```

FridaTveit:
No I get the same error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain...

stkent:
What about if you check out 54cf4a6c28a86303faa9c8c6f9b131dc945f8046 (before the Gradle 4.1 update)?

If that doesn't work, I suggest finding a commit in the history that _does_ work, and using `git bisect` to narrow down where the issue was introduced 🤷‍♂️

FridaTveit:
Okay, will try that :)

FridaTveit:
That worked! Checking out [`54cf4a6`](https://github.com/exercism/java/commit/54cf4a6c28a86303faa9c8c6f9b131dc945f8046) worked!

stkent:
Ok, we have a start point! Now how about the next commit, a54c18e52ed0e37b2e808249c61c1329d39dd517, the one that actually bumped Gradle. Does that work locally?

FridaTveit:
Will try now! :)

FridaTveit:
No, that commit doesn't work :(

stkent:
So weird! Can you try locally `git revert`ing that commit on master and seeing if the result runs for you?

FridaTveit:
Sure, will give it a go :)

FridaTveit:
That works :)

stkent:
Ok, then let's just push that up to master, skip Gradle 4.1, and try when 4.2 rolls around? :D

FridaTveit:
Sounds sensible :P

stkent:
Once this is rebased on master I believe it should pass!

FridaTveit:
Great, will rebase! :)

PR #796, opened 2017-09-28
Add topics to rna-transcription

### LINKED ISSUE #782: rna-transcription: add topics

benwicks:
Fix #782 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -27,7 +27,12 @@
       "core": false,
       "difficulty": 2,
       "slug": "rna-transcription",
-      "topics": null,
+      "topics": [
+        "strings",
+        "maps",
+        "algorithms",
+        "loops"
+      ]
       "unlocked_by": null,
       "uuid": "8e983ed2-62f7-439a-a144-fb8fdbdf4d30"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -27,7 +27,12 @@
       "core": false,
       "difficulty": 2,
       "slug": "rna-transcription",
-      "topics": null,
+      "topics": [
+        "strings",
+        "maps",
+        "algorithms",
```
FridaTveit: `strings` and `loops` definitely seem like topics covered by this exercise, great suggestions! :)

I'm not sure about `maps` and `algorithms` though. I think `maps`  refers to the data type,
 e.g. https://docs.oracle.com/javase/8/docs/api/java/util/Map.html which you would be unlikely to use in this exercise? You could definitely argue that you're using an algorithm to solve this exercise, but that argument could probably be made for every exercise. And I don't think it's worth adding `algorithms` as a topic to every exercise, maybe just ones that require you to use some specific algorithm for example `binary-search`. But feel free to disagree! What do you think? :)

benwicks:
Thanks for your patience and help!

FridaTveit:
No problem! We always love having new people contribute! :)

PR #797, opened 2017-09-29
Add topics to raindrops exercise

### LINKED ISSUE #786: raindrops: add topics

Kaioru:
Resolves #786 

Happy Hacktoberfest! 🎉 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -94,7 +94,6 @@
       "topics": [
         "strings",
         "integers",
-        "loops",
         "conditionals",
         "mathematics"
       ],
--- a/config.json
+++ b/config.json
@@ -91,7 +91,13 @@
       "core": false,
       "difficulty": 3,
       "slug": "raindrops",
-      "topics": null,
+      "topics": [
+        "strings",
+        "integers",
+        "loops",
+        "conditionals",
+        "mathematics"
+      ],
       "unlocked_by": "hello-world",
       "uuid": "d916e4f8-fda1-41c0-ab24-79e8fc3f1272"
     },


config.json
```diff
@@ -91,7 +91,13 @@
       "core": false,
       "difficulty": 3,
       "slug": "raindrops",
-      "topics": null,
+      "topics": [
+        "strings",
+        "integers",
+        "loops",
+        "conditionals",
+        "mathematics"
```
FridaTveit: These look like great topic suggestions! `strings`, `integers`, `conditionals` and `mathematics` all definitely seem like topics associated with this exercise :)

I'm not sure about `loops` though. Since you're just checking three different factors using a loop might be overdoing it? So adding `loops` to the suggested topics might be encouraging people to use a non-ideal solution? What do you think? :)

Kaioru: You're right. But the reason I included `loops` is because they might be used when getting the factors of the number. Other than that there wouldn't be any other ideal solutions that makes use of loops.

But maybe the `loops` tag can be covered by the `mathematics` tag instead.. In that case I can remove the `loops` tag.

FridaTveit: You're right, you could use loops to find the factors of a number. But for this problem you don't need to find all the factors of a number, you just need to check whether 3, 5, and 7 are factors. So I think adding `loops` as a topic might be a bit confusing to the people solving the exercise. What do you think? :)

Kaioru: Yeah you're right. I'll change the tags accordingly.

PR #798, opened 2017-09-29
Add Isogram topics

RollandMichael7:
Referencing #788

conditionals: finding if a char is repeated
loops: looping through a string's characters
parsing: parsing a string for repeated characters
strings: self-explanatory :)

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -231,7 +231,6 @@
         "conditionals",
         "loops",
         "strings",
-        "algorithms",
         "parsing"
       ],
       "unlocked_by": "hello-world",
--- a/config.json
+++ b/config.json
@@ -227,7 +227,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "isogram",
-      "topics": null,
+      "topics": [
+        "conditionals",
+        "loops",
+        "strings",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -227,7 +227,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "isogram",
-      "topics": null,
+      "topics": [
+        "conditionals",
+        "loops",
+        "strings",
+        "algorithms",
+        "parsing"
```
FridaTveit: Great suggestions @RollandMichael7! :) The only one I'm not sure about is `algorithms`. I also commented about this on PR #796. You could argue that you're using an algorithm to solve this exercise, but you could make that argument for every exercise. And I'm not sure if it's worth adding `algorithms` as a topic for every exercise (similarly we could add `classes` as a topic for every exercise but I'm not sure that's worth it either). So I think it's best to leave the `algorithms` topic for exercises which require the usage of some specific algorithm such as `binary search` and `luhn`. What do you think? :)

RollandMichael7: Makes sense to me! :)

PR #801, opened 2017-09-30
issue fixed

### LINKED ISSUE #787: pig-latin: add topics

yashasingh:
Added topics for question Pig-latin. This PR fixes #787 

---

Reviewer Resources:


yashasingh:
@FridaTveit Welcome :) :+1:

PR #814, opened 2017-10-01
Add topics for acronym

pvtuan10:
PR for #785 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -75,7 +75,12 @@
       "core": false,
       "difficulty": 3,
       "slug": "acronym",
-      "topics": null,
+      "topics": [
+        "loops",
+        "strings",
+        "searching",
+        "parsing"
+      ],
       "unlocked_by": null,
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -75,7 +75,12 @@
       "core": false,
       "difficulty": 3,
       "slug": "acronym",
-      "topics": null,
+      "topics": [
+        "loops",
+        "strings",
+        "searching",
```
FridaTveit: `loops`, `strings` and `parsing` all seem like great suggestions :) I'm not sure about `searching` though. How would you say `searching` relates to converting a string to its acronym?

pvtuan10: @FridaTveit I think we need to do a linear search through the string.

FridaTveit: Okay, that makes sense :) Happy to merge it then, thanks for explaining! :)

PR #815, opened 2017-10-01
scrabble-score: add topics

hugmanrique:
Adding a list of topics to the scrabble-score problem statement.

There isn't an issue created for it, but it was listed on https://github.com/exercism/java/issues/523#issuecomment-329258303

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -84,6 +84,7 @@
       "difficulty": 3,
       "slug": "scrabble-score",
       "topics": [
+        "games",
         "loops",
         "pattern_matching",
         "strings",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -83,7 +83,12 @@
       "core": true,
       "difficulty": 3,
       "slug": "scrabble-score",
-      "topics": null,
+      "topics": [
+        "loops",
+        "pattern_matching",
+        "strings",
+        "transforming"
```
FridaTveit: Great suggestions! :) I think I would also add `games` as a topic since this is to do with scrabble. Most of the exercises this exercise unlocks are also to do with games (e.g. `minesweeper` and `queen-attack`) so it would be nice if they all have the `games` tag so people can see how they're related :)

PR #817, opened 2017-10-02
Added topics for luhn

AndresReyMaz:
First pull request ever! Just added five topics to the exercise 'luhn'.

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/xjava/blob/master/POLICIES.md#event-chec
https://github.com/exercism/xjava/blob/master/POLICIES.md#event-checklist)

### DIFF
--- a/config.json
+++ b/config.json
@@ -137,7 +137,7 @@
       "difficulty": 4,
       "slug": "luhn",
       "topics": [
-      	"algorithms",
+        "algorithms",
         "booleans",
         "loops",
         "mathematics",
--- a/config.json
+++ b/config.json
@@ -136,7 +136,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "luhn",
-      "topics": null,
+      "topics": [
+        "booleans",
+        "loops",
+        "mathematics",
+        "strings",
+        "type_conversion"
+      ],
       "unlocked_by": "hamming",
       "uuid": "5227a76c-8ecb-4e5f-b023-6af65a057c41"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -136,7 +136,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "luhn",
-      "topics": null,
+      "topics": [
+        "booleans",
+        "loops",
+        "mathematics",
+        "strings",
+        "type_conversion"
```
FridaTveit: Great suggestions! The only one I'm not sure about is `type_conversion`. How would you say that relates to this exercise? 

I would also maybe add the `algorithms` tag since this exercise is to do with implementing a specific algorithm :)

AndresReyMaz: The `algorithms` tag is a great idea! I will add that one. 

Regarding the `type_conversion`, I added it because it's necessary to convert the String representation of the number (since that is the parameter the function takes in the skeleton) to an integer array (or other integer-based representation) in order to do the processing and validate.

Let me know if this reasoning sounds solid to you. If not, I can simply remove it (:

FridaTveit: Sure, that sounds fair :) Thanks for explaining! :) Happy to merge it once you've added the `algorithms` tag then! :)

AndresReyMaz: Ok! It's done now :)

config.json
```diff
@@ -136,7 +136,14 @@
       "core": false,
       "difficulty": 4,
       "slug": "luhn",
-      "topics": null,
+      "topics": [
+      	"algorithms",
```
FridaTveit: Sorry to be very picky but this should have one more space of indentation I think. Would you mind fixing it? If you don't want to then that's fine, just let me know and I can fix it :)

FridaTveit:

PR #818, opened 2017-10-02
Updated Food-Chain Topics :)

altDaemon:
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -398,7 +398,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "food-chain",
-      "topics": null,
+      "topics": [
+ +      	"arrays",
+ +      	"lists",
+ +      	"strings",
+ +      	"transforming"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -398,7 +398,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "food-chain",
-      "topics": null,
+      "topics": [
+ +      	"arrays",
```
FridaTveit: These `+`s shouldn't be here, that's what's making the build fail :) Apart from that I think you've got a good list of suggested topics! :) The only topic I'm not sure about is `transforming`. How would you say that's related to this exercise?

altDaemon: Oh goodness, I can't believe I missed that. haha

Thank you so much! :)

altDaemon:
Hey there, I believe this is ready now! :)

Issue #808, opened 2017-10-01
nucleotide-count: add topics


FridaTveit:
As part of the update to the exercism website, we want to add topics to each exercise. This should be topics that will most likely be relevant to the exercise even though you might be able to solve it without encountering that topic. There is a [list of topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) used across all exercism tracks but you can add track specific ones as well if needed.

See issue #523 for a discussion about adding topics. There are also some exercises (e.g. hello-world and two-fer) that already have topics so you can look at them as examples. The topics need to be added to the entry for the relevant exercise in `config.json`.

sjwarner-bp:
I think I'll try this one, if it is still up for grabs. I'll try just doing one to familiarise myself with the process and make sure I am doing it right, then afterwards I'll do a few, if that is ok?

sjwarner-bp:
I have just submitted the PR - thank you @FridaTveit, I did submit it to the wrong place originally! :smiley:

FridaTveit:
Fixed in PR #820. Thanks @sjwarner-bp! :)

sjwarner-bp:
Thanks for letting me help out @FridaTveit !

FridaTveit:
No problem, thanks for wanting to help! :)

PR #820, opened 2017-10-02
Update nucleotide-count topics as in #808

sjwarner-bp:
Added four tags to nucleotide-count topics as described in #808.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -300,8 +300,11 @@
       "topics": [
         "conditionals",
         "exception_handling",
+        "integers",
+        "maps",
         "parsing",
-        "searching"
+        "searching",
+        "strings"
       ],
       "unlocked_by": "hamming",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -297,7 +297,12 @@
       "core": false,
       "difficulty": 5,
       "slug": "nucleotide-count",
-      "topics": null,
+      "topics": [
+        "conditionals",
+        "exception_handling",
+        "parsing",
+        "searching"
```
FridaTveit: These look like great suggestions! :) I think `maps` should definitely also be a topic since this is one of the few exercises where you have to use maps. And maybe also `integers`? And `strings`? What do you think? :)

sjwarner-bp: Yes, this all sounds good too! I can update it :smile:

PR #821, opened 2017-10-02
Add topics for largest-series-product

pvtuan10:
PR for #806 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -155,7 +155,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "largest-series-product",
-      "topics": null,
+      "topics": [
+        "loops",
+        "type_conversion",
+        "integers",
+        "strings",
+        "mathematics"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -155,7 +155,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "largest-series-product",
-      "topics": null,
+      "topics": [
+        "loops",
+        "type_conversion",
+        "integers",
+        "strings",
+        "mathematics"
```
FridaTveit: These seem like great suggestions! :) The only one I'm not sure about is `type_conversion`. How would you say that related to this exercise?

FridaTveit: Actually, someone else just explanined their use of this tag on a similar exercise (`luhn`) so I agree now that it's fine for it to be used :)

PR #822, opened 2017-10-02
added diamond topics

spmvanmierlo:
#791

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -219,7 +219,13 @@
       "core": false,
       "difficulty": 4,
       "slug": "diamond",
-      "topics": null,
+      "topics": [
+		"strings",
+		"loops",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -219,7 +219,13 @@
...
       "difficulty": 4,
       "slug": "diamond",
-      "topics": null,
+      "topics": [
+		"strings",
+		"loops",
+		"arrays",
+		"lists",
+		"text_formatting"
+		] ,
```
FridaTveit: Great suggested topics! :) Just a very minor thing, there shouldn't be a space here after the square bracket and before the comma. Would you mind removing it? :)

PR #834, opened 2017-10-03
Update config.json 

### LINKED ISSUE #829: queen-attack: add topics

Akatsuki06:
added topics for the queen-attack problem
fixes #829

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -461,7 +461,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "queen-attack",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "matrices",
+        "conditionals",
+        "mathematics",
+      ],
       "unlocked_by": "scrabble-score",
       "uuid": "5404109e-3ed9-4691-8eaf-af8b36024a44"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -461,7 +461,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "queen-attack",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "matrices",
+        "conditionals",
+        "mathematics"
```
FridaTveit: These look like great suggestions! :) I would also add `games` as a topic since this is to do with chess :) Do you agree?

Akatsuki06: @FridaTveit 
games and also how about **tuples** ? Can we add even though its not java datatype? or I should specify as **classes** as this specific program contains custom class BoardCoordinate .

FridaTveit: I wouldn't add `tuples` since it's not a java datatype so it could be confusing. But `classes` is a very good suggestion! :)

Akatsuki06: I have done the updates

PR #836, opened 2017-10-03
added topics for word search

### LINKED ISSUE #832: word-search: add topics

Akatsuki06:
added topics for word search
fixes #832

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -573,11 +573,11 @@
       "difficulty": 7,
       "slug": "word-search",
       "topics": [
-        "string",
-        "words",
-        "loops",
+        "strings",
+        "games",
+        "logic",
         "matrices",
-        "matching"
+        "pattern_matching"
       ],
       "unlocked_by": "scrabble-score",
       "uuid": "b53bde52-cb5f-4d43-86ec-18aa509d62f9"
--- a/config.json
+++ b/config.json
@@ -577,6 +577,7 @@
         "games",
         "logic",
         "matrices",
+        "searching",
         "pattern_matching"
       ],
       "unlocked_by": "scrabble-score",
--- a/config.json
+++ b/config.json
@@ -572,7 +572,13 @@
       "core": false,
       "difficulty": 7,
       "slug": "word-search",
-      "topics": null,
+      "topics": [
+        "string",
+        "words",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -572,7 +572,13 @@
       "core": false,
       "difficulty": 7,
       "slug": "word-search",
-      "topics": null,
+      "topics": [
+        "string",
```
FridaTveit: This should be `strings` to match the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt). You can add language specific topics that aren't there if you feel the need, but otherwise we should remain consistent with that list as those topics will be used by the other language tracks :)

config.json
```diff
@@ -572,7 +572,13 @@
...
+        "words",
```
FridaTveit: This isn't a topics in the topics list and I'm not sure it makes sense as a topic? `words` doesn't seem like a language concept to me. But feel free to disagree :)

config.json
```diff
@@ -572,7 +572,13 @@
...
+        "loops",
+        "matrices",
+        "matching"
```
FridaTveit: This is also not a topic in the topics list. Maybe `searching` would be better? I would also add `games` to this list since `word-search` is a game :)

Akatsuki06: hmm okay I will make the changes you want. ;)

Akatsuki06: So you mean we have to pick a topic from the TOPICS.txt

FridaTveit: The list of topics (TOPICS.txt) is there to make sure the topics that are shared are consistent. This is to avoid the case where for example one track uses `string` as a topic and a different track uses `strings`. So when adding a topic you should first check the list to see if the topic you want (or something very similar) is already there. If so you should add the version in the list to be consistent. If the topic you want is not there then there are a few options:
- If you think it's a valid topic that could be used in several language tracks then you can open a PR to update the topics list.
- If you think it's a valid topic that is specific to this language track then it's fine to add it without it being in the topics list.

Does that make sense? :)

Akatsuki06: Yep , now it makes sense :p hehe

PR #838, opened 2017-10-03
Added topics for minesweeper

### LINKED ISSUE #830: minesweeper: add topics

Akatsuki06:
fixes #830

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -513,7 +513,9 @@
         "integers",
         "matrices",
         "games",
-        "conditionals"
+        "conditionals",
+        "strings",
+        "lists"
       ],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -509,7 +509,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "minesweeper",
-      "topics": null,
+      "topics": [
+        "integers",
+        "matrices",
+        "games",
+        "conditionals"
```
FridaTveit: Great suggestions! :) Maybe also `strings` and `lists`? :)

Akatsuki06: OKK

Issue #700, opened 2017-07-12
binary-search-tree: add angle brackets when calling constructor


FridaTveit:
Since the `BST` class uses generics it should be constructed using angle brackets. For example:
```
BST<Integer> binarySearchTree = new BST<>();
```
This is done sometimes in `BSTTest` but not always. It should be done everywhere where a `BST` object is created in `BSTTest`.

stkent:
(Reminder that this exercise is pending review for use of generics per https://github.com/exercism/java/issues/230 also.)

ribhav-pahuja:
@FridaTveit , I would like to contribute to this.

Smarticles101:
@ribhav-pahuja Feel free to tackle this! 👍

FridaTveit:
Yes, like @Smarticles101 said, feel free to work on it :)

FridaTveit:
Fixed in PR #842. Thanks @szielinski! :) 

I'm sorry if you were still working on it @ribhav-pahuja. I forgot about your comment when merging the PR, sorry. There should be lots of other `good-first-patch` issues you can work on though. Let me know if you have any problems finding any :)

PR #842, opened 2017-10-03
BST: add angle brackets when calling constructor

szielinski:
- Always use angle brackets in BSTTest when constructing BST objects

---

Reviewer Resources:


szielinski:
#700

Issue #800, opened 2017-09-30
Triangle:  test suite diverges from canonical data


lbesaw:
The test diverges from canonical data here: https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/canonical-data.json

FridaTveit:
In my opinion (and I think this is an opinion other exercism maintainers share?) the tests in different tracks don't have to be exactly the same as the canonical data. As long as they're testing the same things and especially that they don't contradict each other, then I think it's fine for tracks to restructure the tests to more suit the language. For example the Java implementation of this exercise uses an enum to return the classification of the triangle, instead of having separate methods for checking whether a triangle is scalene, equilateral or isosceles which the canonical data might suggest. This would naturally result in slightly  different tests. Do you agree that this is okay @exercism/java? Or should we change the tests to be more like the canonical data?

Was there anything in particular you think should be changed in the tests @lbesaw?

stkent:
IMHO, having our test cases match 1:1 with canonical test cases (at least in spirit) is a good thing  because it ensures thorough coverage. In fact, I'm such a fan of this that I generally avoid implementing any exercise without canonical tests.

Since Java doesn't auto-generate test suites, it is not unexpected that we sometimes fall out of sync with the canonical data. This is generally undesirable (though less so than in the past, now that language tracks "own" their READMEs also) and so we appreciate it being brought to our attention!

In this particular case, it looks like we will need to rethink the Java implementation fairly significantly in light of this new test:

>  equilateral triangles are also isosceles

Given that, we may want to introduce queryable methods (as hinted at by the canonical tests) and dump the classification enum.

lbesaw:
My only real issue was that there is a test involving a degenerate triangle which is meant to fail in testing, however the instructions state that a degenerate triangle should pass. Just some consistency would be appreciated.

FridaTveit:
@lbesaw , yes issue #799 covers that problem, which I absolutely agree is a problem. I don't particularly mind that our tests are slightly different to the canonical data, but I also don't mind reworking our solution if you think that's best @stkent. So having `isIsosceles`, `isEquilateral` and `isScalene` methods instead of the enum? I'm happy to open a PR for that :) I guess if we're then sticking 100% to the canonical data then we just don't test for the degenerate case at all?

stkent:
@lbesaw you're right, the README for this exercise is currently "locked" at a version that doesn't match the test suite. This is a problem that will go away over time now that tracks "own" both pieces of data :)

[...]

PR #843, opened 2017-10-03
Triangle: Replace enum with isEquilateral, is Isosceles, and isScalene methods

jackattack24:
In reference to change 1 in #800.

---

Reviewer Resources:

### DIFF
--- a/exercises/triangle/src/main/java/Triangle.java
+++ b/exercises/triangle/src/main/java/Triangle.java
@@ -4,8 +4,4 @@ class Triangle {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
-    TriangleKind getKind() {
-        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
-    }
-
 }
--- a/exercises/triangle/src/main/java/Triangle.java
+++ b/exercises/triangle/src/main/java/Triangle.java
@@ -4,4 +4,16 @@ class Triangle {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
+    public boolean isEquilateral() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    public boolean isIsosceles() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    public boolean isScalene() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
 }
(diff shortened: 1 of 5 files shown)


jackattack24:
@FridaTveit I wasn't sure if I should include stubs for the new methods in the starter implementation, so I left them out.

exercises/triangle/src/main/java/Triangle.java
```diff
@@ -4,8 +4,4 @@
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
-    TriangleKind getKind() {
-        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
-    }
```
FridaTveit: Please add stubs for the new methods. The policy for starter implementation is in the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations). This is the 17th exercise in the track by difficulty (from the order in `config.json`) so it should have stubs for all required methods :)

exercises/triangle/src/test/java/TriangleTest.java
```diff
@@ -14,79 +14,79 @@
     public void equilateralTrianglesHaveEqualSides() throws TriangleException {
         Triangle triangle = new Triangle(2, 2, 2);
 
-        assertEquals(TriangleKind.EQUILATERAL, triangle.getKind());
+        assertEquals(true, triangle.isEquilateral());
```
FridaTveit: You can use `assertTrue` instead in all these assertions :)

exercises/triangle/src/main/java/Triangle.java
FridaTveit: I did see your comment asking about stubs and it's perfectly fine that you weren't sure of course :) The documentation for things can be a bit hard to find sometimes :P

exercises/triangle/src/main/java/Triangle.java
```diff
@@ -4,7 +4,15 @@
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
-    TriangleKind getKind() {
+    public boolean isEquilateral() {
```
FridaTveit: A very minor point (sorry to be picky) but you don't need `public` here. It's best practice to be as restrictive as possible with access modifiers and you only need this to be `package private` (which, as you might know, is the default so you don't have to specify it). Especially since the constructor and class are both `package private`, it makes sense for these methods to be as well :)

jackattack24:
Should be good now. Thank you for the feedback @FridaTveit :)

PR #857, opened 2017-10-04
meetup: add topics

### LINKED ISSUE #828: meetup: add topics

deepakavs:
Fixes #828

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -606,7 +606,17 @@
       "core": false,
       "difficulty": 7,
       "slug": "meetup",
-      "topics": null,
+      "topics": [
+        "conditionals",
+        "dates",
+        "loops",
+        "maps",
+        "parsing",
+        "pattern_recognition",
+        "strings",
+        "tuples",
+        "type_conversion"
+      ],
       "unlocked_by": "gigasecond",
       "uuid": "602511d5-7e89-4def-b072-4dd311816810"
     },
--- a/config.json
+++ b/config.json
@@ -610,12 +610,7 @@
         "conditionals",
         "dates",
         "loops",
-        "maps",
-        "parsing",
-        "pattern_recognition",
-        "strings",
-        "tuples",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -606,7 +606,17 @@
...
+      "topics": [
+        "conditionals",
+        "dates",
+        "loops",
+        "maps",
+        "parsing",
+        "pattern_recognition",
+        "strings",
+        "tuples",
+        "type_conversion"
```
FridaTveit: These look like good suggestions :) I agree that `conditionals`, `dates` and `loops` are good topics :) I would also add `enumerations` since it uses an enum :)

I'm not sure about `maps`, `parsing`, `strings`, `pattern_recognition`, `type_conversion` or `tupes`. None of those topics seem to be used in the [tests](https://github.com/exercism/java/blob/master/exercises/meetup/src/test/java/MeetupTest.java) or [example solution](https://github.com/exercism/java/blob/master/exercises/meetup/.meta/src/reference/java/Meetup.java). How do you think they're relevant to the exercise?

deepakavs: Well, I was thinking of a different solution. I am not aware that I should follow the example solution and add topics for only the features used in example solution.
In that case, I ll take them out. :)

FridaTveit: You absolutely don't have to only add topics used in the example solution. Sorry if I was being confusing :) We should add topics that are used in the tests and topics we think could be used in a good solution. The only topics we want to avoid are topics that are not at all relevant, and topics that would only be used in a "bad" solution, because we don't want to confuse the people or stear them towards a non-ideal solution. Does that make sense? 

So if you think that those topics could be used in a solution that would pass the tests, then that's fine, we can add them :) It would be great if you could explain vaguely what that solution would be like though? :)

FridaTveit:

PR #858, opened 2017-10-04
Add topics to transpose exercise

### LINKED ISSUE #856: transpose: add topics

delbertlegg:
Fixes #856. Thanks!

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -456,7 +456,14 @@
       "core": false,
       "difficulty": 6,
       "slug": "transpose",
-      "topics": null,
+      "topics": [
+        "loops",
+        "arrays",
+        "lists",
+        "matrices",
+        "strings",
+        "text_formatting"
+      ],
       "unlocked_by": "matrix",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -456,7 +456,15 @@
...
       "slug": "transpose",
-      "topics": null,
+      "topics": [
+        "loops",
+        "recursion",
+        "arrays",
+        "lists",
+        "matrices",
+        "strings",
+        "text_formatting"
```
FridaTveit: These are great suggestions! :) The only one I'm not sure about is `recursion`. I think it should be fairly straight-forward to solve the exercise without recursion and I would hesitate to encourage people to use resursion in Java when it isn't needed since using iteration is more suited to the language. What do you think? :)

delbertlegg: @FridaTveit That's fine with me...I figured it may be an opportunity to practice it, but it's probably square peg in a round hole for this one. I'll update it.

PR #860, opened 2017-10-04
Add topics to pascals-triangle

### LINKED ISSUE #854: pascals-triangle: add topics

puripant:
Fixes #854 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -396,11 +396,12 @@
       "difficulty": 5,
       "slug": "pascals-triangle",
       "topics": [
-	 "recursion",
-	 "arrays",
-	 "matrices",
-	 "algorithms",
-	 "mathematics"
+        "arrays",
+        "matrices",
+        "algorithms",
+        "mathematics",
+        "exception_handling",
+        "integers"
       ],
       "unlocked_by": "matrix",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -395,7 +395,13 @@
       "core": false,
       "difficulty": 5,
       "slug": "pascals-triangle",
-      "topics": null,
+      "topics": [
+	 "recursion",
+	 "arrays",
+	 "matrices",
+	 "algorithms",
+	 "mathematics"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `recursion`. You could quite easily solve this exercise without recursion I think, and I would hesitate to encourage people to use recusion in Java when they don't need to since using iteration is more suited to the language. What do you think? :)

I would also add `exception_handling` and `integers` as topics since the tests expect an exception to be thrown and it uses integers. Do you agree? :)

Also, all of these should be indented two spaces in from `topics`, like the entries for any other exercise with topics :)

puripant: I am fine with your suggestions. I will add `exception_handling` and `integers` as well. I guess it depends on whether you want the topic list to be more on the exhaustive side or not. I was even considering adding `loops` :)

PR #862, opened 2017-10-04
Add topics to Secret Handshake exercise

### LINKED ISSUE #848: secret-handshake: add topics

filipolszewski:
Fix #848 

First time contributing, hope this helps :)

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -135,9 +135,13 @@
       "slug": "secret-handshake",
       "topics": [
         "loops",
-	"transforming",
         "conditionals",
-	"cryptography"
+        "integers",
+        "mathematics",
+        "transforming",
+        "cryptography",
+        "enumerations",
+        "bitwise_operations"
       ],
       "unlocked_by": null,
       "uuid": "71c7c174-7e2c-43c2-bdd6-7515fcb12a91"
--- a/config.json
+++ b/config.json
@@ -133,7 +133,12 @@
       "core": true,
       "difficulty": 3,
       "slug": "secret-handshake",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -133,7 +133,12 @@
       "core": true,
       "difficulty": 3,
       "slug": "secret-handshake",
-      "topics": null,
+      "topics": [
+        "loops",
+	"transforming",
+        "conditionals",
+	"cryptography"
```
FridaTveit: Great suggestions! I would add `enumerations` since it uses an `enum` and maybe also `mathematics` and `bitwise_operations`? What do you think?

Also, the indentation seems to have gone a bit wrong. They should all be indented two spaces in from `topics`, like `loops` and `conditionals` are :)

FridaTveit: Oh and also `integers` would be a good topic :)

filipolszewski: Thanks for all the informations and suggestions. I'm up for adding more topics - I wasn't sure if there's a suggested/maximum number of topics for each exercise, so I kept it minimal.

filipolszewski:
I added all suggestions and fixed that strange indentation. Good point on 'bitwise_operations" topic - must have miseed it the first time I've checked through the topic list.

config.json
FridaTveit: I don't think we have a suggested maximum or minimum number of topics at the moment. Adding topics is still a relatively new things so that might change in the future, but for the moment I think we can just add as many topics as we feel are relevant :)

PR #863, opened 2017-10-04
added topics to perfect numbers

### LINKED ISSUE #845: perfect-numbers: add topics

edgaryu:
This fixes #845.
This is my first PR, and I hope I did this correctly according to the PR guidelines.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -141,7 +141,11 @@
       "core": false,
       "difficulty": 3,
       "slug": "perfect-numbers",
-      "topics": null,
+      "topics": [
+	"integers",
+	"sequences",
+	"mathematics"
+      ],
       "unlocked_by": "difference-of-squares",
       "uuid": "d0dcc898-ec38-4822-9e3c-1a1829c9b2d9"
     },


config.json
```diff
@@ -141,7 +141,11 @@
       "core": false,
       "difficulty": 3,
       "slug": "perfect-numbers",
-      "topics": null,
+      "topics": [
+	"integers",
+	"sequences",
+	"mathematics"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `sequences`. How would you say that relates to this exercise? 

Also, `enumerations` and `exception_handling` would be good topics since the exercise uses an enum and the tests expect an exception to be thrown :)

And a minor thing, these should all be indented two spaces in from `topics`, like the other exercises that have topics (e.g. `hello-world` or `two-fer`) :)

edgaryu: Thanks for the advice. 
About `sequences`, I was thinking about how the exercise filters a sequence of integers, but in fact, `sequences` is supposed to refer to a data type as stated in the topic list. So `sequences` shouldn't be included. Also will fix the indentation.
I will include `enumerations` and `exception_handling`. 

I also want to propose `filtering` since multiples of the `naturalNumber` are being filtered from the `IntStream`, according to the reference file.

FridaTveit: Great :) Sure, you can add `filtering` if you think that would be useful :)

edgaryu:
Sorry I didn't the formatting correct the second time. Will try to fix again.

FridaTveit:
What's wrong with the formatting?

edgaryu:
My text editor's default spacing uses tabs instead of spaces. So in github, the formatting is incorrect.

PR #864, opened 2017-10-04
list-ops: add hint of using static methods per #378

### LINKED ISSUE #378: list-ops: include hints to explain why methods are static
As per #177, we prefer using instance methods over class methods (see discussion in that issue for rationale).

However, there _are_ conditions in which it is absolutely appropriate for a function to be `static`.  A common situation is because the function is effectively an extension of an existing type — you are defining behavior you wished were defined on a given type.

Since classes in Java are closed for modification (i.e. you cannot add members to a class outside its definition like you can in other languages like Ruby or JavaScript), you cannot add the new behavior to the class, directly.  The idiomatic solution in this case is to write a utility method.

Collections of these kinds of methods are often referred to as "utility classes".  Examples of such classes from within the JRE include [`Arrays`](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html) and [`Collections`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html).

In the case of the exercise `list-ops`, we "wish" that [`List`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html) had `map()`, `reduce()`, `filter()`, etc. methods on it.  It doesn't, so we're including them here.

Please include a `HINT.md` file in `list-ops` that points out that these methods are static and capture the description above, in your own words as to why this class has this shaped API.

m-a-ge:
Hopefully resolves #378 

---

Reviewer Resources:

### DIFF
--- a/exercises/list-ops/.meta/hints.md
+++ b/exercises/list-ops/.meta/hints.md
@@ -1,3 +1,7 @@
 ## Hints
 
+In Java it's considered best practice to use instance methods over class methods. However, there are conditions in which it is absolutely appropriate for a function to be `static`. Since classes in Java are closed for modification (i.e. you cannot add members to a class outside its definition like you can in other languages like Ruby or JavaScript), you cannot add new behavior to the class directly. What to do if you still want to define behavior for a given type? The idiomatic solution in this case is to write a utility method. 
+Collections of these kinds of methods are often referred to as "utility classes". Examples of such classes from within the JRE include [Arrays](https://docs.oracle.com/javase/9/docs/api/java/util/Arrays.html) and [Collections](https://docs.oracle.com/javase/9/docs/api/java/util/Collections.html).
+In this exercise we want a List to have `map()`, `reduce()`, `filter()`, etc. methods. It doesn't, so we're using static methods.
+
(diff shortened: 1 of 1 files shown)


exercises/list-ops/.meta/hints.md
```diff
@@ -1,3 +1,7 @@
 ## Hints
 
+In Java it's a best pracite to use instance methods over class methods. However, there are conditions in which it is absolutely appropriate for a function to be `static`. Since classes in Java are closed for modification (i.e. you cannot add members to a class outside its definition like you can in other languages like Ruby or JavaScript), you cannot add new behavior to the class, directly. What to do if you still want to define behavior for a given type? The idiomatic solution in this case is to write a utility method.
```
FridaTveit: Should be "practice" :) And I don't think there should be a comma before "directly"? I'm not an expert on comma usage though so feel free to tell me if I'm wrong :P

FridaTveit: Also I think it should be "it's considered best practice" instead of "it's a best practice". What do you think?

m-a-ge: @FridaTveit you are right, I'll fixes those later today.
Thanks for pointing out 👍

m-a-ge:
@FridaTveit I've corrected those annoying mistakes

PR #865, opened 2017-10-04
wordy: add topics for exercise to config.json per #850

### LINKED ISSUE #850: wordy: add topics

m-a-ge:
Fixes #850 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -579,7 +579,16 @@
       "core": false,
       "difficulty": 6,
       "slug": "wordy",
-      "topics": null,
+      "topics": [
+        "strings",
+        "integers",
+        "logic",
+        "parsing",
+        "pattern_matching",
+        "regular_expressions",
+        "transforming",
+        "exception_handling"
+      ],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -579,7 +579,12 @@
       "core": false,
       "difficulty": 6,
       "slug": "wordy",
-      "topics": null,
+      "topics": [
+        "strings",
+        "parsing",
+        "pattern_matching",
+        "transforming"
```
FridaTveit: Great suggestions! :) I would maybe add `logic` and `integers` as well. And maybe `regular_expressions`? What do you think?

FridaTveit: And also `exception_handling` since the tests expect an exception to be thrown :)

m-a-ge: Good points, I'll update the PR shortly

m-a-ge:
@FridaTveit I've updated PR based on your suggestions

Issue #230, opened 2016-12-29
binary-search;binary-search-tree;custom-set;linked-list;simple-linked-list: use generics?


jtigger:
`linked-list`'s tests only insert and expect integers.  However, the tests force the practitioner to generify the class.

We should:
1. remove type variables from the class signature in the tests and either expect `Number` or `Integer`
2. or expand the test suite to put all kinds of objects in the collection.

What do you think?

matthewmorgan:
@jtigger I think the use of generics is a good idea.  We don't use them much in this track, and it's a key feature of the language.  I think we should go 'all in' for generics here.

stkent:
@jtigger neither of these options seem to be consistent with [our recent discussion](https://github.com/exercism/xjava/pull/186#discussion_r89926622) on `binary-search`?

@matthewmorgan good point, but how should be handle existing canonical test data? Add tests above and beyond the canonical tests? PR into the canonical test data with new tests that use different data types for all tracks, or for statically typed tracks only?

jtigger:
@stkent said:
> @jtigger neither of these options seem to be consistent with [our recent discussion](https://github.com/exercism/xjava/pull/186#discussion_r89926622) on `binary-search`?

Ahhh... my apologies, Stuart. I was not paying close enough attention / thinking holistically enough to catch the fact that the type variable was being *required* by the test suite. In my comments in that discussion, I was focused on the example code, not the test.

Whether or not the example code uses more language features than are strictly required by the test suite, I currently believe is largely an implementation detail.  If the contributor wants to write a super generic solution... as long as it passes the tests and *at least* covers the behavior, I'm personally not too terribly concerned... I'm sure there's a length to the leash, but I haven't seen it yet.

But the tests themselves.... well!  That's another story!  I think we should be strict and intentional in the tests as they are the heart of the user experience on the track.

Did I clear things up?

jtigger:
I see this is also the case for `binary-search-tree`... and there are other collections like `simple-linked-list` and `custom-set`. I guess this question is in three parts:

1. should we use generics (especially in exercises that call for creating a collection)?
2. if so, should we do it for exercise X?
3. if so, should we extend the test coverage on our own, or should we also submit back to x-commons?

I'm hearing rough consensus that the answer to 1 is "YES!"

stkent:
[...]

PR #868, opened 2017-10-06
binary-search: use generics

FridaTveit:
The binary-search tests force the practitioner to generify the class
but only one type (integer) was actually being used in the tests.
This should be changed so that more than one type is actually used, to
make it more clear that generics are necessary and to demonstrate how
generics is used.

As discussed in #230, move linked-list to one of the last difficulty 6
exercises, so that binary-seach is the first exercise that introduces
generics.

Add a hint to the binary-search exercise to explain generics as this
is now the first exercise to require it.

Should hopefully fix the `binary-search` part of #230.

---

Reviewer Resources:


mortenaa:
Is this up for grabs? If so, I can do it :-)

FridaTveit:
What exactly do you mean by "up for grabs"? I've made this PR to solve the issue of using generics in `binary-search`. So this isn't an issue you can solve, it's a PR solving an issue. If you want to look through it and review it then you can :) There should still be plenty of `good-first-patch` issues to do if you want to do more PRs :)

mortenaa:
Ah, sorry, I thought I was looking at an issue :-)

FridaTveit:
I see, no problem :)

exercises/binary-search/.meta/hints.md
```diff
@@ -0,0 +1,24 @@
...
+The Java [List](https://docs.oracle.com/javase/8/docs/api/java/util/List.html) and [Map](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html) implementations are both examples of classes that use generics.
+By using them you can construct a `List` containing `Integers` or a list containing `Strings` or any other type.
+
+There are a few constraints on the types used in generics.
+One of them is that once you've constructed a `List` containing `Integers`, you can't put `Strings` into it.
+You have to specify which type you want to put into the class when you construct it, and that instance can then only be used with that type.
+
+For example you could construct a list of `Integers`:
+
+`List<Integer> someList = new LinkedList();`
```
stkent: Should use `LinkedList<>();` here.

exercises/binary-search/.meta/hints.md
```diff
@@ -0,0 +1,24 @@
...
+
+Now `someList` can only contain `Integers`. You could also do:
+
+`List<String> someOtherList = new LinkedList()`
```
stkent: Should use `LinkedList<>();` here.

exercises/binary-search/src/test/java/BinarySearchTest.java
```diff
@@ -13,23 +13,23 @@
 
     @Test
     public void findsAValueInAnArrayWithOneElement() {
-        List<Integer> listOfUnitLength = Collections.singletonList(6);
+        List<Character> listOfUnitLength = Collections.singletonList('6');
```
stkent: Same question as for `custom-set`; is there a pattern for the generic types being used now?

FridaTveit: Not really. I've tried to use `Integer`, `Character` and `String` and to make sure I've got a couple of tests using each, but I've not got a strict pattern as I didn't think it mattered too much. I could make it a more strict pattern if you think that would be useful?

stkent: Nah, I think it's good as-is!

FridaTveit: Will change it to cycle through `Character`, `String`, `Integer` as it's probably best to be consistent :)

FridaTveit: As I was changing it I remembered why I didn't do that. Some of them can't use `Character` because they use higher numbers, e.g. `11`... So it's only possible to use `Character` for some of the earlier tests, at least for this exercise. I could make the rest cycle through `String` and `Integer`?

FridaTveit: I didn't see your ealier comment (didn't refresh the page :P ), sorry! Should I just leave it as it is then? For `custom-set` and `binary-search-tree` too?

stkent: > As I was changing it I remembered why I didn't do that. Some of them can't use `Character` because they use higher numbers, e.g. `11`... So it's only possible to use `Character` for some of the earlier tests, at least for this exercise. I could make the rest cycle through `String` and `Integer`?

This sounds like a good compromise to me!

FridaTveit: Okay, will do that :) Will do that for the others too, try to cycle through all three but skip `Character` if not possible :)

FridaTveit: Sadly, as it turns out `String` causes problems as well since it will sort it lexicographically and we're using numbers in all our test cases... So should I instead try to cycle through all three, skip `Chracter` if not possible and also skip `String` if not possible? And make sure that `Character` and `String` are used at least once?

stkent: That sounds ideal, yep!

FridaTveit: Great, will do that then :)

FridaTveit:
Also removed an unused method that I found in the reference implementation :)

stkent:
❤️

exercises/binary-search/.meta/hints.md
```diff
@@ -0,0 +1,24 @@
...
+
+Now `someOtherList` can only contain `Strings`.
+
+Another constraint is that any type used with generics cannot be a [primitive type](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html), such as `int` or `long`.
```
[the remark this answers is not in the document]
m-a-ge: | Premitive Type | Reference Equvalent |
| -------------- | ------------------- |
| boolean | [Boolean](http://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html) |
| byte | [Byte](http://docs.oracle.com/javase/8/docs/api/java/lang/Byte.html) |
| short | [Short](http://docs.oracle.com/javase/8/docs/api/java/lang/Short.html) |
| char | [Character](http://docs.oracle.com/javase/8/docs/api/java/lang/Character.html) |
| int | [Integer](http://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html) |
| long | [Long](http://docs.oracle.com/javase/8/docs/api/java/lang/Long.html) |
| float | [Float](http://docs.oracle.com/javase/8/docs/api/java/lang/Float.html) |
| double | [Double](http://docs.oracle.com/javase/8/docs/api/java/lang/Double.html) |

PR #878, opened 2017-10-06
Add new exercise section to the contributing guide

FridaTveit:
When adding new exercises myself and reviewing other people's PRs for adding new exercises, I've missed having a guide which explains specifically what you need to do on the Java track and has links to general information you need about adding an exercise. So I thought it could be useful to add that to our contributing guide :)

I've also added a link to the general contributing guide because I think that could also be useful. I know in the README we link to both, but there are places where we only link to the Java guide, in which case people might not find the general one.

Please let me know if anything I've said is wrong, not well put or insufficient :) And of course if I've made any spelling mistakes or other errors :)

---

Reviewer Resources:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,7 @@
   * [The Problem Submodules](#the-problem-submodules)
 * [Advanced: Complete Local Setup](#advanced-complete-local-setup)
   * [Tip: `gradle clean` before `exercism fetch`](#tip-gradle-clean-before-exercism-fetch)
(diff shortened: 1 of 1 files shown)


stkent:
Looks great! j'adore!

m-a-ge:
Wow, maybe it'll be a good idea to update https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md as well?

FridaTveit:
Yes, that's a good point! I'll try and open a pull request to update it :)

PR #879, opened 2017-10-06
Added Topics to Collatz Conjecture #870

### LINKED ISSUE #870: collatz-conjecture: add topics

mortenaa:
Added Topics to Collatz Conjecture 
fixes #870

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -274,7 +274,14 @@
       "core": false,
       "difficulty": 4,
       "slug": "collatz-conjecture",
-      "topics": null,
+      "topics": [
+        "exception_handling",
+        "recursion",
+        "conditionals",
+        "equality",
+        "integers",
+        "mathematics"
+      ],
       "unlocked_by": "triangle",
       "uuid": "1500d39a-c9d9-4d3a-9e77-fdc1837fc6ad"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -274,7 +274,14 @@
...
       "difficulty": 4,
       "slug": "collatz-conjecture",
-      "topics": null,
+      "topics": [
+        "exception_handling",
+        "recursion",
+        "conditionals",
+        "equality",
+        "integers",
+        "mathematics"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `equality`. How do you think that's relevant to the exercise?

mortenaa: I was thinking about the termination check `start == 1` and the check if the number is even `start % 2 == 0`
but if the equality topic is more about equality of objects, i'll remove it :-)

FridaTveit: Yes, I think the equality topic should be more about equality of objects, otherwise we could add it to almost every exercise :) Thanks for explaining! :)

FridaTveit:
Will merge it as soon as the build finishes :)

PR #880, opened 2017-10-06
Added stubs to Matrix #547

### LINKED ISSUE #547: matrix: create and/or validate starter implementations

mortenaa:
fixes #547

---

Reviewer Resources:

### DIFF
--- a/exercises/matrix/src/main/java/Matrix.java
+++ b/exercises/matrix/src/main/java/Matrix.java
@@ -1,24 +1,23 @@
-import java.util.regex.Pattern;
 
-public class Matrix {
+class Matrix {
 
-    public Matrix(String matrixAsString) {
+    Matrix(String matrixAsString) {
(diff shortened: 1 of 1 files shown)


exercises/matrix/src/main/java/Matrix.java
```diff
@@ -0,0 +1,24 @@
+import java.util.regex.Pattern;
```
FridaTveit: I don't think you need this import :)

exercises/matrix/src/main/java/Matrix.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class Matrix {
```
FridaTveit: The class and the methods can all be [`package private`](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html). It's best practice to be as restrictive as possible with access modifiers :)

PR #881, opened 2017-10-06
complex-numbers: Add topics for exercise to config.json per #873

### LINKED ISSUE #873: complex-numbers: add topics

m-a-ge:
Resolves #873 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -796,7 +796,11 @@
       "core": false,
       "difficulty": 8,
       "slug": "complex-numbers",
-      "topics": null,
+      "topics": [
+        "mathematics",
+        "floating_point_numbers",
+        "classes"
+      ],
       "unlocked_by": "triangle",
       "uuid": "52d11278-0d65-4b5b-b387-1374fced3243"
     },


config.json
```diff
@@ -796,7 +796,12 @@
       "core": false,
       "difficulty": 8,
       "slug": "complex-numbers",
-      "topics": null,
+      "topics": [
+        "mathematics",
+        "floating_point_numbers",
+        "classes",
+        "structs"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `structs`. `structs` aren't really a thing in Java as far as I'm aware so I would hesitate to add it as a topic. How do you feel it's relevant to the exercise?

m-a-ge: @FridaTveit you are right, it's from `C`, I thought of it as shorthand for `structures`. 
Not sure why it is presented in `TOPICS.txt` since it's a language specific thing. Probably it should be removed

m-a-ge:
@FridaTveit PR updated accordingly

config.json
FridaTveit: It's certainly not present in all languages but it is present in several. I think their plan with the topics list was to have a list of topics that could be used in more than one language, but not necessarily all. That way the topics will be consistent across tracks. In that case `structs` should still be there. I could very well be wrong though :)

m-a-ge: Ah, it makes sense

PR #882, opened 2017-10-06
Added topics to Matrix exercise. Fixes #852

### LINKED ISSUE #852: matrix: add topics

Nik-Mouz:
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -199,7 +199,12 @@
       "core": true,
       "difficulty": 4,
       "slug": "matrix",
-      "topics": null,
+      "topics": [
+	"loops",
+	"arrays",
+	"pattern_matching",
+	"integers"
+	],
       "unlocked_by": null,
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -199,7 +199,12 @@
       "core": true,
       "difficulty": 4,
       "slug": "matrix",
-      "topics": null,
+      "topics": [
+	"loops",
+	"arrays",
+	"pattern_matching",
+	"integers"
```
FridaTveit: Great suggestions! :) I would also add `strings` since it uses strings, and `type_conversion` since you have to convert a `string` to an `int`. Do you agree? :)

Also all these topics should be indented two spaces in from `topics`. And the closing square bracket should be indented in line with `topics` and `unlocked_by`. Does that make sense? :)

Nik-Mouz:
Hey updated the file with your suggested changes, thanks for being so friendly!

PR #885, opened 2017-10-07
Adding topics to problem 'Prime Factors'

bobatrance:
Fixes issue #872 regarding prime factors.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -405,6 +405,7 @@
         "integers",
         "lists",
         "loops",
+        "mathematics",
         "recursion"
       ],
       "unlocked_by": "triangle",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -399,7 +399,15 @@
...
       "slug": "prime-factors",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "conditionals",
+        "integers",
+        "lists",
+        "loops",
+        "mathematics",
+        "recursion"
```
m-a-ge: Just for your consideration: use of `recursion` is discouraged in Java

FridaTveit: Great suggestions! :) But I agree with @m-a-ge, I wouldn't add `recursion`. You sould be able to solve the problem without using recursion and using iteration is more suitable for Java :) What do you think? :)

PR #886, opened 2017-10-07
Add topics to sieve

### LINKED ISSUE #847: sieve: add topics

puripant:
How about another tag for `memorization`? It is not listed in topics.txt though. Should fix #847 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -233,7 +233,8 @@
         "loops",
         "arrays",
         "integers",
-        "mathematics",
+        "algorithms",
+        "mathematics"
       ],
       "unlocked_by": "difference-of-squares",
       "uuid": "6791d01f-bae4-4b63-ae76-86529ac49e36"
--- a/config.json
+++ b/config.json
@@ -231,7 +231,7 @@
       "slug": "sieve",
       "topics": [
         "loops",
-        "arrays",
+        "lists",
         "integers",
         "algorithms",
         "mathematics"
--- a/config.json
+++ b/config.json
@@ -229,7 +229,12 @@
       "core": false,
       "difficulty": 4,
       "slug": "sieve",
-      "topics": null,
+      "topics": [
+        "loops",
+        "arrays",
+        "integers",
+        "mathematics",
+      ],
       "unlocked_by": "difference-of-squares",
       "uuid": "6791d01f-bae4-4b63-ae76-86529ac49e36"
     },


config.json
```diff
@@ -229,7 +229,12 @@
       "core": false,
       "difficulty": 4,
       "slug": "sieve",
-      "topics": null,
+      "topics": [
+        "loops",
+        "arrays",
+        "integers",
+        "mathematics",
```
FridaTveit: There shouldn't be a comma after the last topic, that's why the build is failing :)

FridaTveit: Also I would add `lists` and maybe `algorithms` since this uses a specific algorithm. What do you think? 

You said you wanted to add `memorization`. How do you think that relates to this exercise? Generally the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) should contain all topics that could be shared between tracks. So if you have a topic you want to add that isn't there then there are two choices:
1. If you think the topic is specific to Java then feel free to just add it to an exercise.
2. If the topic is more general then you should open a PR to the [problem specifications repository](https://github.com/exercism/problem-specifications/) to add it to the topics list. Then you can add it to an exercise here. 

Does that make sense? :)

puripant: If `algorithms` is added, maybe `memorization` is not necessary. Wouldn't `lists` be redundant, given that `lists` is there?

FridaTveit: What do you mean "given that `lists` is there"? It's not in the lists of topics you've added from what I can see. Do you mean that lists is there in the tests? That's the point of the topics, they're meant to list things you will or may encounter when doing the exercise. This includes things used in the tests. One of the reasons we're adding topics is so that people can choose to do exercises to do with a certain topic, and they can easily see at a glance what topics an exercise involves without having to look at the tests. Does that make sense? :)

puripant: Sorry, I meant to say `lists` would be redundant to `arrays` in this exercise. Also, does the tag `lists` refer to linked lists?

m-a-ge: @puripant `lists` and `arrays` are two very different things in Java.
Moreover, [`LinkedList`]**http://download.java.net/java/jdk9/docs/api/java/util/LinkedList.html) is just an implementation of an interface [`List` ](http://download.java.net/java/jdk9/docs/api/java/util/List.html). So, it's better to stick with interfaces in topics

FridaTveit: Yes, like @m-a-ge said, [`lists`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html) and [`arrays`](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/arrays.html) are different :) The tests expect you to return a list so I would definitely add `lists` as a topic. You don't need to have `arrays` as a topic since you don't have to use arrays to solve this exercise. But you could use arrays internally if you wanted to, so I don't mind if you keep it :) Does that make sense?

puripant: I understand the difference between arrays and lists in Java but I just don't understand the point of adding a topic that is related to the implementation of a problem or its test. I cannot imagine anyone would come to practice programming with problems related to `lists` but not to `arrays` or vice versa. For this problem, a solution, which would not pass the test, could easily return an array instead of any list. In a high-level abstraction, at least for this problem, arrays and lists are essentially the same.

I think this is beyond the scope of this issue anyway. And I probably don't understand the objectives of this project well enough. I have replaced `arrays` with `lists`.

m-a-ge: @puripant this is a good point, that's why choosing topics is very difficult.
But I think it would be a good starting point to add them based on reference solution and then modify them according to learners' solutions in future

PR #890, opened 2017-10-08
Tournament

mortenaa:
Hey, I had a go at adding the Tournament exercise, hope this is ok, let me know if I've forgotten something :-)

---

Reviewer Resources:


m-a-ge:
Wow, this looks massive 👍 

You forgot to add it in [`config.json`](https://github.com/exercism/java/blob/master/config.json), that's why tests failed
> An implementation for 'tournament' was found, but config.json does not reference this exercise.

`uuid` can be generated using online generator
Common list of topics among tracks - https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt

exercises/tournament/src/main/java/Tournament.java
```diff
@@ -0,0 +1,18 @@
+
```
FridaTveit: You only need this started implementation if this exercises is one of the first 20 exercises on the track (i.e. one of the first 20 exercises in `config.json`), which I don't think this should be as it's relatively difficult. See our [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations) for policies like this one. So please remove this file :) Unless you think it should be one of the first 20 exercises. When you add an entry to `config.json` for this exercise you can estimate the difficulty for it :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -0,0 +1,158 @@
...
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void justTheHeaderIfNoInput() {
+        Tournament tournament = new Tournament();
+        assertEquals("Team                           | MP |  W |  D |  L |  P", tournament.printTable());
+    }
+
+    @Test
```
FridaTveit: This (and all the following tests) need to be igored. See our [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests) for a description of how :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -0,0 +1,158 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class TournamentTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void justTheHeaderIfNoInput() {
+        Tournament tournament = new Tournament();
```
FridaTveit: This line is repeated in all the tests to it can be put in a [`@Before` block](https://www.mkyong.com/unittest/junit-4-tutorial-1-basic-usage/) :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test
+    public void aWinIsThreePointsALossIsZeroPoints() throws Exception {
+        Tournament tournament = new Tournament();
+        tournament.applyResults("Allegoric Alaskans;Blithering Badgers;win");
+        assertArrayEquals(new String[]{
+                "Team                           | MP |  W |  D |  L |  P",
+                "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3",
+                "Blithering Badgers             |  1 |  0 |  0 |  1 |  0"
+                }, tournament.printTable().split("\n"));
```
FridaTveit: Why don't you just compare the string? I think that makes it more obvious to the users what they need to implement. What do you think? :)

exercises/tournament/.meta/src/reference/java/Tournament.java
```diff
@@ -0,0 +1,108 @@
...
+                    home.applyResult(Result.DRAW);
+                    away.applyResult(Result.DRAW);
+                    break;
+            }
+            results.put(column[0], home);
+            results.put(column[1], away);
+        }
+    }
+
+    enum Result {
```
FridaTveit: I would put both the enum and the `TeamResult` class in a separate file :) That way this file doesn't get too big and its more readable :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -0,0 +1,158 @@
...
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void justTheHeaderIfNoInput() {
+        Tournament tournament = new Tournament();
+        assertEquals("Team                           | MP |  W |  D |  L |  P", tournament.printTable());
+    }
+
+    @Test
+    public void aWinIsThreePointsALossIsZeroPoints() throws Exception {
```
FridaTveit: Why do you have `throws Exception` here? If you're expecting it to throw an exception you should specify exactly which exception you're expecting. But from what I can tell your methods don't throw any exceptions?

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class TournamentTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
```
FridaTveit: Why do you need this? You only need this if you're expecting an exception to be thrown in one of your tests and from what I can tell you're not :)

mortenaa:
Thanks for the feedback! I'll try to address your issues

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -1,158 +1,160 @@
...
 import static org.junit.Assert.assertEquals;
 
 public class TournamentTest {
 
-    @Rule
-    public ExpectedException expectedException = ExpectedException.none();
+    private Tournament tournament;
+
+    @Before
+    public void setUp() throws Exception {
```
FridaTveit: I don't think this needs `throws Exception`. You should never just throw `Exception` without specifying which exception you're throwing, but in this case you're not throwing any as far as I'm aware, so you don't need it at all :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -1,158 +1,160 @@
...
     public void justTheHeaderIfNoInput() {
-        Tournament tournament = new Tournament();
-        assertEquals("Team                           | MP |  W |  D |  L |  P", tournament.printTable());
+        assertEquals("Team                           | MP |  W |  D |  L |  P\n", tournament.printTable());
     }
 
     @Test
-    public void aWinIsThreePointsALossIsZeroPoints() throws Exception {
-        Tournament tournament = new Tournament();
+    @Ignore
```
FridaTveit: This should have:
```
@Ignore("Remove to run test")
@Test
```

How to add `@Ignore`s is described in the [`Ignore` section of the POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests) :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -1,158 +1,160 @@
...
-        assertArrayEquals(new String[]{
-                "Team                           | MP |  W |  D |  L |  P",
-                "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3",
-                "Blithering Badgers             |  1 |  0 |  0 |  1 |  0"
-                }, tournament.printTable().split("\n"));
+        assertEquals(
+                "Team                           | MP |  W |  D |  L |  P\n" +
+                "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3\n" +
+                "Blithering Badgers             |  1 |  0 |  0 |  1 |  0\n"
+                , tournament.printTable());
```
FridaTveit: The indentation and position of the comma here and in later tests is a bit inconsistent. The comma should come after the first parameter (on the same line) and the second parameter should be indented four spaces in from `assertEquals`. I would indent it like this;
```
 assertEquals(
    "Team                           | MP |  W |  D |  L |  P\n" +
    "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3\n" +
    "Blithering Badgers             |  1 |  0 |  0 |  1 |  0\n" ,
    tournament.printTable()
);

I also see now that the canonical data has the output as an array (sorry, didn't look at it closely enough before). I think it's fine to leave your output as a string if you prefer, it shouldn't matter too much. Or you can have it as an array, but in that case I think the `printTable` method should return an array, it shouldn't be converted to an array in the tests. Does that make sense? Do you agree? But what you have at the moment it fine as well :)

File: config.json [hunk 11]
```diff
@@ -679,6 +679,14 @@
     },
     {
       "core": false,
+      "difficulty": 6,
+      "slug": "tournament",
+      "topics": null,
```
FridaTveit: If you want to add some topics to this exercise then that would be great :) It's fine if you don't want to though, I can add it to the list of exercises we still need to add topics to :)

FridaTveit:
Build seems to have succeeded now, must have just been a random failure :)

exercises/tournament/src/test/java/TournamentTest.java
mortenaa: @FridaTveit I blame intellij for that one :-)

FridaTveit: Fair enough :)

config.json
```diff
@@ -694,7 +694,13 @@
       "core": false,
       "difficulty": 6,
       "slug": "tournament",
-      "topics": null,
+      "topics": [
+        "loops",
+        "maps",
+        "sorting",
+        "parsing",
+        "text_formatting"
```
FridaTveit: Great! thanks for adding these :)

exercises/tournament/src/test/java/TournamentTest.java
```diff
@@ -19,92 +19,92 @@ public void justTheHeaderIfNoInput() {
     }
 
     @Test
-    @Ignore
+    @Ignore("Remove to run test")
```
FridaTveit: We would normally put the `@Ignore` tag above the `@Test` tag to be consistent with the other test files. It doesn't matter very much though :)

config.json
```diff
@@ -692,6 +692,20 @@
...
+      "difficulty": 6,
+      "slug": "tournament",
+      "topics": [
+        "loops",
+        "maps",
+        "sorting",
+        "parsing",
+        "text_formatting"
+      ],
+      "unlocked_by": "kindergarten-garden",
```
FridaTveit: I almost didn't notice this; this needs to be a core exercise, sorry! Exercises can only be unlocked by core exercises, and `kindergarten-garden` isn't a core exercise. Sorry for not spotting that sooner! `kindergarten-garden` is unlocked by `matrix` which is a core exercise, so maybe you want it to be unlocked by `matrix`? Or you can choose some other exercise if you want :)

mortenaa: Ah, my fault for just jumping in without reading the contributing guide i guess :-)

mortenaa:
oops, looks like the build failed the same way as earlier

exercises/tournament/src/test/java/TournamentTest.java
FridaTveit: Thanks for changing these! :) 👍

PR #891, opened 2017-10-09
Renamed classes and added @Before method to test class

### LINKED ISSUE #376: binary-search-tree: extract initialization and rename test class
In the `binary-search-tree` exercise, the test class is named `BSTTest` and the target class name is `BST`.  For better or for worse, idiomatic Java avoids abbreviations.  Also, each test method creates its own instance of `BST`.  Unfortunately, [Shalloway's Law](http://www.netobjectives.com/blogs/shalloways-law-and-shalloways-principle) applies.

Please:
1. Rename the test suite to a more descriptive name.
1. Rename the target class to a congruent name (ideally, the name of the test suite is the name of the class with a `Test` suffix).
1. Extract construction of the subject under test in [`setUp()`](http://junit.sourceforge.net/javadoc/org/junit/Before.html).

delbertlegg:
Fixes #376 

---

Reviewer Resources:

### DIFF
--- a/exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
+++ b/exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
@@ -6,15 +6,21 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.junit.Before;
 
-public class BSTTest {
+public class BinarySearchTreeTest {
+    private BinarySearchTree<Integer> binarySearchTree;
+
+    @Before
+    public void setUp() {
+        binarySearchTree = new BinarySearchTree<>();
+    }
 
     @Test
     public void dataIsRetained() {
-        BST<Integer> bst = new BST<>();
         final int actual = 4;
-        bst.insert(actual);
-        final BST.Node<Integer> root = bst.getRoot();
+        binarySearchTree.insert(actual);
+        final BinarySearchTree.Node<Integer> root = binarySearchTree.getRoot();
         assertNotNull(root);
         final int expected = root.getData();
         assertEquals(expected, actual);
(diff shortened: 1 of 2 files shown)


exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
```diff
@@ -6,15 +6,21 @@
 import static org.junit.Assert.assertNotNull;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.junit.Before;
 
-public class BSTTest {
+public class BinarySearchTreeTest {
+    BinarySearchTree<Integer> bst;
```
FridaTveit: Just as a heads up, this will probably be changed as part of  issue #230 so that not all the binary search trees used in the tests will be of type `Integer`. In which case we won't be able to save it as a private variable anymore :( So it's fine to leave it as it is for now, as the tests are it's a really good change :) But just be aware that it might be changed back in the near future :)

Also also, if you do decide to keep it then it should be made `private` :)

exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
```diff
@@ -6,15 +6,21 @@
...
+
+    @Before
+    public void setUp() {
+        bst = new BinarySearchTree<>();
```
FridaTveit: I would probably rename this as well to `binarySearchTree` because I find that clearer. That's just my preference though, feel free to leave it as it is if you prefer it that way :)

delbertlegg:
@FridaTveit I can make those changes, no problem. Quick question though...do you want me to change the BinarySearchTree type to Object or leave the instantiation in the tests like it was before?

delbertlegg:
@FridaTveit I can make those changes, no problem. It looks like part of #376 was changed when #230 was made? It sounded like the intent was to instantiate the BinarySearchTree once for the tests, but now it looks like we can't since it may change types for each test?

FridaTveit:
True, I didn't see that part of #376, sorry! I think it's fine to do what you've done and instantiate it in a `@Before` block, that's the best thing to do at the moment. But it will probably be changed once #230 is done. So I just wanted to let you know in case you saw that change and were annoyed that your changes were being undone :) Does that make sense? Sorry if I've caused any confusion!

delbertlegg:
@FridaTveit Makes sense, and no worries! I just want to make sure I understand the problem. I'll change the variable name and make it private for now. And I won't be annoyed about any changes, here to help :).

PR #901, opened 2017-10-09
saddle-points: make MatrixCoordinate implement Comparable

FridaTveit:
The saddle-points tests make it seem like you can use any kind of set for your output. However, if you use `TreeSet` then the tests will fail because `MatrixCoordinate` doesn't implement `Comparable`.

`MatrixCoordinate` does now implement `Comparable` to prevent this failure.

Another solution would be to make the tests enforce use of `HashSet` instead of `TreeSet`. I'd be happy to change to that solution if anyone thinks that's preferable :)

---

Reviewer Resources:

### DIFF
--- a/exercises/saddle-points/.meta/src/reference/java/MatrixCoordinate.java
+++ b/exercises/saddle-points/.meta/src/reference/java/MatrixCoordinate.java
@@ -1,4 +1,4 @@
-class MatrixCoordinate {
+class MatrixCoordinate implements Comparable<MatrixCoordinate> {
     private final int row;
     private final int col;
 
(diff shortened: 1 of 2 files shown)


stkent:
I'm curious what the failure looked like when using `TreeSet`?

FridaTveit:
```
 java.lang.ClassCastException: MatrixCoordinate cannot be cast to java.lang.Comparable
        at java.util.TreeMap.compare(TreeMap.java:1294)
        at java.util.TreeMap.put(TreeMap.java:538)
        at java.util.TreeSet.add(TreeSet.java:255)
        at Matrix.getSaddlePoints(Matrix.java:26)
        at MatrixTest.testCanIdentifySingleSaddlePoint(MatrixTest.java:23)
```

stkent:
Ah, makes sense!

PR #902, opened 2017-10-09
added topics in bracket-push, simple-cipher, crpto-square and atbash-…

### LINKED ISSUE #899: bracket-push: add topics

anu0012:
…cipher

Fixes #899 #893 #892 #877 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -486,7 +486,10 @@
       "core": false,
       "difficulty": 5,
       "slug": "bracket-push",
-      "topics": null,
+      "topics": [
+        "stacks",
+        "strings"
+      ],
       "unlocked_by": "flatten-array",
       "uuid": "85aa50ac-0141-49eb-bc6f-62f3f7a97647"
     },
@@ -508,7 +511,11 @@
       "core": false,
       "difficulty": 5,
       "slug": "atbash-cipher",
-      "topics": null,
+      "topics": [
+        "strings",
+        "cryptography",
+        "security"
+      ],
       "unlocked_by": "rotational-cipher",
       "uuid": "d36ce010-210f-4e9a-9d6c-cb933e0a59af"
     },
@@ -785,7 +792,13 @@
       "core": false,
       "difficulty": 7,
       "slug": "crypto-square",
-      "topics": null,
+      "topics": [
+        "strings",
+        "text_formatting",
+        "mathematics",
+        "cryptography",
+        "security"
+      ],
       "unlocked_by": "rotational-cipher",
       "uuid": "162bebdc-9bf2-43c0-8460-a91f5fc16147"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -839,7 +852,12 @@
       "core": false,
       "difficulty": 8,
       "slug": "simple-cipher",
-      "topics": null,
+      "topics": [
+        "strings",
+        "cryptography",
+        "security",
+        "mathematics"
```
FridaTveit: I would add `randomness` and `exception_handling` as well.

config.json
```diff
@@ -775,7 +782,13 @@
       "core": false,
       "difficulty": 7,
       "slug": "crypto-square",
-      "topics": null,
+      "topics": [
+        "strings",
+        "text_formatting",
+        "mathematics",
+        "cryptography",
+        "security"
```
FridaTveit: I would add `lists` as well

anu0012:
@FridaTveit Thanks a lot for your valuable suggestions. I'll definitely take care of those in future. Just one more thing should I make another PR for each issue separately?

FridaTveit:
If you make any more PRs to fix other issues then yes :) But for this one you can keep it as it is since I've already reviewed it and I can imagine it would be a bit of a pain for you to try and split it into four PRs now :) You could do if you wanted to though, I don't particularly mind :)

anu0012:
@FridaTveit If possible you can merge this, I'll be more careful in my future PRs :)

FridaTveit:
Sure. But could you add the topics I suggested first? Or explain why you don't think they don't should be added :)

config.json
anu0012: I think it can be done without using lists also. Strings can be separated using Split() function also.

config.json
anu0012: Yes, exception_handling can be used also but at the same time, we can check the input manually without worrying about the exception. Actually, I didn't think in a language-specific way.

config.json
FridaTveit: No, the `getPlaintextSegments()` method needs to return a `List` in the [tests](https://github.com/exercism/java/blob/master/exercises/crypto-square/src/test/java/CryptoSquareTest.java) so you have to use lists :)

anu0012: Alright. I'll add it. Thanks for the correction :)

config.json
FridaTveit: Again the [tests](https://github.com/exercism/java/blob/master/exercises/simple-cipher/src/test/java/SimpleCipherStepThreeTest.java) expect an exception to be thrown so you have to use exceptions. These are meant to be language specific topics for doing the exercises on the java track :)

config.json
FridaTveit: No problem :) Thanks for explaining your reasoning :)

FridaTveit:
This has been fixed in PR #910.

PR #903, opened 2017-10-09
added topics for robot simulator

PatrM:
Added topics for #849.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -629,7 +629,6 @@
       "slug": "robot-simulator",
       "topics": [
         "classes",
-        "algorithms",
         "logic",
         "loops",
         "enumerations"
--- a/config.json
+++ b/config.json
@@ -622,7 +622,13 @@
       "core": false,
       "difficulty": 6,
       "slug": "robot-simulator",
-      "topics": null,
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -622,7 +622,13 @@
       "core": false,
       "difficulty": 6,
       "slug": "robot-simulator",
-      "topics": null,
+      "topics": [
+        "classes",
+        "algorithms",
+        "logic",
+        "loops",
+        "enumerations"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `algorithms`. You could argue that you're using an algorithm to solve this exercise, but then you could make that argument for every exercise and I'm not sure it would be very useful to add it as a topic to every exercise. So I think it's best to reserve `algorithms` for exercises that require you to implement a specific algorithm, like `luhn` for example. Does that make sense? Do you agree?

FridaTveit:

PatrM:
Thanks for your feedback Frida, I think you're right. As you already saw I removed _algorithms_ from the topics

PR #906, opened 2017-10-10
robot-simulator: improve equals for GridPosition

FridaTveit:
We recently improved the implementation of equals for `GridPosition` so that we could use `assertEquals` in the tests. However, we only changed the reference implementation, not the starter implementation. This changes the starter implementation to be the same as the reference implementation.

---

Reviewer Resources:


stkent:
Nice catch!

PR #908, opened 2017-10-10
topics added for rotational-cipher

PatrM:
Added topics for #875. `ascii` and `chars` aren't in the topics list but I think these are fine here.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -629,7 +629,6 @@
       "slug": "robot-simulator",
       "topics": [
         "classes",
-        "algorithms",
         "logic",
         "loops",
         "enumerations"
--- a/config.json
+++ b/config.json
@@ -263,7 +263,13 @@
       "core": true,
       "difficulty": 4,
       "slug": "rotational-cipher",
-      "topics": null,
+      "topics": [
+        "integers",
+        "ascii",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -263,7 +263,13 @@
       "core": true,
       "difficulty": 4,
       "slug": "rotational-cipher",
-      "topics": null,
+      "topics": [
+        "integers",
+        "ascii",
+        "chars",
+        "strings",
+        "cryptography"
```
FridaTveit: Great suggestions! :) I'm not sure about `ascii` and `chars` though. I think `chars` is pretty much covered by `strings` so I don't think we need it as a separate topic. I'm also not sure `ascii` is that relevant to this exercise since you don't really need to know anything about `ascii` to solve the exercise as far as I'm aware?

PatrM:
`chars` covered by `strings` is actually true 👍 . I thought about adding `ascii` as additional topic just so people could read more into the topic of characters, but im fine with leaving it out :).

Thanks a lot for the link on the git issue, I think that will help me just enough

I will close this PR and open a new one with only one commit

PR #912, opened 2017-10-10
Adding the topics for the clock example

### LINKED ISSUE #896: clock: add topics

Jiri-Kremser:
fixes #896
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -805,7 +805,14 @@
       "core": false,
       "difficulty": 7,
       "slug": "clock",
-      "topics": null,
+      "topics": [
+        "equality",
+        "integers",
+        "logic",
+        "object_oriented_programming",
+        "strings",
+        "time"
+      ],
       "unlocked_by": "saddle-points",
       "uuid": "97c8fcd4-85b6-41cb-9de2-b4e1b4951222"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -805,7 +805,14 @@
...
       "difficulty": 7,
       "slug": "clock",
-      "topics": null,
+      "topics": [
+        "classes",
+        "equality",
+        "integers",
+        "logic",
+        "object_oriented_programming",
+        "time"
```
FridaTveit: Great suggestions! The only ones I'm not sure object are `object_oriented_programming` and `classes`. They do apply to this exercise but they also apply to every other exercise and I'm not sure it would be useful to add them to every exercise. What do you think?

I would also add `strings` as a topic since you have to implement `toString` :)

Jiri-Kremser: I've added `strings` and removed `classes` (no new class was created when solving this exercise). However, I left the `object_oriented_programming` in place, because IMHO `hashcode()` and `equals()` methods are pretty inherent to OOP in Java.

FridaTveit: Okay, fair enough :)

PR #927, opened 2017-10-11
added topics for robot-name

PatrM:
added topics for #920. 
I feel like something is missing though. What do you think @FridaTveit?

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -434,7 +434,11 @@
       "core": true,
       "difficulty": 5,
       "slug": "robot-name",
-      "topics": null,
+      "topics": [
+        "regular_expressions",
+        "strings",
+        "randomness"
+      ],
       "unlocked_by": null,
       "uuid": "d7c2eed9-64c7-4c4a-b45d-c787d460337f"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -434,7 +434,11 @@
       "core": true,
       "difficulty": 5,
       "slug": "robot-name",
-      "topics": null,
+      "topics": [
+        "regular_expressions",
+        "strings",
+        "randomness"
```
FridaTveit: Maybe `text_formatting` and `pattern_matching` could be useful? I think you've included the most important ones though, so I don't mind if you add those two or not :)

PatrM: `text_formatting` is a good idea, I gonna add that later on. 
Isn't `pattern_matching` already covered by regex though?

FridaTveit: You could try and look for something matching a pattern without using a regex. And some people might be looking for exercises to do with pattern matching without thinking to look for exercises to do with regular expressions. But you're right, they are very similary so I'm happy for you to just stick with `regular_expressions` if you think that's best :)

FridaTveit:

PR #928, opened 2017-10-11
Added topics for sublist problem

### LINKED ISSUE #924: sublist: add topics

badanomaly:
fixes #924

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -760,7 +760,13 @@
       "core": false,
       "difficulty": 7,
       "slug": "sublist",
-      "topics": null,
+      "topics": [
+        "lists",
+        "searching",
+        "loops",
+        "generics",
+        "enumeration"
+      ],
       "unlocked_by": "linked-list",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -760,7 +760,14 @@
...
       "difficulty": 7,
       "slug": "sublist",
-      "topics": null,
+      "topics": [
+        "lists",
+        "sequences",
+        "logic",
+        "equality",
+        "searching",
+        "loops"
```
FridaTveit: Great suggestions! :) I agree that `lists`, `searching` and `loops` are good topics for this exercise :)

I'm not sure about are `sequences`, `logic` and `equality` however. How do you think they relate to this exercise? `logic` and `equality` are used in this exercise, but they are used in most exercises. I would save `equality` for exercises that require you to for example override `equals`. I would save `logic` for exercises that are more logic heavy than this. Does that make sense? I'm not sure that `sequences` is a good topic for Java in general. Feel free to disagree with me though! :)

I would also add `generics` and `enumerations` :)

PR #931, opened 2017-10-12
added topics for book store

PatrM:
added topics for #915 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -900,7 +900,12 @@
       "core": false,
       "difficulty": 8,
       "slug": "book-store",
-      "topics": null,
+      "topics": [
+        "integers",
+        "lists",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -900,7 +900,12 @@
       "core": false,
       "difficulty": 8,
       "slug": "book-store",
-      "topics": null,
+      "topics": [
+        "integers",
+        "lists",
+        "algorithms",
+        "mathematics"
```
FridaTveit: Great suggestions! I would add `floating_point_numbers` as well since the price returned is a double :)

PatrM: good point, I'm adding it 👍

PR #932, opened 2017-10-13
Issue #923 add topics to linked list

ryan-neff:
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -643,7 +643,13 @@
       "core": true,
       "difficulty": 6,
       "slug": "linked-list",
-      "topics": null,
+      "topics": [
+        "algorithms",
+        "logic",
+        "lists",
+        "integers",
+        "conditionals"
+      ],
       "unlocked_by": null,
       "uuid": "7ba5084d-3b75-4406-a0d7-87c26387f9c0"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -643,7 +643,13 @@
       "core": true,
       "difficulty": 6,
       "slug": "linked-list",
-      "topics": null,
+      "topics": [
+        "algorithms",
+        "logic",
+        "lists",
+        "integers",
+        "conditionals"
```
FridaTveit: These are good suggestions @ryan-neff :) I agree that `lists` and `algorithms` are good topics for this exercise. I would also add `generics` :)

I'm not sure about adding `logic`, `integers` and `conditionals` as topics. `logic` and `conditionals` are used in this exercise but they are used in most exercises and I wouldn't say that they're the main focus of this exercise. So it might be best to leave them out. `integers` are used but since using generics is partly the main focus of the exercise, you shouldn't need to think about integers at all when implementing it. Does that make sense? :)

ryan-neff: The config now has only algorithms, generics, and lists in its topics section f338ddd

FridaTveit:

PR #934, opened 2017-10-14
added topics for run length encoding

PatrM:
added topics for run length encoding #876 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -436,7 +436,7 @@
         "strings",
         "integers",
         "pattern_matching",
-	      "transforming"
+        "transforming"
       ],
       "unlocked_by": "rotational-cipher",
       "uuid": "4499a3f9-73a7-48bf-8753-d5b6abf588c9"
--- a/config.json
+++ b/config.json
@@ -432,7 +432,11 @@
       "core": false,
       "difficulty": 5,
       "slug": "run-length-encoding",
-      "topics": null,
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -432,7 +432,11 @@
       "core": false,
       "difficulty": 5,
       "slug": "run-length-encoding",
-      "topics": null,
+      "topics": [
+        "strings",
+        "integers",
+        "sequences"
```
FridaTveit: Great suggestions :) I agree with `strings` and `integers` but I'm not sure about `sequences`. How do you think that relates to the exercise?

I would also maybe add `transforming` and maybe `pattern_matching`? What do you think? :)

PatrM: good catch, I will add these changes 👍

config.json
```diff
@@ -432,7 +432,12 @@
       "core": false,
       "difficulty": 5,
       "slug": "run-length-encoding",
-      "topics": null,
+      "topics": [
+        "strings",
+        "integers",
+        "pattern_matching",
+	"transforming"
```
FridaTveit: A small indentation issue here. `transforming` should be indented two spaces afer `topics`, same as the other topics :)

PatrM:
sorry, I had some issues with vim and spacing 😅

PR #942, opened 2017-10-15
Add topics to list-ops exercise

### LINKED ISSUE #936: list-ops: add topics

themakshter:
Close #936

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -951,7 +951,13 @@
       "core": false,
       "difficulty": 8,
       "slug": "list-ops",
-      "topics": null,
+      "topics": [
+        "filtering",
+        "functional_programming",
+        "generics",
+        "lists",
+        "loops"
+      ],
       "unlocked_by": "linked-list",
       "uuid": "a9836565-5c39-4285-b83a-53408be36ccc"
     },


config.json
```diff
@@ -951,7 +951,16 @@
...
-      "topics": null,
+      "topics": [
+        "variables",
+        "lists",
+        "loops",
+        "filtering",
+        "searching",
+        "conditionals",
+        "callbacks",
+        "functional_programming"
```
FridaTveit: These look like good suggestions :) I agree with `functional_programming`, `filtering` and `lists`. `variables`, `loops` and `conditionals` are definitely used in this exercise but I wouldn't say they were important topics so I'm not sure they should be included. I'm not sure how `callbacks` and `searching` are relevant to the exercise. What do you think? :)

FridaTveit: And I would add `generics` as a topic :)

themakshter: I feel `variables` is too broad a term to apply honestly and I think `conditionals` don't come too much in play here either in my opinion. `loops` seems a decent shout so I think they might be okay to be included. I will update it accordingly.

PR #943, opened 2017-10-15
Add rectangles topics to config

### LINKED ISSUE #922: rectangles: add topics

themakshter:
Close #922

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -902,7 +902,12 @@
       "core": false,
       "difficulty": 8,
       "slug": "rectangles",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "logic",
+        "pattern_recognition",
+        "strings"
+      ],
       "unlocked_by": "robot-name",
       "uuid": "64cda115-919a-4eef-9a45-9ec5d1af98cc"
     },


config.json
```diff
@@ -902,7 +902,11 @@
       "core": false,
       "difficulty": 8,
       "slug": "rectangles",
-      "topics": null,
+      "topics": [
+        "algorithms",
+        "logic",
+        "pattern_recognition"
```
FridaTveit: Good suggestions :) I agree with `logic` and `pattern_recognition`. I would also add `strings` and `arrays`. I'm not sure about `algorithms`. You could technically add that as a topic to every exercise but I'm not sure that would be very useful. So I think it's best to only add `algorithms` to exercises where you have to use a specific algorithm which I don't think is true for this exercise.

themakshter: I understand what you mean. I will update once I am home. 

In the meanwhile, I suppose it would be good to update the other projects which have the same tags, don't you think? [The python config](https://github.com/exercism/python/blob/master/config.json#L729) has the topics that I have added in this PR

FridaTveit: No, the topics should not be the same across tracks. The topics are meant to be topics you might encounter when solving the exercise in this specific language. Those topics might very well be relevant to the python exercise. Therefore, to choose topics you should look at the Java implementation of the tests and the reference implementation (as explained in the issue). Does that make sense? :)

themakshter: I understand. Thank you for explaining, I will update it accordingly!

themakshter:
@FridaTveit I have updated with your suggested changes, please take a look when you have time

PR #945, opened 2017-10-15
Fix940 Added topics: bank-account

### LINKED ISSUE #940: bank-account: add topics

yashasingh:
Added topics to the question Bank Account. This PR fixes #940 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -753,7 +753,6 @@
       "difficulty": 6,
       "slug": "bank-account",
       "topics": [
-        "arrays",
         "concurrency",
         "exception_handling",
         "integers"
--- a/config.json
+++ b/config.json
@@ -752,7 +752,12 @@
       "core": true,
       "difficulty": 6,
       "slug": "bank-account",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "concurrency",
+        "exception_handling",
+        "integers"
+      ],
       "unlocked_by": null,
       "uuid": "a242efc5-159d-492b-861d-12a1459fb334"
     },


config.json
```diff
@@ -737,7 +742,14 @@
...
       "difficulty": 6,
       "slug": "bank-account",
-      "topics": null,
+      "topics": [
+      	"floating_point_numbers",
+      	"files",
+      	"mathematics",
+      	"logic",
+        "object_oriented_programming",
+      	"variables"
```
FridaTveit: Thanks for tackling this :) I'm not sure I agree with your suggestions. `floating_point_numbers` and `files` don't appear to be used in this exercise at all. `mathematics`, `logic`, `object_oriented_programming` and `variables` are all used but I wouldn't say they are very important topics. How do you feel they are relevant to the exercise?

I would add `exception_handling` since they tests expect an exception to be thrown, and `concurrency` since that is a very important part of the exercise. You could maybe also add `integers`. What do you think? :)

There are some indentation problems. Each topic should be indented two spaces after `topics`, like `object_oriented_programming`.

yashasingh: Alright, I'll come up with more accurate topics and update the PR

yashasingh:
@FridaTveit I've made the required changes. Please review.

config.json
```diff
@@ -752,7 +752,12 @@
       "core": true,
       "difficulty": 6,
       "slug": "bank-account",
-      "topics": null,
+      "topics": [
+        "arrays",
```
FridaTveit: Why did you add this as a topic? I don't think you need to use arrays to solve this exercise. I know arrays are used in one of the tests, but not in a way that requires the person implementing the exercise to use arrays at all. Does that make sense? So I'm not sure arrays are important enough to this exercise to require it to be added as a topic. What do you think?

yashasingh: :+1:

yashasingh: I included this topic because it was being used at one the places. But, now just as you mentioned , I agree that arrays don't have anything to do with problem concept , so I'm removing it. Thank you.

yashasingh:
@FridaTveit Please review.

PR #952, opened 2017-10-16
binary-search-tree: use generics

FridaTveit:
Should hopefully fix the `binary-search-tree` part of issue #230.

Also swaps `expected` and `actual` around in the first test as they were, as far as I can tell, the wrong way around.

---

Reviewer Resources:

### DIFF
--- a/exercises/binary-search-tree/src/main/java/BinarySearchTree.java
+++ b/exercises/binary-search-tree/src/main/java/BinarySearchTree.java
@@ -1,0 +1,35 @@
+import java.util.List;
+
+class BinarySearchTree<T extends Comparable<T>> {
+    void insert(T value) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    List<T> getAsSortedList() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    List<T> getAsLevelOrderList() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    Node<T> getRoot() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static class Node<T> {
+
+        Node<T> getLeft() {
+            throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+        }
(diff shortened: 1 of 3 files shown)


stkent:
These changes look great!

I wonder if we should also take this opportunity to provide stubbed APIs for this exercise. The main class signature is complex (`public class BinarySearchTree<T extends Comparable<T>>`) and the tests also rely on a `Node` type. Does that fit here, or should we make it a separate task @FridaTveit ?

FridaTveit:
Sure, I can add that to this PR :)

FridaTveit:
Do you think we should provide implementation for `Node`? We could extract it and add the class as starter implementation like we do for enums and e.g. `BoardCoordinate`? Or we could keep it inside `BinarySearchTree` and just provide a stubbed API for it? What do you think?

stkent:
I think keeping internal and stubbing the API for this exercise is probably the way forward; providing the `Node` implementation would be a bit much I think!

FridaTveit:
Okay, will do! :)

stkent:
@FridaTveit I see a new commit here, but not the `Node` stubbing. Is that something you'd still like to take care of here? Sorry if I'm jumping the gun and the PR is just not yet ready for re-review!

exercises/binary-search-tree/src/main/java/BinarySearchTree.java
```diff
@@ -0,0 +1,35 @@
...
+
+    List<T> getAsLevelOrderList() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    Node<T> getRoot() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    static class Node<T> {
```
FridaTveit: Here's the node stubbing :)

FridaTveit:
I've added two new commits, one of which is adding starter implementation including node stubbing. I put a comment on it to highlight where it is :)

exercises/binary-search-tree/src/main/java/BinarySearchTree.java
stkent: Nice!

PR #956, opened 2017-10-17
Spiral matrix : add topics

mohshabbir:
fixed issue #855

### DIFF
--- a/config.json
+++ b/config.json
@@ -214,12 +214,7 @@
       "core": true,
       "difficulty": 4,
       "slug": "triangle",
-      "topics": [
-        "integers",
-        "algorithms",
-        "mathematics",
-        "floating_point_numbers"
-      ],
+      "topics": null,
       "unlocked_by": null,
       "uuid": "ec268d8e-997b-4553-8c67-8bdfa1ecb888"
     },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -536,7 +536,13 @@
       "core": false,
       "difficulty": 6,
       "slug": "spiral-matrix",
-      "topics": null,
+      "topics": [
+        "loops",
+        "arrays",
+        "numbers",
+        "mathematics",
+        "logic"
```
FridaTveit: Great suggestions! :) I agree with `arrays`, `mathematics` and `loops`. I'm not so sure about `numbers` and `logic`. `numbers` isn't a topic listed in the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt), `integers` would probably be a better topic. `logic` is used in this exercise, but you could argue that about every exercise. I wouldn't say logic is very important in this exercise compared to other exercises so I would be tempted to leave that topic out.

I would also add `matrices` as a topic :)

mohshabbir: Hi @FridaTveit , yes i agree with your suggestion. I've changed it

PR #957, opened 2017-10-17
custom-set: use generics and tidy up tests

FridaTveit:
Should hopefully fix `custom-set` part of issue #230.

---

Reviewer Resources:


stkent:
Looks good! I just have a couple questions:

Was there a method to deciding when a test should use int/char/String? I can't see a pattern and it's not hugely important to make sure they balance out, but I normally cycle through to make sure they distribution is roughly equal. Up to you if you'd like to do something similar.

Also, in reviewing against the canonical data, it looks like this test is currently up to spec (or almost) up to [spec](https://github.com/exercism/problem-specifications/blob/master/exercises/custom-set/canonical-data.json). If you could spend a minute double-checking my work and adding a `version` file in `.meta`, that would be awesome!

FridaTveit:
As I just commented on the `binary-search` review, no I don't have a strict pattern for when a test should use int/char/string. Can change it to be more consistent if you think that would be useful? 

And yes, will make sure the tests follow the canonical data and add a `version` file :)

FridaTveit:
Will change it to cycle through `Character`, `String`, `Integer` as it's probably best to be consistent :)

stkent:
Looking 💯! Thanks for your hard work on this **and** hacktoberfest triage!

PR #969, opened 2017-10-23
Add topic to pythagorean-triplet

### LINKED ISSUE #874: pythagorean-triplet: add topics

the7thobserver:
fixes #874 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -1034,7 +1034,7 @@
       "difficulty": 9,
       "slug": "pythagorean-triplet",
       "topics": [
-        "integer",
+       "integer",
 	"lists",
 	"logic",
 	"mathematics"
--- a/config.json
+++ b/config.json
@@ -1033,7 +1033,10 @@
       "core": false,
       "difficulty": 9,
       "slug": "pythagorean-triplet",
-      "topics": null,
+      "topics": [
+		"mathematics",
+		"logic"
+	  ],
       "unlocked_by": "triangle",
       "uuid": "88505f95-89e5-4a08-8ed2-208eb818cdf1"
     },
--- a/config.json
+++ b/config.json
@@ -1034,8 +1034,10 @@
       "difficulty": 9,
       "slug": "pythagorean-triplet",
       "topics": [
-		"mathematics",
-		"logic"
+		"integer",
+		"lists",
+		"logic",
+		"mathematics"
 	  ],
       "unlocked_by": "triangle",
       "uuid": "88505f95-89e5-4a08-8ed2-208eb818cdf1"
--- a/config.json
+++ b/config.json
@@ -1034,11 +1034,11 @@
       "difficulty": 9,
       "slug": "pythagorean-triplet",
       "topics": [
-		"integer",
-		"lists",
-		"logic",
-		"mathematics"
-	  ],
+        "integer",
+	"lists",
+	"logic",
+	"mathematics"
+      ],
       "unlocked_by": "triangle",
       "uuid": "88505f95-89e5-4a08-8ed2-208eb818cdf1"
     },


config.json
```diff
@@ -1033,7 +1033,10 @@
       "core": false,
       "difficulty": 9,
       "slug": "pythagorean-triplet",
-      "topics": null,
+      "topics": [
+		"mathematics",
+		"logic"
```
FridaTveit: Great suggestions! I would also add `integers` and `lists` since you have to use both to solve the exercise :)

A couple of other minor things, the topics should be indented two spaces in from `topics` and the closing square bracket should be indented in line with `topics` and `unlocked_by`. Does that make sense?
And the topics should be sorted alphabetically :)

the7thobserver: I thought about adding data structures but I didn't want to limit them. Eg add lists but they want to use trees. Although I do agree most people would use lists. What do you think?

Ah I'll sort alphabetically,

FridaTveit: The [tests](https://github.com/exercism/java/blob/master/exercises/pythagorean-triplet/src/test/java/PythagoreanTripletTest.java) expect a list to be returned so you do have to use lists in some way to solve the exercise :) And the tests also expect you to use integers.

The way we look at the topics is to see them as things the user either definitely has to use to solve the exercise, or might want to use to solve the exercise. So it doesn't have to only be things they have to use. The only topics we don't want are things that are completely unrelated or things that we think would only be used in a relatively 'bad' solution.

Let me know if you disagree with any of that or if it didn't make sense :)

Great, thanks :)

the7thobserver: Ah that makes sense, will update.

config.json
```diff
@@ -1033,7 +1033,12 @@
       "core": false,
       "difficulty": 9,
       "slug": "pythagorean-triplet",
-      "topics": null,
+      "topics": [
+		"integer",
+		"lists",
+		"logic",
+		"mathematics"
+	  ],
```
FridaTveit: Sorry to be picky, but this square bracket should be indented one more space so that it is in line with `topics` and `unlocked_by`. And the the list of topics should be indented one more space so that they're two spaces in from `topics`. Does that make sense? It might help to look at entries for other exercises and copy their formatting :)

the7thobserver:
Ah no worries. This is weird, in Notpad++ it shows the spacing as the same as the others, but in git it's shifted over a bit.

Very strange but I've fixed it.

config.json
```diff
@@ -1033,7 +1033,12 @@
       "core": false,
       "difficulty": 9,
       "slug": "pythagorean-triplet",
-      "topics": null,
+      "topics": [
+        "integer",
+	"lists",
+	"logic",
+	"mathematics"
```
FridaTveit: The indentation for `integer` and the closing square bracket is now correct :) But the other topics should be in  line with `integer` which they're not at the moment. Maybe opening the file in Intellij or Sublime text would help? I don't use Notepad++ much so I don't know much about it, but it does seem to be doing something strange to your spacing :P

the7thobserver:
Ack, even in Intellij it looks fine. Maybe it's a Windows thing :/

Updated the config, and it looks fine in the "Changed Files" tab. Hopefully it's good :D

config.json
```diff
@@ -1033,7 +1033,12 @@
       "core": false,
       "difficulty": 9,
       "slug": "pythagorean-triplet",
-      "topics": null,
+      "topics": [
+       "integer",
+	"lists",
+	"logic",
+	"mathematics"
```
FridaTveit: I'm afraid the indentation is still not quite right. `lists`, `logic` and `mathematics` all need to be indented two spaces more than `topics`, they currently appear to be indented quite a few spaces less than `topics`. `integer` needs to be indented two spaces in from `topics`, it's currently indented one space in from `topics`. You say it looked fine in the "Changed files" tab. Did you look at all changes? It doesn't look fine when I look at it...

Sorry to be a bit picky about this but it is important to get the spacing right, otherwise the file becomes harder to read and maintain :)

Strange, I'm using Intellij on Windows and I haven't seen any issues where spacing appears correct in Intellij but then isn't when pushed to Github. Any idea what's going on @exercism/java?

stkent: No idea unless it's a tabs vs spaces thing. We can merge as-is and it will be fixed automatically the next time we (the maintainers) run `configlet fmt`.

FridaTveit: Sure :)

PR #970, opened 2017-10-23
Normalize the config file to minimize noise later

kytrinyx:
We will be scripting a number of changes to the track configuration.
This runs configlet fmt, which ensures that any further changes can be minimized.

See https://github.com/exercism/meta/issues/95

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,9 +28,9 @@
       "difficulty": 2,
       "slug": "rna-transcription",
       "topics": [
-        "strings",
         "loops",
-        "maps"
+        "maps",
+        "strings"
       ],
       "unlocked_by": null,
       "uuid": "8e983ed2-62f7-439a-a144-fb8fdbdf4d30"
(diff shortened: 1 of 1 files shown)


stkent:
Merging, though please note we have a number of "add topics" issues pending that are likely to (re-)introduce inconsistencies (as you know, per https://github.com/exercism/configlet/pull/78). Until there's a way to check whether format is up-to-date per `fmt` in CI, it might be safest to execute `fmt` as part of the scripted change itself.

bitmoji

kytrinyx:
There's some discussion in exercism/meta#95 about how this normalization degrades the readability of the configs.
Unfortunately, this means that I'm going to have to submit yet another normalization once that discussion concludes and we update the serializer.

I'm sorry about the noise, and any extra work that this is causing you!

Meanwhile, if you have any opinions on the topic, I'd appreciate it if you added your thoughts in exercism/meta#95

PR #971, opened 2017-10-23
Add topics for OCR Numbers exercise

### LINKED ISSUE #921: ocr-numbers: add topics

robvanderleek:
A contribution in the spirit of Hacktoberfest 2017. Hope these are the correct topics, I've touched those programming areas when doing the exercise myself.
Fixes #921

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -922,7 +922,6 @@
         "loops",
         "lists",
         "strings",
-        "algorithms",
         "parsing"
       ],
       "unlocked_by": "robot-name",
--- a/config.json
+++ b/config.json
@@ -917,7 +917,14 @@
       "core": false,
       "difficulty": 8,
       "slug": "ocr-numbers",
-      "topics": null,
+      "topics": [
+        "exception_handling",
+        "loops",
+        "lists",
+        "strings",
+        "algorithms",
+        "parsing"
+      ],
       "unlocked_by": "robot-name",
       "uuid": "5cbc6a67-3a53-4aad-ae68-ff469525437f"
     },


config.json
```diff
@@ -917,7 +917,14 @@
...
       "difficulty": 8,
       "slug": "ocr-numbers",
-      "topics": null,
+      "topics": [
+        "exception_handling",
+        "loops",
+        "lists",
+        "strings",
+        "algorithms",
+        "parsing"
```
FridaTveit: Great suggestions! :) The only one I'm not sure about is `algorithms`. You could argue that you use an algorithm to solve every exercise but I don't think it would be very useful to add it as a topic to every exercise. Therefore we've been trying only to add it to exercises that require the use of a specific algorithm, e.g. `binary-search`. So I might be tempted to leave that topic out. What do you think? :)

robvanderleek: Totally agree :-) I puzzled a bit on how to solve this exercise. Eventually ended up using simple string comparisons, but I wondered if others would use a solution based on `pattern_matching` or perhaps `pattern_recognition`. Anyway, I've limited the list to 5 topics that are definitely relevant for this exercise.

stkent: @robvanderleek in case you are interested in the way I solved this exercise in the reference solution: https://github.com/exercism/java/tree/master/exercises/ocr-numbers/.meta/src/reference/java. Thanks for the PR!

robvanderleek: Thanks @stkent, glad to contribute! This is my solution: http://exercism.io/submissions/778fc33721c948908cd7b7ba402c7292 pretty similar 😄

stkent: Nice! 💯

PR #973, opened 2017-11-03
Add additional luhn test for fuller coverage as in #972

### LINKED ISSUE #972: luhn: Add test enforcing that non-digit characters are invalid
I discovered that my solution to luhn (http://exercism.io/submissions/05a450216e9d4fc5965831a58760efe8) passed all the tests even if I didn't check that all digits excluding spaces were non-numeric. There are two reasons for this:

- I used `Character.getNumericValue()` to parse the string, and that function still returns numeric output for non-digits
- The tests checking that symbols and punctuation are invalid passed simply because the mathematical calculation using the numeric output from `Character.getNumericValue()` produces values that are not multiples of 10. 

I propose adding the following test to require checking if digits are non-numeric:

```
public void testThatStringContainingSymbolsIsInvalid2() {
     assertFalse(luhnValidator.isValid("34&"));
}
```
Without the check, the string `"34&"` evaluates to 10 since `Character.getNumericValue('&')` = -1 and so `isValid()` will return true, failing the test.

sjwarner-bp:
Add additional luhn test for fuller coverage in the Java stream. This fixes #972.

---

Reviewer Resources:

### DIFF
--- a/exercises/luhn/src/test/java/LuhnValidatorTest.java
+++ b/exercises/luhn/src/test/java/LuhnValidatorTest.java
@@ -66,15 +66,14 @@ public class LuhnValidatorTest {
         assertFalse(luhnValidator.isValid("055-444-285"));
     }
 
+    /* The following test diverges from the canonical test data. This is because the corresponding canonical test does
+     * not account for Java specific functions (such as Character.getNumericValue()), which can be part of incorrect yet
+     * passing implementations. For more detail, check out issue #972 here:
+     * (http://exercism.io/submissions/05a450216e9d4fc5965831a58760efe8).
+    */
     @Ignore("Remove to run test")
     @Test
-    public void testThatStringContainingSymbolsIsInvalidV1() {
-        assertFalse(luhnValidator.isValid("055£ 444$ 285"));
-    }
-
-    @Ignore("Remove to run test")
-    @Test
-    public void testThatStringContainingSymbolsIsInvalidV2() { assertFalse(luhnValidator.isValid("34&")); }
+    public void testThatStringContainingSymbolsIsInvalid() { assertFalse(luhnValidator.isValid("34&")); }
 
     @Ignore("Remove to run test")
     @Test
--- a/exercises/luhn/src/test/java/LuhnValidatorTest.java
+++ b/exercises/luhn/src/test/java/LuhnValidatorTest.java
@@ -73,7 +73,9 @@ public class LuhnValidatorTest {
     */
     @Ignore("Remove to run test")
     @Test
-    public void testThatStringContainingSymbolsIsInvalid() { assertFalse(luhnValidator.isValid("34&")); }
+    public void testThatStringContainingSymbolsIsInvalid() {
+        assertFalse(luhnValidator.isValid("34&"));
+    }
 
     @Ignore("Remove to run test")
     @Test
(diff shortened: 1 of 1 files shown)


stkent:
Thanks for this fix!

This issue makes me muse on a few broader questions also:

- is `testThatStringContainingSymbolsIsInvalidV1` still providing value? If so, can we articulate more clearly (via the test names) what the difference between `testThatStringContainingSymbolsIsInvalidV1` and `testThatStringContainingSymbolsIsInvalidV2` are?
- in the case where it makes sense for the Java track to deviate from canonical data (as it does here), how do we communicate that to future-us and future-maintainers?
    1. comment in the suite (which is then visible to users)?
    2. comment in the version file (which is then invisible to users but less visible to maintainers)?
    3. some other option I have not thought of.
- should the method of communication be the same for extra tests vs omitted tests vs modified tests?

This PR feels like the right time to address these issues, I think? cc @exercism/java for input too :)

sjwarner:
I completely agree. I'd suggest that the first test actually doesn't provide value now, and should probably be omitted.

Additionally, I think that a comment in the test suite would make sense. It would be visible, but probably not distracting to users. 

Do you think this sounds sensible?

rstockbridge:
Good ideas @stkent! I agree with @sjwarner. 

I think it may actually be interesting for the (non-beginner) user to see the comment in the test suite as it may prompt interesting thinking about variations among languages. It was definitely interesting to me to think about why my observation shouldn't apply to all language tracks.

stkent:
I'm a 👍 to both suggestions (_replacing_ the test in the Java track, and leaving a comment in the file for future maintainers)!

exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -66,11 +66,14 @@ public void testThatStringContainingPunctuationIsInvalid() {
         assertFalse(luhnValidator.isValid("055-444-285"));
     }
 
+    /* The following test diverges from the canonical test data. This is because the corresponding canonical test does
+     * not account for Java specific functions (such as Character.getNumericValue()), which can be part of incorrect yet
+     * passing implementations. For more detail, check out issue #972 here:
+     * (http://exercism.io/submissions/05a450216e9d4fc5965831a58760efe8).
```
FridaTveit: Is this definitely the right link? Do you not want to link to the issue: https://github.com/exercism/java/issues/972?

sjwarner-bp:
And once again, I've been struck with a case of clipboard-blindness... :scream: 

You are correct - I'll change that asap!

exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -66,11 +66,14 @@ public void testThatStringContainingPunctuationIsInvalid() {
...
+     * not account for Java specific functions (such as Character.getNumericValue()), which can be part of incorrect yet
+     * passing implementations. For more detail, check out issue #972 here:
+     * (https://github.com/exercism/java/issues/972).
+    */
     @Ignore("Remove to run test")
     @Test
-    public void testThatStringContainingSymbolsIsInvalid() {
-        assertFalse(luhnValidator.isValid("055£ 444$ 285"));
-    }
+    public void testThatStringContainingSymbolsIsInvalid() { assertFalse(luhnValidator.isValid("34&")); }
```
FridaTveit: Sorry, I just noticed that this is all on one line. I would put the method body on its own line. It's not very important but it does make it a bit more readable and it's more consistent with the other tests :)

sjwarner-bp:
Done :slightly_smiling_face:

PR #974, opened 2017-11-03
Add topics to House as per #793.

### LINKED ISSUE #793: house: add topics

sjwarner-bp:
Add topics to the House exercise in config.json, for the future version of exercism.

This fixes #793 .

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -604,8 +604,11 @@
       "slug": "house",
       "topics": [
         "arrays",
+        "conditionals",
         "loops",
-        "strings"
+        "strings",
+        "text_formatting",
+        "variables"
       ],
       "unlocked_by": "two-fer",
       "uuid": "6b51aca3-3451-4a64-ad7b-00b151d7548a"
@@ -616,11 +619,8 @@
       "slug": "food-chain",
       "topics": [
         "arrays",
-        "conditionals",
         "lists",
-        "strings",
-        "text_formatting",
-        "variables"
+        "strings"
       ],
       "unlocked_by": "two-fer",
       "uuid": "dd3e6fd6-5359-4978-acd7-c39374cead4d"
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
I've also noticed a couple more open issues regarding topics. While they currently seem to be assigned, should the contributors turn out to be inactive, I'd be happy to do those too :slightly_smiling_face:

config.json
```diff
@@ -602,7 +602,11 @@
       "core": false,
       "difficulty": 6,
       "slug": "house",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "loops",
+        "strings"
```
stkent: Maybe also `conditionals`?

stkent:
This commit seems to include `food-chain` changes too? Not sure if those are meant to be in here 🙊

sjwarner:
Oops, you're right! I seem to have added those three extra topics to `food-chain` - I'll move them all over to `house`! Not sure how I managed that...

config.json
```diff
@@ -602,7 +602,14 @@
...
       "difficulty": 6,
       "slug": "house",
-      "topics": null,
+      "topics": [
+        "arrays",
+        "conditionals",
+        "loops",
+        "strings",
+        "text_formatting",
+        "variables"
```
FridaTveit: Great suggestions :) The only one I'm not sure about is `variables`. How do you think that relates to this exercise? You'd have to use variables to solve pretty much every exercise so I would only add it as a topic if using variables was in some way particularly important to the exercise. What do you think? :)

sjwarner-bp:
You're right - I thought it was a bit odd when I looked through the list of all tags, but stuck it in anyway! I agree that it probably should be reserved for exercises where variables are important :slightly_smiling_face:

PR #985, opened 2017-11-09
Protein Translation: Implement problem in Java language stream.

sjwarner-bp:
After adding the `canonical-data.json` to problem-specifications [here](https://github.com/exercism/problem-specifications/pull/997), I wanted to implement the problem in Java. I assume that the fact there was no canonical data was the only reason this problem is yet to be implemented.

Please bear in mind that the `canonical-data.json` has not yet been merged, though it is approved. Hopefully this should not impact this PR.

---

Reviewer Resources:


exercises/protein-translation/src/test/java/ProteinTranslatorTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Before
+    public void setUp() {
+        proteinTranslator = new ProteinTranslator();
+    }
+
+    @Test
+    public void testMethionineRnaSequence() {
+        List<String> expected = Arrays.asList("Methionine");
+        assertEquals( expected , proteinTranslator.translate("AUG"));
```
FridaTveit: There shouldn't be any whitespace before and after expected :)

exercises/protein-translation/.meta/src/reference/java/ProteinTranslator.java
```diff
@@ -0,0 +1,47 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class ProteinTranslator {
```
FridaTveit: This could be package private :)

sjwarner-bp:
That's great! I'm not sure why I blanked on the `Contributing.md` and the `POLICIES.md` - I do normally check them out :scream: - I'll get on those changes soon, thank you for reviewing!

exercises/protein-translation/src/main/java/ProteinTranslator.java
```diff
@@ -0,0 +1,8 @@
+import java.util.List;
+
+class ProteinTranslator {
+
+    List<String> translate (String rnaSequence) {
```
FridaTveit: There shouldn't be a space after translate :)

exercises/protein-translation/.meta/src/reference/java/ProteinTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final String[] METHIONINE = {"AUG"};
+    private static final String[] PHENYLALANINE = {"UUU", "UUC"};
+    private static final String[] LEUCINE = {"UUA", "UUG"};
+    private static final String[] SERINE = {"UCU", "UCC", "UCA", "UCG"};
+    private static final String[] TYROSINE = {"UAU", "UAC"};
+    private static final String[] CYSTEINE = {"UGU", "UGC"};
+    private static final String[] TRYPTOPHAN = {"UGG"};
+    private static final String[] STOP = {"UAA", "UAG", "UGA"};
+
+    List<String> translate (String rnaSequence) {
```
FridaTveit: There shouldn't be a space after translate :)

exercises/protein-translation/.meta/src/reference/java/ProteinTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+
+        List proteinList = new ArrayList();
```
FridaTveit: This should be `List<String> proteinList = new ArrayList<String>()`. Otherwise you're using [a list of generic `Object`s](https://www.ntu.edu.sg/home/ehchua/programming/java/JavaGeneric.html) which isn't [typesafe](https://stackoverflow.com/questions/260626/what-is-type-safe). Does that make sense? :)

sjwarner-bp:
Oops! Spaces are always a silly thing for me, especially without using linters! The typesafety thing makes sense too - I was reading about it not too long ago :slightly_smiling_face:

FridaTveit:
Not sure why the build failed. I've restarted it. Sometimes it just fails for no reason and then runs fine when we restart it so we'll see :P

FridaTveit:
The build is fine now so it doesn't look like it was because of anything you did :)

sjwarner-bp:
Perfect, thank you :smile:

PR #995, opened 2017-11-10
bowling: improve topics

FridaTveit:
Removed `functions` as I don't think it's a very suitable topic. Added `exception_handling` since the tests expect an exception to be thrown.

---

Reviewer Resources:


stkent:
Great, thanks Frida!

PR #999, opened 2017-11-14
Add topics to problem 'binary-search' #938

arshanq:
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -711,7 +711,9 @@
       "slug": "binary-search",
       "topics": [
         "recursion",
-        "arrays"
+        "arrays",
+        "searching",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -709,7 +709,10 @@
       "core": true,
       "difficulty": 6,
       "slug": "binary-search",
-      "topics": null,
+      "topics": [
+        "recursion",
+        "arrays"
```
stkent: Good choices! I'd recommend adding `searching` here too, but I think that's the only change needed!

FridaTveit: I would add `generics` as well :)

PR #1003, opened 2017-11-16
saddle-points: add starter implementation.

### LINKED ISSUE #992: saddle-points: add starter implementation
We've recently decided that all exercises with difficulty 4 or lower should have starter implementation (see issue #977 for discussion). The exercise `saddle-points` has difficulty 4 but no starter implementation, therefore starter implementation should be added. See the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations) for details of our policy on adding starter implementation.

sjwarner-bp:
Fixes #992

---

Reviewer Resources:

### DIFF
--- a/exercises/saddle-points/src/main/java/Matrix.java
+++ b/exercises/saddle-points/src/main/java/Matrix.java
@@ -3,7 +3,7 @@ import java.util.Set;
 
 class Matrix {
 
-    Matrix(List<List<Integer>> lists) {
+    Matrix(List<List<Integer>> values) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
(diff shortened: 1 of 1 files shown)


exercises/saddle-points/src/main/java/Matrix.java
```diff
@@ -0,0 +1,13 @@
+import java.util.List;
+import java.util.Set;
+
+class Matrix {
+
+    Matrix(List<List<Integer>> lists) {
```
FridaTveit: Maybe calling this parameter something like `values` or `matrixValues` might make it a bit more clear what it is? :)

sjwarner-bp: Sure thing!

PR #1020, opened 2017-11-22
kindergarten-garden: add starter implementation

### LINKED ISSUE #989: kindergarten-garden: add starter implementation
We've recently decided that all exercises with difficulty 4 or lower should have starter implementation (see issue #977 for discussion). The exercise `kindergarten-garden` has difficulty 4 but no starter implementation, therefore starter implementation should be added. See the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations) for details of our policy on adding starter implementation.

sjwarner-bp:
Fixes #989 

---

Reviewer Resources:

### DIFF
--- a/exercises/kindergarten-garden/src/main/java/KindergartenGarden.java
+++ b/exercises/kindergarten-garden/src/main/java/KindergartenGarden.java
@@ -1,4 +1,14 @@
+import java.util.List;
+
 class KindergartenGarden {
+
+    KindergartenGarden(String garden, String[] students) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    KindergartenGarden(String garden) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
 
     List<Plant> getPlantsOfStudent(String student) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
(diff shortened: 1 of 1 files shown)


exercises/kindergarten-garden/src/main/java/KindergartenGarden.java
```diff
@@ -0,0 +1,7 @@
+class KindergartenGarden {
+
+    List<Plant> getPlantsOfStudent(String student) {
```
FridaTveit: You're missing an import for `List` which is why the build failed :) You also need to add the two constructors used in the tests :)

sjwarner-bp: Ah, wasn't sure if we wanted the constructors or not! I'll get on it now :slightly_smiling_face:

FridaTveit: We want to add enough to the starter implementation so that the tests will compile without errors. If we didn't add constructors then the tests would have errors :) We want users to be able to run the tests straight away without having to deal with scary errors first :)

sjwarner-bp: Makes sense to me :smile:

PR #1021, opened 2017-11-22
collatz-conjecture: add start implementation.

### LINKED ISSUE #990: collatz-conjecture: add starter implementation
We've recently decided that all exercises with difficulty 4 or lower should have starter implementation (see issue #977 for discussion). The exercise `collatz-conjecture` has difficulty 4 but no starter implementation, therefore starter implementation should be added. See the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations) for details of our policy on adding starter implementation.

sjwarner-bp:
Fixes #990 

---

Reviewer Resources:

### DIFF
--- a/exercises/collatz-conjecture/src/main/java/CollatzCalculator.java
+++ b/exercises/collatz-conjecture/src/main/java/CollatzCalculator.java
@@ -1,6 +1,6 @@
 class CollatzCalculator {
 
-    int computeStepCount(final int start) {
+    int computeStepCount(int start) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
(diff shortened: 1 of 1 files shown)


exercises/collatz-conjecture/src/main/java/CollatzCalculator.java
```diff
@@ -0,0 +1,7 @@
+class CollatzCalculator {
+
+    int computeStepCount(final int start) {
```
FridaTveit: I would remove the `final` keyword here. It's fine to keep it in the reference solution but I think it might be a bit confusing to people who are new to Java :) What do you think?

sjwarner-bp: That makes sense to me - I'll get it done when I'm back at my computer!

PR #1037, opened 2017-11-25
Add exercise parallel-letter-frequency

matthewmorgan:
I will add this exercise to the track.

---

Reviewer Resources:


matthewmorgan:
Hmm.  I think I may have painted myself into a corner.  Turns out that implementing this is one thing, and creating a test suite that can verify that the user is employing some sort of concurrent processing is another.

Any recommendations on how to do that @exercism/java?

stkent:
Hi, @matthewmorgan! Good to see you :)

My first thought would be: how are other track verifying the parallel nature of the implementation? Is there an approach we can ~steal~ borrow from one of them?

matthewmorgan:
Hey, @stkent!  Hope you are well.

Only a few tracks have implemented this, and in fact I saw the Python implementation just got pulled because non-parallel solutions were not being identified by the test suite.

The only approach I've seen that makes sense to me is on the [Go track](https://github.com/exercism/go/tree/master/exercises/parallel-letter-frequency).  They provide a single-threaded implementation of a solution and then benchmark the provided 'concurrent' solution against it.    The other tracks (unless I missed something) seem to provide tests that even a sequential process would pass.

I think this exercise is worth doing, but I also feel like the nature of it is to demand a specific internal implementation.  It's not enough to produce the right outputs from the given inputs, but we're also prying in to the internals of the solution, which feels weird, though perhaps necessary.

And, benchmarking doesn't prove or disprove concurrency.  It _might_ be able show that a solution is slower than could be expected, on a machine with X cores, when compared to a single-threaded solution run on the same machine, I guess? 

I think Go's example may be the closest to useful, but I would be curious to hear your thoughts.

sjwarner-bp:
I would definitely agree with you @matthewmorgan, I think this exercise is absolutely worth implementing because of how different it is to the existing Java problems. Of course, with those valuable differences come some issues because we as a community aren't sure on the best ways of implementing them.

I'd be interested to see exactly what the figures are when benchmarking between the two different implementations (threaded and not threaded), and if the results are different enough then that could be a way to move forward.

Something I think could be relevant is our implementation of [sieve](https://github.com/exercism/java/tree/master/exercises/sieve). In the [`README.md`](https://github.com/exercism/java/blob/master/exercises/sieve/README.md), it explains that the implementation is not tested, and only the results are tested (i.e. the user is left with the responsibility that the sieve has been correctly implemented using the appropriate algorithm).

While we should obviously strive for the most complete testing strategy possible, if testing concurrency isn't strictly possible then I still see merit in this exercise, as prompts the user to use a certain implementation to solve a problem (and it is vastly different to existing exercises).

Of course, I'd be really interested in seeing what everyone else has to say, or if the [Go stream's](https://github.com/exercism/go/tree/master/exercises/parallel-letter-frequency) examples prove useful to us too! :slightly_smiling_face:

sjwarner-bp:
I've just noticed something else! It seems as though [the test suite for `bank-account`](https://github.com/exercism/java/blob/master/exercises/bank-account/src/test/java/BankAccountTest.java) also has some idea of concurrency - have you checked this out @matthewmorgan ? :slightly_smiling_face:

matthewmorgan:
@sjwarner-bp thanks for your input.  I think the `bank-account` tests have a somewhat different goal in mind-- they're looking at allowing multiple threads to update the state of the account.  Sort of, the inverse of what we want to do, I believe.

I'm still puzzling on this one...

stkent:
Nothing brilliant springs to mind. Benchmarking seems fragile at best. Perhaps

> While we should obviously strive for the most complete testing strategy possible, if testing concurrency isn't strictly possible then I still see merit in this exercise, as prompts the user to use a certain implementation to solve a problem (and it is vastly different to existing exercises).

is the right way to go here, with the real "checking" for this exercise happening (in theory) at the feedback stage*.

We could require solutions to accept an injected thread pool, and then benchmark the solution when the thread pool contains a single thread vs multiple threads? Forces quite a specific implementation, however.

*I am much more a maintainer than a mentor; I don't mean to pass the buck here.

stkent:
Ref.: http://www.baeldung.com/java-8-parallel-streams-custom-threadpool

matthewmorgan:
@stkent that's an interesting idea.  That might be worth pursuing, thank you.

matthewmorgan:
@stkent @sjwarner-bp I'm planning on pursuing the strategy of injecting a thread pool.  I think it's implementation-specific, as Stuart said, but I also think it may be the only way to capture the value of this exercise, and as Sam said, I think that promoting the concurrent approach has merit.

I'll submit something for review in the next few days.

matthewmorgan:
@exercism/java here you go.

There is no canonical test suite, so I used the tests on the C# track as the template.  I added a final test that simply compares the performance of a 10-thread solution with a 1-thread solution in an attempt to find out if the user's solution can take advantage.

Let me know if there's anything you'd like changed or if you have any questions.  Thanks!

exercises/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
```diff
@@ -0,0 +1,45 @@
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class ParallelLetterFrequency {
```
FridaTveit: This (and the public methods) could be package private :)

exercises/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static Map<Integer, Integer> map;
```
FridaTveit: Maybe calling it something like `letterFrequencyMap` would make it more clear what it is? :)

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
```diff
@@ -0,0 +1,216 @@
...
+    public void noTextsMeansNoLetters() {
+        String input = "";
+        Map<Integer, Integer> expectedOutput = new HashMap<Integer, Integer>();
+        ParallelLetterFrequency p = new ParallelLetterFrequency(input);
+
+        assertEquals(expectedOutput, p.letterCounts());
+    }
+
+    @Test
+    @Ignore
```
FridaTveit: This should be `Ignore("Remove to run test")` as explained in the [POLICY doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests) :)

config.json
```diff
@@ -23,6 +23,17 @@
       "unlocked_by": null,
       "uuid": "74515d45-565b-4be2-96c4-77e58efa9257"
     },
+    {
+      "core": false,
+      "difficulty": 6,
+      "slug": "parallel-letter-frequency",
+      "topics": [
+        "concurrency",
+        "type_conversion"
```
FridaTveit: Maybe also `maps`? And maybe `strings`? And why `type_conversion`?

matthewmorgan: @FridaTveit I'm happy to add those-- didn't know if I should just emphasize what made this different, or add more things and risk larding it up too much.

I put in `type_conversion` because of the way that `int`, `char`, and `Integer` are all use in the handling of the "letters".  But if you think it's not appropriate I'm happy to remove it, LMK.

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
stkent: `@Ignore(...)`s should also be placed above the `@Test` annotation, per the POLICIES doc!

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
```diff
@@ -0,0 +1,216 @@
...
+        Long tenStartTime = System.currentTimeMillis();
+        Map<Integer, Integer> actualCounts = tenThreadPool.submit(
+                () -> (new ParallelLetterFrequency(b.toString()).letterCounts())
+        ).get();
+        Long tenJobTime = System.currentTimeMillis() - tenStartTime;
+
+        System.out.println("Single threaded computation took: " + singleJobTime + " ms.");
+        System.out.println("Multi threaded computation took: " + tenJobTime + " ms.");
+
+        assertTrue(singleJobTime - tenJobTime > 0);
```
stkent: A non-parallelized implementation might well pass this test with non-trivial frequency. Would it make sense to either

1. run the computations many times for each pool, and average the results, or
2. assert something like `singleJobTime - 2 * tenJobTime > 0`?

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertEquals(new Integer(2), p.letterCounts().get((int) 'ü'));
+    }
+
+    @Test
+    @Ignore
+    public void tenThreadPoolIsFasterThanSingleThreadPool()
+            throws InterruptedException, ExecutionException {
+
+        ForkJoinPool singleThreadPool = new ForkJoinPool(1);
+        ForkJoinPool tenThreadPool = new ForkJoinPool(10);
```
stkent: I'm not super-familiar with `ForkJoinPool`. Does the mechanism we're using for testing place any implicit constraints on how solutions must be structured in order to leverage the multiple threads? If so, that might be something to call out in the hints file for this exercise.

stkent:
Couple comments in, this is looking great!

config.json
FridaTveit: We've been adding topics that users might encounter when implementing the exercise. Topics that make this exercise different are possibly most important but I think more common topics like `strings` and `maps` are important as well. Adding them means that users who particularly want to practice using maps can look for all exercises with that topic and then do them. That's how I'm imagining the topics could be used anyway, I might be wrong :)

I would only add `type_conversion` if a typical solution required explicitly casting from one type to another. And this exercise seems to only require you to count the frequency of each letter which you could do without any type conversions? But if you think type conversion would be part of a good solution then I'm happy for it to stay in :)

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
matthewmorgan: @stkent I had originally done it like (2)! The results can vary quite a bit, but I think 2x is safe.  It's usually been closer to 3x on my laptop.

stkent: Sweet; let's go with `2 *` as a first attempt then?

matthewmorgan: Yeah, um...on that: I did some testing this AM and was pretty consistently getting ~1.8x rather than two, which seems weird to me.  Not sure what's changed since the first pass, but I'm going to investigate a bit more.

stkent: 🤞

We could also drop to 1.5x?

matthewmorgan: @stkent 1.5 seems to be working :).  I also made a change where the number of threads defaults to the number of CPUs allocated to the JVM.  Otherwise I think the performance of multiple threads might actually be worse in this case.

exercises/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
matthewmorgan: Done!

config.json
matthewmorgan: You're right-- my solution uses that technique, but it wouldn't be required.  It _is_ used in the tests, but still, I don't think the user would need to use it.

exercises/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
```diff
@@ -3,17 +3,17 @@
...
 
 
-    public ParallelLetterFrequency(String letters) {
-        map = getMapFromLetters(letters.toLowerCase().replaceAll(NOT_A_LETTER, ""));
+    ParallelLetterFrequency(String letters) {
+        letterFrequencyMap = getMapFromLetters(letters.toLowerCase().replaceAll(NOT_A_LETTER, ""));
     }
 
-    private Map<Integer, Integer> getMapFromLetters(String letters) {
+    Map<Integer, Integer> getMapFromLetters(String letters) {
```
FridaTveit: This can still be `private` :)

exercises/parallel-letter-frequency/src/test/java/ParallelLetterFrequencyTest.java
```diff
@@ -151,13 +152,13 @@ public void allThreeAnthemsTogetherProduceCorrectCounts() {
         assertEquals(new Integer(2), p.letterCounts().get((int) 'ü'));
     }
 
+    @Ignore("Remove to run test")
     @Test
-    @Ignore
     public void tenThreadPoolIsFasterThanSingleThreadPool()
```
FridaTveit: This test failed when Travis ran it which failed the build :(

matthewmorgan: Thanks @FridaTveit I see, I wonder if that is an artifact of the CI environment..I am starting to feel like this test is too fragile.  Let me work further on it.

matthewmorgan:
@stkent @FridaTveit @sjwarner-bp 

I've taken a different approach.  This _may_ be what some of you were suggesting at the beginning, but here's the idea: 
1. Make a pool of threads of a custom class, so we have some visibility into whether they've been started.
2. Submit our solution to the pool.
3. Ensure all threads in the pool were at least started.  This should not happen with a single-threaded solution, and the test should fail.

I arbitrarily picked a 4-thread pool to ensure we had multiple threads.  It would be possible for a user on a single core machine (Travis?) to get only a single thread if we peg the number of threads to the JVM cores.  This would allow a single-thread solution to still pass, which we don't want.

This solution does away with benchmarking, which is clearly very brittle.

Let me know what you think, re @stkent's earlier comment about adding something to the docs about the required properties of the solution, I am happy to do that if this approach seems sound.

Thanks!

stkent:
@matthewmorgan ingenious, I like it!

matthewmorgan:
@stkent I have to give credit to a friend at work @dbraley!  The implementation was mine, but the concept follows suggestions from him, and of course all of you fine maintainers.

I'll noodle something for the docs.

matthewmorgan:
@exercism/java this is ready for a look

config.json
FridaTveit: I think these topics still need updating? :)

config.json
```diff
@@ -23,6 +23,17 @@
...
+      ],
+      "unlocked_by": null,
```
FridaTveit: It would be great it you could choose a core exercise for this to be unlocked by :) Normally we'd choose an easier core exercise with relatively similar topics. If you don't want to choose one that's also fine, we can open an issue for it :)

exercises/parallel-letter-frequency/README.md
```diff
@@ -0,0 +1,27 @@
+# Parallel Letter Frequency
+
+Count the frequency of letters in texts using parallel computation.
+
+Parallelism is about doing things in parallel that can also be done
+sequentially. A common example is counting the frequency of letters.
+Create a function that returns the total frequency of each letter in a
+list of texts and that employs parallelism.
+
+# Hints
```
FridaTveit: This should probably be added in a `hints.md` file in the .meta directory as well so that if the readme is regenerated it will generate this bit too :) (see the [contribution doc](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#adding-a-new-exercise) for details on how to generate the readme)

exercises/parallel-letter-frequency/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Single-threaded (non-concurrent) solutions can pass all tests [but the last.](https://www.youtube.com/watch?v=mJZZNHekEQw)  Your solution will be tested for concurrency by submitting it as a [Runnable](https://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html) to an [ExecutorService.](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) Your solution must leverage multiple [Threads](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html) to pass the final test.
+
+Java documentation on [parallel streams](https://docs.oracle.com/javase/tutorial/collections/streams/parallelism.html) may provide some help.
+# Running the tests
```
FridaTveit: Should there be an empty line before this heading?

config.json
matthewmorgan: Oops...I did that, honest!  I must have forgotten to commit the work.  Thanks.

config.json
matthewmorgan: I chose`nucleodtide-count`.  That's what I based my solution on to begin with.  LMK if you think that's too big a leap.

exercises/parallel-letter-frequency/README.md
matthewmorgan: @FridaTveit the docs are a little unclear to me.  

1. Do I simply add the `hints.md` at the `.meta` root?  This seems to be what was done in EG pig-latin.
2. I have changed the `README.md` to have a "Java Tips" section instead of a "Hints" section.  Also, I have left the text in `README.md`, as I was unable to get the configlet tool to work after a couple of attempts.

I assumed the answer to both of the was 'yes'.  If not, let me know and I may need some guidance on additional required changes.

matthewmorgan:
OK @FridaTveit I've made the changes as requested.  I didn't squash yet, happy to do it if you feel this is ready to go.

config.json
FridaTveit: It has to be unlocked by a core exercise and `nucleotide-count` isn't a core exercise, sorry. `nucleotide-count` is unlocked by the core exercise `hamming`, so that might be a good one for this to be unlocked by as well?

exercises/parallel-letter-frequency/README.md
FridaTveit: The answer to 1 is definitely yes :) For 2, I think that should be fine. I assume doing it manually is fine if you can't get `configlet` to work, @stkent?

config.json
matthewmorgan: Oh sure. Sorry about the misunderstanding!

exercises/parallel-letter-frequency/README.md
stkent: I think that's fine; the next time we run an all-track update it will get refreshed :)

matthewmorgan:
@FridaTveit Thanks for your help!

I hope people have fun with it.

PR #1044, opened 2017-11-27
reverse-string: add to track

bmkiefer:
Implemented reverse string because this track did not have it. Not sure if I missed anything or that you even want this exercise on this track. Wasn't sure if I added README.md correctly.

---

Reviewer Resources:


stkent:
This looks great; thanks! The only change I'd request is that we move this exercise from in front of `two-fer` to directly behind `two-fer` (because `two-fer` has some additional files that help introduce newcomers to TDD).

To answer some of your questions:

> Not sure if I missed anything

Not that I can see :)

> or that you even want this exercise on this track

I think this is a worthy addition! Sure, the quickest possible solution is, well, pretty quick, but it will teach newcomers about a couple standard library tools they may not otherwise discover for a while.

http://exercism.io/languages/java/todo is an up-to-date list of the exercises that the Java track (1) hasn't implemented yet, and (2) hasn't explicitly decided not to implement (such exercises are marked as 'forgone' in the `config.json` file at the root of the repository.)

> Wasn't sure if I added README.md correctly.

It looks like it's missing a couple bits. The steps to generate README's aren't well-documented right now, but they are:

- download [`configlet`](https://github.com/exercism/configlet/releases) and put it somewhere in your `PATH`;
- run `configlet generate . --only reverse-string` from the root of this repository.

stkent:
(I don't remember if that `configlet generate` call requires you to have a local copy of the `problem-specifications` repo somewhere; it seems likely upon reflection, but I forget the details. `configlet`'s usage output should clarify.)

bmkiefer:
Yeah, you do have to have a local copy in order to generate the readme correctly. It was unclear to me until I tried adding a problem on the ruby track.

Okay, I will happily put this behind two-fer.

bmkiefer:
Okay, I moved the exercise behind two-fer in the ciriculum and generated the README.md via configlet. This should be good to be merged.

Smarticles101:
I think this looks good, I don't see anything wrong with it now that you've addressed @stkent's comments. I had been debating if it should be `unlocked_by` something, but I think it's a fit choice being `unlocked_by` `null`. Looks to me like it's ready to merge

Smarticles101:
Thanks a ton for the addition @bmkiefer :)

bmkiefer:
Definitely, I enjoyed contributing it! I was also contemplating whether to lock it behind two-fer or not. I think it might be worth it to hiding it behind two-fer given @stkent's comments, but I went with leaving more early exercises available to the developer. Maybe it could be good first patch to change this in the config.json if thats the route we want to go?

Smarticles101:
I think it's just leaving it behind `two-fer` because of the simplicity of two-fer vs the possible complexity of reverse-string. In two-fer, a developer is simply asked to pass a provided name into a string and return that, returning a default if the name is `null`.  two-fer doesn't use any library tools or loops. reverse-string on the other hand, if a user is trying to solve it, I can see it being done across many users with the "reverse for loop" way requiring a bit more thinking. :)

bmkiefer:
Sounds good to me. Out of curiosity, what is the process for the problems actually being added to the platform? Is there a batch task that runs every morning against the config files or how does that work? (Given that I can't see the reverse-string exercise in java track at this moment)

stkent:
I believe deployment is a manual process. There's no immediate way to check last deployment time (at least, there wasn't when I last asked :))

stkent:
Based on @Smarticles101 and your comments about the _possible_ complexity of this exercise, I think moving it to be `unlocked_by` something else makes sense, and an issue would be appropriate for that task :)

Smarticles101:
@stkent sounds good to me, I can open an issue for that :)

PR #1066, opened 2017-12-01
linked-list: make classes and methods package private #1055

aryasaatvik:
---

Reviewer Resources:

### DIFF
--- a/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
+++ b/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
@@ -48,7 +48,7 @@ final class DoublyLinkedList<T> {
         private Element<T> prev;
         private Element<T> next;
 
-        Element(T value, Element<T> prev, Element<T> next) {
+        public Element(T value, Element<T> prev, Element<T> next) {
             this.value = value;
             this.prev = prev;
             this.next = next;
--- a/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
+++ b/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
@@ -1,7 +1,7 @@
-public final class DoublyLinkedList<T> {
+final class DoublyLinkedList<T> {
     private Element<T> head;
 
-    public void push(T value) {
+    void push(T value) {
         if (head == null) {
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
Looks good to me! Thanks @aryasaatvik

exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
```diff
@@ -43,12 +43,12 @@ public T shift() {
         return value;
     }
 
-    private static final class Element<T> {
```
sjwarner-bp: I think that this nested class can probably stay as `private`, unless it causes any problems! :slightly_smiling_face:

aryasaatvik: I understand. I have committed the requested changes

exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
```diff
@@ -48,7 +48,7 @@ T shift() {
         private Element<T> prev;
         private Element<T> next;
 
-        Element(T value, Element<T> prev, Element<T> next) {
+        public Element(T value, Element<T> prev, Element<T> next) {
```
sjwarner-bp: I think you meant to make this one `private`! :slightly_smiling_face:

sjwarner-bp: I think I must have explained myself badly! I think that the nested class declaration on line 46 can remain `private` as it was originally, and the constructor method on line 51 can be made package-private.

Does this make sense? :slightly_smiling_face:

aryasaatvik: Yes, is it because line 46 is declaring a generic type and 51 is creating a constructor for it?

sjwarner-bp: The class can remain private, as we prefer to [limit scope](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) , and this causes no problems.

The constructor method for the class can be reduced to package-private instead of public (as it was before your edits) as it isn't used outside of this package.

PR #1064, opened 2017-11-30
Fix journey test

FridaTveit:
Attempt to fix #1063. Includes temporary commit to check that it's working (changing something in settings.gradle to make sure the script doesn't try to run settings.gradle like an exercise)

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/xjava/blob/master/POLICIES.md#event-checklist)

PR #1068, opened 2017-12-02
Ignore build.gradle when CI runs tests against changeset

bmkiefer:
I was curious if the CI also tried to run tests against build.gradle similar to how settings.gradle was having tests run against it if it was in the change set. It was fixed in #1064 for settings.gradle as a reference.

---

Reviewer Resources:


bmkiefer:
Okay, that previous build verifies that `build.gradle` changes also cause a CI failure.

bmkiefer:
Okay two builds ago showed that when build.gradle is in the change set and the fix is present that the CI runs as normal. and the most recent reverts the changes to build.gradle. should be good to merge.

Issue #1072, opened 2017-12-05
armstrong-numbers: add to track


sjwarner-bp:
As in [problem specifications 1018](https://github.com/exercism/problem-specifications/pull/1018), a new exercise has been added to exercism - Armstrong Numbers!

sjwarner-bp:
Closed via #1073

PR #1073, opened 2017-12-05
armstrong-numbers: add to track

sjwarner-bp:
As per #1072 

---

Reviewer Resources:


exercises/armstrong-numbers/src/test/java/ArmstrongNumbersTest.java
```diff
@@ -0,0 +1,87 @@
...
+    public void setup() {
+        armstrongNumbers = new ArmstrongNumbers();
+    }
+
+    @Test
+    public void singleDigitsAreArmstrongNumbers() {
+        int input = 5;
+        boolean expected = true;
+
+        assertEquals(expected, armstrongNumbers.isArmstrongNumber(input));
```
FridaTveit: You can use `assertTrue` and `assertFalse` instead of `assertEquals` for these tests :)

config.json
```diff
@@ -33,6 +33,17 @@
...
     },
+    {
+      "core": false,
+      "difficulty": 2,
+      "slug": "armstrong-numbers",
+      "topics": [
+        "integers",
+        "mathematics"
+      ],
+      "unlocked_by": null,
```
FridaTveit: Maybe this should be unlocked by some other exercise? Or do you think it should be always accessible? I think either is fine :)

sjwarner-bp:
Don't worry about it @FridaTveit, I have been a little busy too! :slightly_smiling_face: 

I'll get on those changes in a moment - however because the exercise is quite an easy one, and one of the first arithmetic-based exercises that it should remain always accessible :slightly_smiling_face:

exercises/armstrong-numbers/src/main/java/ArmstrongNumbers.java
```diff
@@ -0,0 +1,9 @@
+class ArmstrongNumbers {
+
+	boolean isArmstrongNumber(int i) {
```
FridaTveit: Sorry, just one thing I missed last time; `i` is normally only used as a variable name for an index. Maybe `numberToCheck`, `possibleArmstrongNumber` or `inputNumber` or somthing might be more descriptive? :)

sjwarner-bp:
Sorry this took so long @FridaTveit - I have been studying for my OCA!

PR #1076, opened 2017-12-06
Extend linked list tests

rdavid1099:
Increase DoublyLinkedList functionality to be more verbose and versatile.
Updated the test suite with the following new methods:
- `count()` - returns an `int` with the number of nodes in the list
- `getValueAtIndex(int index)` - returns the data stored in the node at the given index
- Initialize list with data - functionality to initialize the list with an initial set of data; ex: `new DoublyLinkedList<String>("Hello", "World");` will create a new linked list with two linked nodes containing "Hello" and "World" respectively.
- `insert(T data, int index)` - Inserts a new node with the given data at the given index shifting all prior nodes back and other nodes forward.
- `find(T data)` - returns the index number of the given data. If multiple results, it returns the index of the first index. If it is not found, it returns `null`.
- `replace(int index, T data)` - replaces the node's data at the given index with the given data

Furthermore, I updated the final `testExample` with all the new functionality.

---

Reviewer Resources:

### DIFF
--- a/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
+++ b/exercises/linked-list/.meta/src/reference/java/DoublyLinkedList.java
@@ -1,57 +1,177 @@
-final class DoublyLinkedList<T> {
-    private Element<T> head;
+public class DoublyLinkedList<T> {
+    private Node<T> initialNode;
+    private int count;
 
-    void push(T value) {
-        if (head == null) {
-            head = new Element<>(value, null, null);
-            head.next = head;
-            head.prev = head;
-            return;
-        }
-
-        Element<T> oldTail = head.prev;
-        Element<T> tail = new Element<>(value, oldTail, head);
-        oldTail.next = tail;
-        head.prev = tail;
+    public DoublyLinkedList(T... initData) {
+        this.count = 0;
+        for(T data : initData) { push(data); }
     }
 
-    T pop() {
-        head = head.prev;
-        return shift();
+    public void push(T data) {
+        Node<T> newNode = new Node<>(data);
+        if (this.count == 0) {
+            this.initialNode = newNode;
+        } else {
+            Node<T> lastNode = getLastNode(this.initialNode);
+            makeConnection(lastNode, newNode);
+        }
+        this.count++;
     }
 
-    void unshift(T value) {
-        push(value);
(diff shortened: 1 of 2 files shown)


rdavid1099:
Not ready to be merged. Updating `.meta/src/reference/java/DoublyLinkedList.java` to pass TravisCI checks.

FridaTveit:
Thank you for taking the time to do this @rdavid1099 :) It looks like good code :)

However, this is adding a lot of scope to this exercise beyond what the README suggests. I'd worry that it would make the exercise too difficult and out of sync with the other tracks. What do you think @exercism/java?

sjwarner-bp:
I'd tend to agree with @FridaTveit - furthermore I think this might diverge from the current [problem specification](https://github.com/exercism/problem-specifications/tree/master/exercises/linked-list) too much.

Maybe this would be a good issue / discussion to have over there? It might be interesting to see what others think of the existing linked-list problem, and if there is scope to add another (harder) linked-list exercise later?

FridaTveit:
Good idea! :) If you're interested @rdavid1099 you could open an issue in the [problem specifications repo](https://github.com/exercism/problem-specifications) (where the description and test data for exercises are defined) to discuss whether the scope for linked-list should be extended. The linked-list exercise also doesn't have canonical data at the moment ([here](https://github.com/exercism/problem-specifications/blob/master/exercises/acronym/canonical-data.json) is an example of the canonical data for another exercise, acronym, in case you're not familiar with canonical data) so you could try implementing that if you want :)

rdavid1099:
Awesome. Thank you so much for the advice. I will check that out today.

My initial thought was that the functionality of the linked list created in the exercise was fun, but I was excited to make more functionality, kind of mimicking the functionality of Ruby arrays. I will definitely look into the problem specifications repo and the canonical data.

FridaTveit:
No problem :) Thanks for contributing! :)

I agree that the functionality you added is really interesting, so I would be great if it could be added to an exercise in some way through problem-specifications :)

FridaTveit:
@rdavid1099 do you mind if we close this since it sounds like there's a consensus for not adding these changes right now? :)

rdavid1099:
Thanks for all the advice. I'm cool with closing this PR.

PR #1077, opened 2017-12-06
proverb: add to track

rdavid1099:
Add proverb exercise to Java track.

Used the [Ruby version](http://exercism.io/exercises/ruby/proverb/readme) as inspiration.

Test suite covers base functionality covered in the canonical data.

---

Reviewer Resources:


exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,28 @@
+final class Proverb {
+    private String[] words;
+    private String initWord, qualifier, result = "";
+
+
+    public Proverb(String... words) {
+        this.words = words;
+    }
+
+    public void setQualifier(String qualifier) {
```
sjwarner-bp: This is an example of where the package-private access level could be used - we try to [limit scope](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices).

rdavid1099: Thanks for the feedback. I really appreciate it. I'm relatively new to Java and am looking at the docs right now. When you say limiting scope by using package-private access level, are you referring to removing the `public` ahead of the method names? For example `public void setQualifier(......)` would simply become `void setQualifier(.....)`. This means that the method (or whatever member) can only be accessed within its own package. If that's correct, I'll go ahead and push up those changes.

sjwarner-bp: That's exactly right - I think this is a good change, as nothing outside of the package requires access to those files, and generally it is best practise to limit scope 🙂

rdavid1099: Perfect. Thank you so much. Exercism.io is one of the greatest resources when learning a new language.

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,27 @@
+final class Proverb {
+    private String[] words;
+    private String initWord, qualifier, result = "";
```
FridaTveit: These should all be local variables (apart from `words`) in the `getProverb` method since they're only used in that method :)

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,27 @@
+final class Proverb {
```
FridaTveit: This class doesn't really seem to follow the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/proverb/canonical-data.json). It should just have one method, `recite`, which when given a list of words (you can give the list of words in the constructor if you want) outputs the proverb. It should not have, and should not need, a `setQualifier` method. Does that make sense? :)

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,27 @@
...
+    private String[] words;
+    private String initWord, qualifier, result = "";
+
+    Proverb(String... words) {
```
FridaTveit: I would use `String[]`. Using `String...` adds some unnecessary complixity to the exercise and I don't think it's really needed :)

exercises/proverb/src/main/java/Proverb.java
```diff
@@ -0,0 +1,7 @@
+class Proverb {
```
FridaTveit: This class needs to have a signature that matches what is expected in the tests, so it should have the signature for all the methods that are used in the tests and a constructor which matches the one used in the test (or no constructor if the one used in the tests takes no argument, since a constructor with no arguments is provided by default in Java) :)

exercises/proverb/src/test/java/ProverbTest.java
```diff
@@ -0,0 +1,84 @@
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class ProverbTest {
```
FridaTveit: These tests should exactly match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/proverb/canonical-data.json) which I don't think they do at the moment :)

exercises/proverb/.meta/src/reference/java/Proverb.java
FridaTveit: Just to clatify, you can name the method something else if you like, for example `getProverb`, but I think it's very important that it's only one method without any `setQualifier` method :)

rdavid1099:
Thanks for all of the constructive input @FridaTveit! I really appreciate all of the attention to detail you guys put into these exercises. I will work on these fixes later today, and following this, I hope to add more exercises to other tracks. 😄

exercises/proverb/src/test/java/ProverbTest.java
rdavid1099: I was taking a look at the canonical data, and I copied the tests I'm using from the [Ruby exercise](https://github.com/exercism/ruby/blob/master/exercises/proverb/proverb_test.rb) which includes the qualifier and diverts a bit from the canonical data. To be fair, it also looks like its the third version of the Proverb exercise for Ruby. Since this is the first version of this exercise for Java, I will only include the canonical tests and leave the qualifier and such for future iterations.

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,23 @@
+final class Proverb {
+    private String[] words;
+
+    Proverb(String[] words) {
+        this.words = words;
+    }
+
+    String recite() {
+        if (words.length < 1) return "";
+        String initWord = "",
```
FridaTveit: Could this just be set to `words[0]` instead of setting it to an empty string and then immediately checking if it's empty?

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,23 @@
...
+               result   = "";
```
FridaTveit: You could use a [`StringBuilder`](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html) instead of a string for `result`. A `StringBuilder` is more efficient when you keep appending to it in a loop :)

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,23 @@
...
+        for(int i = 0; i < words.length; i++) {
+            String word = words[i];
+            if (initWord.isEmpty()) {
+                initWord = word;
+            } else {
+                result += "the " + word + " was lost.\n";
+            }
+            if (i + 1 != words.length) result += "For want of a " + word + " ";
```
FridaTveit: This should have curly brackets around it, i.e.
```
if (i + 1 != words.length) {
  result += "For want of a " + word + " ";
}
```
It's considered best practice to always put curly brackets around if statements, even when they're only one line, to prevent errors if someone tried to add another line to it. I admit errors are less likely when the body of the if is on the same line as the if statement, but I still think it's worth using curly brackets just in case :)

exercises/proverb/src/main/java/Proverb.java
```diff
@@ -0,0 +1,12 @@
...
+    private String[] words;
```
FridaTveit: You don't need this in the starter implementation. You only need enough so that the tests will compile :)

exercises/proverb/src/main/java/Proverb.java
```diff
@@ -0,0 +1,12 @@
...
+
+    Proverb(String[] words) {
+        this.words = words;
```
FridaTveit: This should be replaced with `throw new UnsupportedOperationException("Delete this statement and write your own implementation.");`. We only want to provide the users with the signatures for the exercise, not any of the implementation :)

rdavid1099:
@FridaTveit Thanks again for all the feedback. Regarding the if-statement syntax, I'm used to Ruby syntax and single-line if statements, so I really appreciate the insight on best Java practices.

I applied all of the fixes and refactored a bit to cut down on all the if-statements. If there is anything else, please feel free to tell me 😃

exercises/proverb/.meta/src/reference/java/Proverb.java
```diff
@@ -0,0 +1,17 @@
+final class Proverb {
+    private String[] words;
+
+    Proverb(String[] words) {
+        this.words = words;
+    }
+
+    String recite() {
+        if (words.length < 1) return "";
```
FridaTveit: Please put curly brackets around the body of this if :)

rdavid1099: Awesome. Thank you so much. I wasn't sure about that if-statement since it was extremely simple, but I do agree it's much more readable and debuggable with curly braces.

sjwarner-bp:
This looks great - once the CI has passed I am happy to merge this in 😄

sjwarner-bp:
Thanks @rdavid1099 - that's a great exercise to have implemented now 🙂

Issue #1081, opened 2017-12-11
meetup: make classes and methods package private in reference solution and associated enums


sjwarner-bp:
All the public class and method [access modifiers](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) in the [meetup reference solution](https://github.com/exercism/java/blob/master/exercises/meetup/.meta/src/reference/java/) can be changed to package private (which is the default therefore the access modifier can be left out). For example, in Meetup.java, `public class Meetup {` should be changed to `class Meetup {`. The same principle can be applied to MeetupSchedule.java in both [reference](https://github.com/exercism/java/blob/master/exercises/meetup/.meta/src/reference/java/) and [starter](https://github.com/exercism/java/tree/master/exercises/meetup/src/main/java) solutions.

We do this since it's best practice to restrict access to a class and its methods as much as possible (see our [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) for more information).

c-thornton:
Suggest closing this issue as it was resolved in merged pull request #1082.

sjwarner-bp:
Thanks for spotting this @c-thornton - knew I was forgetting something!

It might be useful to know that if you put 'fixes' and then an issue number, it will auto-close the issue upon merge! Details [here](https://help.github.com/articles/closing-issues-using-keywords/) :slightly_smiling_face:

sjwarner-bp:
Closed as per #1082

c-thornton:
Awesome, thanks for the info @sjwarner-bp. I will keep that in mind for future issues I work on.

PR #1082, opened 2017-12-12
"#1081 Removed public class modifiers"

c-thornton:
Removed public class modifiers from the Meetup.java and MeetupSchedule.java src and .meta folders.

---

Reviewer Resources:

### DIFF
--- a/exercises/meetup/.meta/src/reference/java/Meetup.java
+++ b/exercises/meetup/.meta/src/reference/java/Meetup.java
@@ -2,7 +2,7 @@ import java.time.LocalDate;
 import java.time.DayOfWeek;
 
 
-public class Meetup {
+class Meetup {
     private final LocalDate startOfMonth;
 
     public Meetup(int monthOfYear, int year) {
--- a/exercises/meetup/.meta/src/reference/java/Meetup.java
+++ b/exercises/meetup/.meta/src/reference/java/Meetup.java
@@ -5,7 +5,7 @@ import java.time.DayOfWeek;
 class Meetup {
     private final LocalDate startOfMonth;
 
-    public Meetup(int monthOfYear, int year) {
+    package-private Meetup(int monthOfYear, int year) {
         startOfMonth = LocalDate.of(year, monthOfYear, 1);
     }
 
--- a/exercises/meetup/.meta/src/reference/java/Meetup.java
+++ b/exercises/meetup/.meta/src/reference/java/Meetup.java
@@ -5,7 +5,7 @@ import java.time.DayOfWeek;
 class Meetup {
     private final LocalDate startOfMonth;
 
-    private Meetup(int monthOfYear, int year) {
+    public Meetup(int monthOfYear, int year) {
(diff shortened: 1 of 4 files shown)


exercises/meetup/.meta/src/reference/java/Meetup.java
```diff
@@ -2,7 +2,7 @@
 import java.time.DayOfWeek;
 
 
-public class Meetup {
+class Meetup {
     private final LocalDate startOfMonth;
 
     public Meetup(int monthOfYear, int year) {
```
FridaTveit: This can also be package private :)

c-thornton: Sorry for missing that one @FridaTveit, thought I got them all. Thanks for the feedback on this. I will get that one modified in a couple hours when I am in front of my computer again and modify the PR.

sjwarner-bp: Perfect!

c-thornton:
Thanks @sjwarner-bp for the tip. I was wondering what I was doing incorrectly there. I should have read what @FridaTveit had written better. This is a learning experience for sure. The tests are running again now so hopefully it will pass this time and we should be good.

c-thornton:
Doh! I should have known that honestly. Thanks for the link, I will read up on that tonight. I will make the correct change shortly and push it to my fork. Thanks for the guidance @sjwarner-bp and for being patient with me.

sjwarner-bp:
That's no problem! Thanks for the speedy updates 👍

c-thornton:
So one thing I just thought about @sjwarner-bp was consistency with the style. Since the Meetup method will not have a private declaration should I go ahead and remove it from the cycleToPrev method as well so that the style is consistent within the Meetup class?

sjwarner-bp:
That's a good question! While changing the `private` methods to package-private wouldn't cause the build to fail, it is worth keeping in mind that we are trying to [minimise the accessibility of classes and members](http://jtechies.blogspot.com/2012/07/item-13-minimize-accessibility-of.html). 

Because these methods are only used in Meetup (i.e. nothing outside of their class calls them), we can keep them as private.

Remember - package-private is **more** accessible than `private`, and we want to keep accessibility to a (working) minimum 🙂

c-thornton:
Awesome, thanks for the link and the info. So I removed private from cycleToPrev and cycleToNext with the commit that I just pushed on my side. I added them back in a pushed again and the checks just finished up for this one and were succesful. Thanks for the extra guidance and patience while working with me on this pull request.

c-thornton:
Awesome, thanks for the link and the info. So I removed private from cycleToPrev and cycleToNext with the commit that I just pushed on my side. I added them back in a pushed again and the checks just finished up for this one and were succesful. Thanks for the extra guidance and patience while working with me on this pull request.

c-thornton:
Awesome! Thanks @sjwarner-bp. Now time to do some reading before tackling some other issues. Thanks for all the feedback.

Issue #1049, opened 2017-11-28
hamming: make classes and methods package private in reference and starter implementations


sjwarner-bp:
All the public class and method [access modifiers](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) in the [hamming reference solution](https://github.com/exercism/java/blob/master/exercises/hamming/.meta/src/reference/java/) and [hamming starter implementation](https://github.com/exercism/java/tree/master/exercises/hamming/src/main/java) can be changed to package private (which is the default therefore the access modifier can be left out). For example, in Hamming.java, `public class Hamming {` should be changed to `class Hamming {`.

We do this since it's best practice to restrict access to a class and its methods as much as possible (see our [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) for more information).

FridaTveit:
This has been fixed for the hamming reference solution in PR #1071. Thanks @t0dd :)

The hamming starter implementation still needs fixing :)

sjwarner-bp:
Is this now resolved (and can be closed) thanks to #1083 ?

FridaTveit:
No, the starter implementation still needs fixing :)

PR #1083, opened 2017-12-12
matrix: make classes and methods package private in reference solution. Resolved the issue

pratikpalashikar:
…ss modifier

I have made the class as package private and removed the public access modifier from the public methods as well. Removed the extra spaces as well to preserve the indentation.  PR #1083

Edited the class Hamming to be package private, retaining the private method as it is. PR #1049 

---

Reviewer Resources:

### DIFF
--- a/exercises/hamming/.meta/src/reference/java/Hamming.java
+++ b/exercises/hamming/.meta/src/reference/java/Hamming.java
@@ -9,7 +9,7 @@ class Hamming {
         return hammingDistance;
     }
 
-    int computeHammingDistance(String leftStrand, String rightStrand) {
+    private int computeHammingDistance(String leftStrand, String rightStrand) {
         if(leftStrand.length() != rightStrand.length()) {
             throw new IllegalArgumentException("leftStrand and rightStrand must be of equal length.");
         }
@@ -22,7 +22,7 @@ class Hamming {
         return distance;
     }
 
-    int hammingContributionAt(int index, String leftStrand, String rightStrand) {
+    private int hammingContributionAt(int index, String leftStrand, String rightStrand) {
       return leftStrand.charAt(index) != rightStrand.charAt(index) ? 1 : 0;
     }
 
--- a/exercises/hamming/.meta/src/reference/java/Hamming.java
+++ b/exercises/hamming/.meta/src/reference/java/Hamming.java
@@ -1,15 +1,15 @@
-public class Hamming {
+class Hamming {
     private final int hammingDistance;
 
-    public Hamming(String leftStrand, String rightStrand) {
+    Hamming(String leftStrand, String rightStrand) {
         hammingDistance = computeHammingDistance(leftStrand, rightStrand);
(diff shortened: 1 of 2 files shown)


exercises/matrix/.meta/src/reference/java/Matrix.java
```diff
@@ -1,11 +1,11 @@
 import java.util.regex.Pattern;
 
-public class Matrix {
+ class Matrix {
```
FridaTveit: I think all the lines you've changed you've indented one space too much? Try to keep the indentation the same as it was :)

exercises/matrix/.meta/src/reference/java/Matrix.java
```diff
@@ -1,11 +1,11 @@
...
     private int[][] matrix;
     private static Pattern spacePattern = Pattern.compile(" ");
     private static Pattern newlinePattern = Pattern.compile("\\n");
 
-    public Matrix(String matrixAsString) {
+     public Matrix(String matrixAsString) {
```
FridaTveit: This can also be package private :)

pratikpalashikar: @FridaTveit thank you for suggestions! I will take care of it.

exercises/matrix/.meta/src/reference/java/Matrix.java
pratikpalashikar: Sure !

pratikpalashikar:
@FridaTveit  I have integrated the changes, please let me know if any more changes are needed!

pratikpalashikar:
Hi @FridaTveit  I have added the changes for the PR #1049  also. Sorry for including the different PR changes in this one. I will take care of it next time.

exercises/hamming/.meta/src/reference/java/Hamming.java
```diff
@@ -1,15 +1,15 @@
...
         hammingDistance = computeHammingDistance(leftStrand, rightStrand);
     }
 
-    public int getHammingDistance() {
+    int getHammingDistance() {
         return hammingDistance;
     }
 
-    private int computeHammingDistance(String leftStrand, String rightStrand) {
+    int computeHammingDistance(String leftStrand, String rightStrand) {
```
FridaTveit: This (and the next method) should remain private. We want to be as restrictive as possible with access modifiers and private is more restrictive than package private, see https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html for an explanation :)

pratikpalashikar: @FridaTveit  Thank you for your suggestion. I am just starting to contribute so its really helpful. And I thought that we have to make all the things package private even the private methods. Thank you for the explanation!

pratikpalashikar:
@FridaTveit  I have committed the changes related to the hamming class and also updated the description.

Issue #1075, opened 2017-12-06
documentation: add hints for the first exercises without starter implementation


daveziegler55:
I like your idea for allergy code.
However, I can't find any main method so I am having some trouble understanding how it really works. 
Also I am having trouble running  a test for `$ gradle test` in my terminal. Perhaps I am missing something in my install?  Thank you.

sjwarner-bp:
Hi @daveziegler55 - the `allergies` exercise does not have a starter implementation with it in line with our [POLICIES.md](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations), as it is rated 4 or higher in terms of difficulty.

However, the enums used for this exercise are in `Allergen.java`. I imagine that you are getting this file confused with the implementation you are meant to create. If you haven't already, I would make another Java file in the `src/main/java/` directory called `Allergies.java`, and then create your implementation in this file!

Please let me know if this helps - there is also guidance available in the [Gitter room](https://gitter.im/exercism/java) if you ever need it :slightly_smiling_face:

stkent:
This raises a good point about the transition from {stubs provided} to {no stubs provided}. We could probably make that transition less abrupt. How would folks feel about adding some `hints` to the earliest* exercises without starter implementations?

*The tricky thing is that we'd need to do this for several exercises in order to cover the set of {exercises that might be the first exercise a user encounters} due to the nextercism branching structure.

sjwarner-bp:
I think that's exactly right - it would be a really good idea to put some hints in and ease users into the idea of making their own files.

This exercise is probably particularly confusing due to the fact that there is a file providing enums, but it isn't a starter implementation.

FridaTveit:
I agree that this is a really good idea :) But as @stkent mentioned, it will be really hard to keep track of which exercises to do this for. We could add hints for the first core exercise with difficulty 5? Or all exercises with difficulty 5? All exercises with difficulty 5 would probably be the most thorough while not being too hard to maintain. But it would require a lot of duplication...

sjwarner-bp:
There are currently 14 exercises with difficulty 5 - I agree with the point you are making regarding duplication @FridaTveit. 14 exercises is quite a lot, however (if I understand correctly regarding nextercism) it is quite possible that many users will not complete all 14, especially if they choose to specialise in a specific area. For this reason, I'd be apprehensive about a less thorough coverage.

Additionally, I agree that this will be easiest to maintain! :slightly_smiling_face:

stkent:
[...]

PR #1084, opened 2017-12-12
Add hints to exercises without starter implementation

FridaTveit:
As exercises with difficulty 5 are the first exercises without starter implementation, they should have a hints file explaining how to fix the compiler errors that occur when an exercise doesn't have any starter
implementation.

This should be detailed in the policies doc.

I thought I would add the file to only one exercise until it's been reviewed so that I don't have to make changes to lots of files :) After the contents of the file has been agreed I could add the other files, or we could open good first patch issues to add them.

Fixes issue #1075.

---

Reviewer Resources:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -171,7 +171,7 @@ You can do this by:
 
   2. Clone [the problem-specifications repository](https://github.com/exercism/problem-specifications).
 
-  3. Run `configlet generate . --only name_of_new_exercise` from the root of this repository.
+  3. Run `configlet generate . --only name_of_new_exercise --spec-path path_to_problem_specifications` from the root of this repository.
 
 * Check if there is canonical data for the exercise you're adding.
 This can be found at `https://github.com/exercism/problem-specifications/tree/master/exercises/EXERCISE-SLUG/canonical-data.json`.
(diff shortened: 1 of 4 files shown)


FridaTveit:
Do you think it's worth making good first patch issues for copying the file to the other difficulty 5 exercises? We only have one good first patch issue open so I think that could be a good idea. What do you think @sjwarner-bp?

sjwarner-bp:
That's a good idea @FridaTveit - I think that they'd make good first patches, and it will give us a few more issues for easier onboarding! 🙂

FridaTveit:
Great, I'll do that then :) And I think I need to regenerate the README for flatten-array now that I've added a hints file. Will do that too :) I do need to do that don't I @stkent?

Do you think that will make adding these files too difficult for a `good-first-patch` issue? Or will it be okay if we give very specific instructions for what they need to do to generate the readme? What do you think @stkent @sjwarner-bp?

sjwarner-bp:
I think that with clear enough instructions this should be an easy enough patch, and will introduce new contributors to a different set of tools (i.e. configlet) than they might otherwise be exposed to :slightly_smiling_face:

FridaTveit:
Cool, update this review with the fixed readme and as soon as this is merged I'll start making good first patch issues :)

sjwarner-bp:
Thanks for doing this @FridaTveit :slightly_smiling_face:

PR #1100, opened 2017-12-17
Fixes #1049

### LINKED ISSUE #1049: hamming: make classes and methods package private in reference and starter implementations

c-thornton:
Fixes the hamming.java class to make it package private.

---

Reviewer Resources:

### DIFF
--- a/exercises/hamming/src/main/java/Hamming.java
+++ b/exercises/hamming/src/main/java/Hamming.java
@@ -1,4 +1,4 @@
-public class Hamming {
+class Hamming {
 
     Hamming(String leftStrand, String rightStrand) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");


sjwarner-bp:
Thanks @c-thornton 🙂

c-thornton:
No problem @sjwarner-bp. I saw that last part in the thread for it and figured I would take care of it really quick.

PR #1105, opened 2017-12-21
parallel-letter-frequency: unlock by bank account

FridaTveit:
Both exercises have concurrency as a topic so it makes sense for them to be grouped together. This will also mean that parallel-letter-frequency won't be encountered until later in the track. See issue #1075.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -871,7 +871,7 @@
         "maps",
         "strings"
       ],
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
Thanks for getting on that so quickly @FridaTveit 🙂

PR #1112, opened 2017-12-27
remove .keep

sjwarner-bp:
Remove `.keep` from repo. It looks to have been mistakenly added with the `list-ops` exercise.

Please let me know if I've misunderstood, and there is an actual reason for this file :slightly_smiling_face: 

---

Reviewer Resources:


stkent:
I just ran said check in #1113 :)

sjwarner-bp:
I saw that one! Wasn't sure if it would work for the top level folder or not though 🙂

stkent:
Hah, no, it only looks in starter dirs :P

PR #1115, opened 2017-12-29
_template: restructure directory

sjwarner-bp:
The template directory seems to the wrong structure. I've updated it to be in line with the 'standard' layout.

I'm not sure if we want an example README.md in there too, maybe with instructions on how to generate a README? But as there currently isn't one, I've not added it. Let me know if you think that would be a good addition 🙂 

---

Reviewer Resources:


sjwarner-bp:
@FridaTveit No problem, I'll strip it out 🙂 

I was also wondering if there should be a blank `version` file put in `.meta` with regards to #1116

FridaTveit:
Hm, I'm not sure. Some exercises don't need a version file because they don't have canonical data yet. So it might be confusing for people implementing exercises which don't have canonical data. Then again, most exercises do have canonical data now I think, so maybe it would be fine? What do you think? :)

sjwarner-bp:
I didn't consider exercises that do not have `canonical-data` yet! This is a good point - maybe we should hold off on adding a `version` to the template 🙂

PR #1163, opened 2018-01-02
Simplify model answers #1119

### LINKED ISSUE #1119: exercises: Simplify model answers
1. `Accumulate` - If we expect end-users to be familiar with Java 8, then we can utilize `Stream` to simplify the model answer. Same goes for `Anagram#match`.

1. `Acronym` - `generateAcronym` uses `Pattern` and `Matcher` classes to split the strings in order to generate the acronym. We can simplify it by using `String.split("\W")`

1. `Anagram` - `canonicalize` converts the input `String` into a `char[]`. This step is unnecessary.

1. `Atbash` - `stripInvalidCharacters` can be simplified to be something like `input.replaceAll("\W", "")`

I can take a look at other answers and attempt to simplify them. Would love to work on this if possible! :)

Zhiyuan-Amos:
Fixes #1119. 

@FridaTveit thank you for your comments at https://github.com/exercism/java/issues/1119#issuecomment-354663182! :) I've done changes for 3 exercises in this PR because they are all pretty bite-sized and only involving small changes.

---

Reviewer Resources:

### DIFF
--- a/exercises/acronym/.meta/src/reference/java/Acronym.java
+++ b/exercises/acronym/.meta/src/reference/java/Acronym.java
@@ -1,5 +1,5 @@
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.util.Arrays;
+import java.util.stream.Collectors;
 
 final class Acronym {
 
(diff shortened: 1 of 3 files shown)


exercises/acronym/.meta/src/reference/java/Acronym.java
```diff
@@ -13,14 +13,12 @@ String get() {
...
-    private String generateAcronym(String phrase){
-        final Pattern BREAK_WORDS = Pattern.compile("[A-Z]+[a-z]*|[a-z]+");
-        final Matcher matcher = BREAK_WORDS.matcher(phrase);
-        final StringBuilder stringBuilder = new StringBuilder();
-        while (matcher.find()){
-            stringBuilder.append(matcher.group().charAt(0));
-        }
-        return stringBuilder.toString().toUpperCase();
+    private String generateAcronym(String phrase) {
+        return Arrays.stream(phrase.split("\\W"))
```
Zhiyuan-Amos: a non-word character consists of `a-z`, `A-Z`, `0-9` & `_`. This may potentially be buggy, since the existing code only considers `a-z` & `A-Z`. Maybe I should use `\p{Alpha}` instead.

Zhiyuan-Amos:
Noted, thank you! :)

I'm thinking of porting over an exercise `Alphametics` over, are all the information required over [here](https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md)? Cos I noticed there's a lot of documents everywhere, not too sure if I've missed anything out ><

sjwarner-bp:
Yes, pretty much all of the language-agnostic info will be there, however it is worth noting that the Java track has it's own [CONTRIBUTING](https://github.com/exercism/java/blob/master/CONTRIBUTING.md) and [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md) guides that should be followed too!

PR #1164, opened 2018-01-02
Simplify tests

Zhiyuan-Amos:
I've opened up an issue in [exercism/problem-specifications](https://github.com/exercism/problem-specifications/issues/1067) to discuss the validity of this PR.

---

Reviewer Resources:

### DIFF
--- a/exercises/anagram/src/test/java/AnagramTest.java
+++ b/exercises/anagram/src/test/java/AnagramTest.java
@@ -58,6 +58,14 @@ public class AnagramTest {
 
     @Ignore("Remove to run test")
     @Test
+    public void testCaseInsensitiveDetectAnagram() {
+        Anagram detector = new Anagram("OrChEsTra");
+        List<String> anagrams = detector.match(Arrays.asList("carthorse", "CarThorsE", "CARTHORSE"));
+        assertThat(anagrams, allOf(hasItem("carthorse"), hasItem("CarThorsE"), hasItem("CARTHORSE")));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
     public void testEliminateAnagramsWithSameChecksum() {
         Anagram detector = new Anagram("mass");
         assertTrue(detector.match(Collections.singletonList("last")).isEmpty());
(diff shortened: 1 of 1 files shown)


Zhiyuan-Amos:
Yup, I've raised an issue in the other repo, but I was quite brief in the issue description because I thought that it would be much more descriptive to have both the changes made to the code together with the commit messages, instead of just trying to elaborate everything in text.

How should I better go about this then? :)

sjwarner-bp:
Once the issue has been resolved in [problem-specifications](https://github.com/exercism/problem-specifications), we can see the outcome of that issue and then begin to apply changes (if any) here :slightly_smiling_face: 

That is unless the rest of @exercism/java have a better idea :slightly_smiling_face:

FridaTveit:
I agree with @sjwarner-bp :) Once the issue has been resolved in problem-specifications we can update our tests to match any changes made to the canonical data :)

FridaTveit:
So if it's okay with you @Zhiyuan-Amos we should probably close this PR for now? It can take a while before problem-specification issues get resolved since they impact many tracks and therefore changes impact many people. Feel free to reopen this PR or make a new one once the problem-specifications issue has been resolved and the canonical data updated :)

Zhiyuan-Amos:
Yes sure! Didn't know that it will take a while for problem-specification issues to be resolved. Thank you :)

exercises/anagram/src/test/java/AnagramTest.java
```diff
@@ -34,14 +34,6 @@ public void testDetectLongerAnagram() {
         assertThat(anagrams, hasItem("inlets"));
     }
 
-    @Ignore("Remove to run test")
-    @Test
-    public void testDetectMultipleAnagrams() {
```
Zhiyuan-Amos: Note to self: This commit should be deleted because there's a difference between both tests. `testDetectMultipleAnagrams()` actually test for 2 matching anagrams, while `testDetectMultipleAnagramsForLongerWord()` test for 3 matching anagrams. As such, we should rename the method names instead for clarity.

PR #1187, opened 2018-01-02
custom-set: update tests to 1.1.0

### LINKED ISSUE #1177: custom-set: update tests
The custom-set tests should be updated to exactly match the canonical data. Also a version file should be added to match the canonical data version.

sjwarner-bp:
Fixes #1177

I am slightly puzzled by all the different data types, as this doesn't seem to be mentioned in the canonical data, but I didn't want to make too big a change! If we think this should be changed, I'll update the PR 🙂 

---

Reviewer Resources:


FridaTveit:
We decided to use different data types to make the exercise less confusing. If only one datatype is used then the signature forces the user to implement it using generics but then generics is never needed because only one data type is used. It is mentioned in the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/custom-set/canonical-data.json) that this is okay:
> "Tests about mixed-type sets are not included because the ability",
    "to implement that will vary by language. If your language supports it",
    "and you want to implement mixed-type sets, feel free."

:)

sjwarner-bp:
Makes sense to me! 🙂

PR #1188, opened 2018-01-03
Implement alphametics exercise

Zhiyuan-Amos:
Hello, I've implemented the alphametics exercise. I hope I didn't miss out anything else! ><
---

Reviewer Resources:


sjwarner-bp:
You can also run tests locally before pushing back to the PR if you want to save yourself a little time :slightly_smiling_face:

Zhiyuan-Amos:
> You can also run tests locally before pushing back to the PR if you want to save yourself a little time

Thanks for the tip! :)

Also, I've removed `hints.md` and squashed the commits.

sjwarner-bp:
Actually, on a closer look can we add an `unlocked_by`? As we get closer to nextercism, it will be good to suggest an exercise to unlock this one.

What do you think this exercise could be unlocked by?

exercises/alphametics/.meta/src/reference/java/Alphametics.java
```diff
@@ -0,0 +1,125 @@
...
+                int digit = letterToDigit.get(word.charAt(i));
+                builder.append(digit);
+            }
+
+            return Long.parseLong(builder.toString());
+        }
+    }
+}
+
+class StringUtil {
```
FridaTveit: Indentation seems to have gone a bit wrong here :) Also could the two inner classes be private?

exercises/alphametics/src/test/java/AlphameticsTest.java
```diff
@@ -0,0 +1,129 @@
...
+        expected.put('B', 9);
+        expected.put('L', 0);
+
+        assertEquals(expected, new Alphametics("I + BB == ILL").solve());
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testUniqueValue() {
+        assertNull(new Alphametics("A == B").solve());
```
FridaTveit: In cases like this I would expect it to throw an exception instead of returning `null`. For error handling it's okay to diverge a bit from the canonical data (I believe they're trying to remove most error cases from the canonical data for this reason) since it can vary a lot from language to language how you'd want to handle invalid input. In my opinion it's better to throw an exception in Java than to return null since that forces to user to deal with the error instead of relying on them checking if the output is null :)

FridaTveit: For an example where we've used custom exceptions in cases like this see the [bank account tests](https://github.com/exercism/java/blob/master/exercises/bank-account/src/test/java/BankAccountTest.java). You could also use `IllegalArgumentException` if you would prefer not to use a custom one :)

exercises/alphametics/src/test/java/AlphameticsTest.java
```diff
@@ -0,0 +1,129 @@
...
...TROLL + TERRIFIES + THE + HORSES + OF + FIRE + THE + TROLL + RESTS + AT + THE + HOLE + OF + LOSSES + IT + IS + THERE + THAT + SHE + STORES + ROLES + OF + LEATHERS + AFTER + SHE + SATISFIES + HER + HATE + OFF + THOSE + FEARS + A + TASTE + RISES + AS + SHE + HEARS + THE + LEAST + FAR + HORSE + THOSE + FAST + HORSES + THAT + FIRST + HEAR + THE + TROLL + FLEE + OFF + TO + THE + FOREST + THE + HORSES + THAT + ALERTS + RAISE + THE + STARES + OF + THE + OTHERS + AS + THE + TROLL + ASSAILS + AT + THE + TOTAL + SHIFT + HER + TEETH + TEAR + HOOF + OFF + TORSO + AS + THE + LAST + HORSE + FORFEITS + ITS + LIFE + THE + FIRST + FATHERS + HEAR + OF + THE + HORRORS + THEIR + FEARS + THAT + THE + FIRES + FOR + THEIR + FEASTS + ARREST + AS + THE + FIRST + FATHERS + RESETTLE + THE + LAST + OF + THE + FIRE + HORSES + THE + LAST + TROLL + HARASSES + THE + FOREST + HEART + FREE + AT + LAST + OF + THE + LAST + TROLL + ALL + OFFER + THEIR + FIRE + HEAT + TO + THE + ASSISTERS + FAR + OFF + THE + TROLL + FASTS + ITS + LIFE + SHORTER + AS + STARS + RISE + THE + HORSES + REST + SAFE + AFTER + ALL + SHARE + HOT + FISH + AS + THEIR + AFFILIATES + TAILOR + A + ROOFS + FOR + THEIR + SAFE == FORTRESSES").solve());
```
FridaTveit: It would be great if you could put some line breaks in this line so it isn't too long (less than 120 characters is a common limit) :)

config.json
```diff
@@ -612,6 +612,18 @@
...
+    {
+      "core": false,
+      "difficulty": 6,
+      "slug": "alphametics",
+      "topics": [
+        "mathematics",
+        "logic",
+        "conditionals"
+      ],
+      "unlocked_by": null,
```
FridaTveit: If you'd like to suggest an exercise this could be unlocked by, that would be great :) It has to be a core exercise and we normally choose a core exercise which has similar topics. Maybe `secret-handshake` could be a good suggestion for this one? It's sort of to do with logic and unlocks some other exercises to do with logic: `robot-simulator`, `wordy`, and `forth`. If you don't want to fill this in now then that's fine, we can open a separate issue for it :)

In case you're not familiar with the exercism structure, users will first only have a few easy exercises plus the first core exercise available to them. Once they complete that core exercise the next core exercise will become available to them plus any exercises that the first core exercise unlocks :)

FridaTveit: Sorry @sjwarner-bp, I didn't see that you'd already put a comment about this! :P

sjwarner-bp: Don't worry about it, your feedback is a bit more detailed 😉

exercises/alphametics/.meta/src/reference/java/Alphametics.java
Zhiyuan-Amos: This isn't an inner class :P I've shifted the method inside `StringUtil` to simply be a private helper method in `Alphametics` instead.

exercises/alphametics/src/test/java/AlphameticsTest.java
Zhiyuan-Amos: Ah I see! ^^ Currently I see 3 options:
1. Throwing a checked Exception e.g. `BankAccountActionInvalidException`.
2. Throwing an unchecked Exception e.g. `IllegalArgumentException`.
3. Returning an `Optional<Map<Character, Integer>>`.

I don't quite like option 2 because the user isn't supposed to catch an unchecked Exception according to the [javadoc](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html).

Comparing between option 1 & 3, I prefer option 3 because not having a result doesn't seem to be that exceptional. Or perhaps, I guess `Optional` does the job well enough, so we don't have to write an additional `Exception` class.

```java
/**
 * Returns the solution to the alphametics puzzle if it exists.
 */
Optional<Map<Character, Integer>> solve() {
    AlphameticsRecursion solver = new AlphameticsRecursion(getDistinctCharacters());
    solver.generate();
    return solver.get();
}
```

Zhiyuan-Amos:
Thanks for your reviews! :) I've made the changes requested except for the one on exceptions.

exercises/alphametics/src/test/java/AlphameticsTest.java
FridaTveit: I'm not sure that would be the best solution. It's an error case, the user (in this case the tests) has provided invalid input, so I would expect it to throw an exception. It would also be more consistent with the other exercises and won't require the person implementing the exercise to learn about optionals. 

We aren't expecting the anything to catch the exception, so I think it's fine to use an unchecked exception. We only want the tests to confirm that an exception has been thrown (by using `ExpectedException`). I normally prefer using a custom exception because I think that's neater and makes it more clear what's going on since you can give it a relevant name :) In that case we would provide the implementation of the exception as starter implementation. Does that make sense? 

Those are just my preferences, feel free to disagree :)

sjwarner-bp: I'd agree - in most of the existing test suites that require this kind of thing, it isn't uncommon for us to use `ExpectedException`.

For consistency (and ease of the users) I probably prefer option 2? But I option 1 sounds good to me too!

FridaTveit:
Your new changes look great @Zhiyuan-Amos, thanks! Once we've agreed on an approach for error handling I think this is ready to merge :)

exercises/alphametics/src/test/java/AlphameticsTest.java
FridaTveit: Having thought about it a bit more, I do see your point @Zhiyuan-Amos that this isn't really a very exceptional case. However, I would still argue for using an exception because:
1. As @sjwarner-bp said, it would be more consistent with other exercises.
2. Exceptions are very well implemented in Java in my opinion, they're one of my favourite parts of the language actually :) Whereas optionals, while very useful in other languages, are a bit of a pain in Java. If we changed the solve method to return an optional we would have to change all the other tests to check whether it's present, or live with the compiler warnings telling us we haven't checked whether it's present.

However, since it's not really a case of an illegal argument, and it's something we would want the user to be able to recover from, you're probably right that it should be a checked exception. Let me know what you think @Zhiyuan-Amos :)

Zhiyuan-Amos: Yes, I agree that using an `Exception` instead is more consistent with other exercises. ^^ I've updated it accordingly. Thanks :)

Zhiyuan-Amos:
On a side note, I noticed that the issues I've raised in `problem-specifications` are not addressed by the developers there :( 

[issue 1](https://github.com/exercism/problem-specifications/issues/1067), [issue 2](https://github.com/exercism/problem-specifications/issues/1068).

Is there anything that I can do about it? :)

sjwarner-bp:
@Zhiyuan-Amos, it could be worth posting in [here](https://gitter.im/exercism/dev) to ask about that.

Sometimes problem-specifications moves slower than the language tracks, so I wouldn't worry too much as it is a busy time of year :slightly_smiling_face:

Zhiyuan-Amos:
Thank you for both of your reviews :)

PR #1191, opened 2018-01-03
Update TESTS.md

svenklaus:
On Mac OS X the command "$ gradle test" (row 72 in this doc) won't work. I propose to change it to "$ ./gradlew test" to execute the gradle wrapper with the 'test' argument. Linux should be the same.

---

Reviewer Resources:

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -69,7 +69,7 @@ If you get stuck, at any point, don't forget to reach out for [help](http://exer
 3. Run the tests:
 
     ```
-    $ gradle test
+    $ ./gradlew test
     ```
    *(Don't worry about the tests failing, at first, this is how you begin each exercise.)*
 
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
`$ gradle test` seems to work on my Mac.

gradletest

What makes you think it does not work on OSX?

svenklaus:
`$ gradle test`
`-bash: gradle: command not found`

I don't have the gradle command on my SYSTEM PATH as it's not necessary and it seems to me that this gradle setup uses the wrapper, so `gradle` is just a directory on my system. Can't be executed.

sjwarner-bp:
It looks like in [INSTALLATION.md](https://github.com/exercism/java/blob/master/docs/INSTALLATION.md#installing-on-mac-os-x-without-homebrew) Mac OSX users are instructed to add it to their `$PATH`.

We should probably decide to keep one or the other, instead of conflicting documentation :slightly_smiling_face:

svenklaus:
I agree. I did not read through "Installing Java" since I thought it was only about the JDK. Thanks for the quick reply.

PR #1211, opened 2018-01-08
all-your-base: update tests to version 2.1.0

### LINKED ISSUE #1160: all-your-base: update tests
The all-your-base tests should be updated to exactly match the canonical data. Also the version file should be updated to match the canonical data version.

FridaTveit:
Fixes #1160.
[Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/all-your-base/canonical-data.json).

I didn't add the last test because it's not possible in our implementation. Our tests expect an exception to be thrown when a `BaseConverter` object is created with an invalid base. Therefore it's not possible to provide both an invalid initial base and an invalid output base. I think it's fine to stick with what we've currently got, let me know if you disagree 🙂 

---

Reviewer Resources:


sjwarner-bp:
Thanks for that @FridaTveit :slightly_smiling_face:

PR #1244, opened 2018-01-22
Update README.md

fedavies:
Possible clarification of the task.

Simple clarification of the language in the space-age README to hopefully make the task clearer.

---

Reviewer Resources:

### DIFF
--- a/exercises/space-age/README.md
+++ b/exercises/space-age/README.md
@@ -1,6 +1,6 @@
 # Space Age
 
-Given an age in seconds, calculate how old someone would be on:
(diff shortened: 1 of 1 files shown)


FridaTveit:
I agree with @sjwarner-bp. The README on the Java track should be consistent with the one in problem-specifications. The only difference could be some added hints at the end that are java specific (e.g. "you might want to use exceptions to solve this exercise"). So issues and PRs like this should be raised in problem-specifications. Does that make sense @fedavies? :)

fedavies:
That makes perfect sense :)

FridaTveit:
Great! :) I'll close this PR for now then and feel free to open a PR with this change in problem-specifications 🙂

PR #1248, opened 2018-01-24
Nucleotide count tests and version #1142

### LINKED ISSUE #1142: nucleotide-count: update tests and add version file
The nucleotide-count tests should be updated to exactly match the canonical data. Also a version file should be added to match the canonical data version.

tshradheya:
Fixes #1142 

[Canonical Data
](https://github.com/exercism/problem-specifications/blob/master/exercises/transpose/canonical-data.json)

---

Reviewer Resources:


tshradheya:
Ready for Review

tshradheya:
Hi @sjwarner-bp , can't seem to find the comments :P

sjwarner-bp:
That would be because for some bizarre reason, they don't seem to have saved :laughing: 

I'll quickly re-add them :slightly_smiling_face:

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
```diff
@@ -35,68 +28,32 @@ public void testEmptyDnaStringHasNoNucleotides() {
...
-        ));
-    }
-
-    @Ignore("Remove to run test")
-    @Test
-    public void testValidatesNucleotides() {
-        expectedException.expect(IllegalArgumentException.class);
-        NucleotideCounter nucleotideCounter = new NucleotideCounter("GACT");
-        nucleotideCounter.count('X');
+        assertThat(nucleotideCounter.count('G'), is(7));
```
sjwarner-bp: Is there a reason you have used `assertThat(nucleotideCounter.count(...` instead of `assertThat(counts, allOf(hasEntry(...` (the structure of the other tests)?

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
```diff
@@ -106,4 +63,12 @@ public void testCountsAllNucleotides() {
                 hasEntry('T', 21)
         ));
     }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testDnaStringHasInvalidNucleotides() {
+        expectedException.expect(NullPointerException.class);
```
sjwarner-bp: Instead of leaving the program to throw a `NullPointerException`, what about using an `IllegalArgumentException` or something similar - it could even be a custom exception if you wanted!

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
```diff
@@ -13,18 +13,11 @@
...
-        NucleotideCounter nucleotideCounter = new NucleotideCounter("");
-        assertThat(nucleotideCounter.count('A'), is(0));
-    }
-
     @Ignore("Remove to run test")
     @Test
     public void testEmptyDnaStringHasNoNucleotides() {
         NucleotideCounter nucleotideCounter = new NucleotideCounter("");
         Map<Character, Integer> counts = nucleotideCounter.nucleotideCounts();
-        assertThat(counts.size(), is(4));
```
sjwarner-bp: This size test hasn't ever been part of the canonical data, but I am wondering if it was in here for a more Java-specific reason...

@exercism/java are we good to omit this? :slightly_smiling_face:

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
sjwarner-bp: That looks reasonable to me :slightly_smiling_face:

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
tshradheya: The size() of counts is always 4 due to the .emptyCount() function, so I preferred to remove it from most places, however testing it in some places

FridaTveit: If it's not part of the canonical data then I think we should remove it :)

exercises/nucleotide-count/.meta/src/reference/java/NucleotideCounter.java
```diff
@@ -27,6 +27,8 @@ private static boolean isCountable(char base) {
   Map<Character, Integer> nucleotideCounts() {
     Map<Character, Integer> counts = emptyCounts();
     for (char c : sequence.toCharArray()) {
+      if (isCountable(c))
```
FridaTveit: I think it's a bit confusing to have this check here. I would expect it to check that the input is valid in the constructor, so to loop though the sequence in the constructor and throw an exception if any of the characters aren't valid. What do you think?

Also please put curly brackets around the check:
```
if (isCountable(c)) {
  throw new IllegalArgumentException(c + " is not a nucleotide");
}
```
Otherwise it's easy to make mistakes, see https://stackoverflow.com/questions/8020228/is-it-ok-if-i-omit-curly-braces-in-java 🙂 

There's also an instance of this `if` on line twelve without curly braces. It would be great if you wouldn't mind changing that too, though of course you don't have to since you didn't add it 🙂

exercises/nucleotide-count/src/test/java/NucleotideCounterTest.java
```diff
@@ -13,18 +13,11 @@
     @Rule
     public ExpectedException expectedException = ExpectedException.none();
 
-    @Test
-    public void testEmptyDnaStringHasNoAdenine() {
-        NucleotideCounter nucleotideCounter = new NucleotideCounter("");
-        assertThat(nucleotideCounter.count('A'), is(0));
-    }
-
     @Ignore("Remove to run test")
```
FridaTveit: Since this is now the first test it shouldn't have an `@Ignore` statement above it 🙂

tshradheya:
Made the requested changes.

Ready for Review again

exercises/nucleotide-count/.meta/src/reference/java/NucleotideCounter.java
tshradheya: Agree. Working on it

tshradheya:
@FridaTveit You can review it now. Made necessary changes

tshradheya:
@FridaTveit Glad to contribute 👍

PR #1222, opened 2018-01-12
Added test cases as per #1161

TimoleonLatinopoulos:
Tried to fix #1161. [Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json)

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/xjava/blob/master/POLICIES.md#event-checklist)

PR #1249, opened 2018-01-27
Pig-Latin Exercise Updates

### LINKED ISSUE #1140: pig-latin: update tests and add version file
The pig-latin tests should be updated to exactly match the canonical data. Also a version file should be added to match the canonical data version.

Kyle-P:
Updated Pig-Latin tests file and added a version file. Could someone please double-check that the test class works as intended? Thanks!

Closes #1140 

---

Reviewer Resources:


exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -53,6 +58,15 @@
                 // Xr is treated like a single vowel
                 {"xray", "xrayay"},
 
+                // 'Y' is treated like a consonant at the beginning of a word
+                {"yellow", "ellowyay"},
+
+                // 'Y' is treated like a vowel at the end of a consonant cluster
+                {"rhythm", "ythmrhay"},
```
c-thornton: So this line is why the build failed. Below is what is showed on the Travis Tests.
PigLatinTranslatorTest > test[19: expected "rhythm" to translate to the pig latin phrase "ythmrhay"] FAILED
    org.junit.ComparisonFailure: expected:<[ythmrh]ay> but was:<[hythmr]ay>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at PigLatinTranslatorTest.test(PigLatinTranslatorTest.java:82)

Not sure why it failed though, because the the canonical-data lists exactly that as the test case. Maybe someone from @exercism can look into it more and find out why.

sjwarner: As said above - just because the test is the correct test, doesn't mean the build will pass!

We also need the reference solution store in `.meta` to pass all of the tests as well. As we can see here, all of the tests pass apart from this one (probably because this 'y' is an edge case where it will be treated as a vowel after a consonant).

The current version of the reference solution gives us `hythmray` instead of `ythmrhay`!

sjwarner-bp: As said above - just because the test is the correct test, doesn't mean the build will pass!

We also need the reference solution store in `.meta` to pass all of the tests as well. As we can see here, all of the tests pass apart from this one (probably because this 'y' is an edge case where it will be treated as a vowel after a consonant).

The current version of the reference solution gives us `hythmray` instead of `ythmrhay`!

c-thornton: Thanks @sjwarner-bp for the explanation, I was writing this comment at the same time you posted the one on the PR so I didn't see that one.

Kyle-P: Alright, thanks @sjwarner-bp and @c-thornton! I'll update it to "hythmray." How can I run the tests locally?

c-thornton: @Kyle-P I remember @sjwarner-bp made a comment a while ago about it. I don't know who to link to it directly but it was in PR #1222. Basically you need to install gradle and run the tests locally. You can find some more information about how to set it up in the contributing guide found here.

https://github.com/exercism/java/blob/master/CONTRIBUTING.md#advanced-complete-local-setup

Kyle-P: Thansk @c-thornton! I'd rather keep things done online for personal reasons, but I definitely see how it's better to run tests locally. Would setting up TravisCI on my fork be a plausible option?

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -53,6 +58,15 @@
                 // Xr is treated like a single vowel
                 {"xray", "xrayay"},
 
+                // 'Y' is treated like a consonant at the beginning of a word
+                {"yellow", "ellowyay"},
+
+                // 'Y' is treated like a vowel at the end of a consonant cluster
+                {"rhythm", "hythmray"},
```
FridaTveit: This should map to `"ythmrhay"`. If the build doesn't pass when you do that then you need to update the reference solution so that it maps it correctly. Does that make sense? 🙂

Kyle-P: Alright, thanks! Two questions:
 1) Is it feasible to set up the same TravisCI builds in my fork (I'd rather run the tests online than 
     locally)?
 2) Does updating the reference solution mean changing the canonical data?

Thank you!

sjwarner-bp:
I'd say while it's possible to set up similar build processes on a fork, it might not be worth it - the feedback cycles will be pretty similar, so if you want to test online then I'd recommend leaving it as is!

Updating the reference solution means to change the example stored in `.meta`. Just imagine you were completing the exercise on exercism!

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
```diff
@@ -21,11 +21,17 @@ String translate(String sentence) {
                 .collect(Collectors.joining(" "));
     }
 
+    private static int yLocation;
```
FridaTveit: Having a private variable which is set in one method and used in another like this makes the code a bit confusing to read. If you need to use this straight after having called `hasYAfterConsonantCluster` then you could just make the method return the yLocation (and rename it something like `getLocationOfYAfterConsonantCluster`)? And it can return -1 if there isn't one. What do you think? 🙂

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
```diff
@@ -21,11 +21,17 @@ String translate(String sentence) {
...
+
     private String translateWord(String word) {
         if (wordStartsWithVowelLike(word)) {
             return word + AY;
         }
 
+        if(hasYAfterConsonantCluster(word)){
```
FridaTveit: There should be white space after `if` and before `{`, like the other `if` statements already in this file 🙂

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
```diff
@@ -41,6 +47,18 @@ private String translateWord(String word) {
         return word.substring(1) + word.toCharArray()[0] + AY;
     }
 
+    private static boolean hasYAfterConsonantCluster(String word) {
+        for(int i = 0; i < word.length() - 1; i++){
```
FridaTveit: There should be white space after `for` and before `{`. And please try to keep the indentation consistent, everything should be indented with four spaces like the rest of the file 🙂

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
Kyle-P: Great idea. I'll take care of it. Thanks!

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
```diff
@@ -21,14 +21,12 @@ String translate(String sentence) {
...
 
-    private static int yLocation;
-
     private String translateWord(String word) {
         if (wordStartsWithVowelLike(word)) {
             return word + AY;
         }
 
-        if(hasYAfterConsonantCluster(word)){
+        if (getLocationOfYAfterConsonantCluster(word) != -1) {
```
FridaTveit: You need to store the result of this method call in a variable and use it inside the `if` statement. Right now the build is failing because you're using the variable `yLocation` which doesn't exist anymore 🙂

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
```diff
@@ -47,16 +45,15 @@ private String translateWord(String word) {
...
-            if(word.substring(i, i + 1).matches(VOWELS_REGEX)){
-                return false;
-            }else if((i != 0) && word.substring(i, i + 1).equals("y")){
-            	yLocation = i;
-                return true;
+    private static int getLocationOfYAfterConsonantCluster(String word) {
+        for (int i = 0; i < word.length() - 1; i++) {
+            if (word.substring(i, i + 1).matches(VOWELS_REGEX)) {
+                return -1;
+            }else if ((i != 0) && word.substring(i, i + 1).equals("y")) {
```
FridaTveit: There should be a white space before `else` 🙂 Also you could use `word.charAt(i)` in both the `if` and `else` statements (e.g.  `word.charAt(i) == 'y'`) instead of using `substring`, that might make it a bit more obvious what you're doing 🙂

exercises/pig-latin/.meta/src/reference/java/PigLatinTranslator.java
Kyle-P: Oops. Thanks for catching that one.

PR #1274, opened 2018-01-31
Implement new exercise rational-numbers

clementi:
---

Reviewer Resources:


exercises/rational-numbers/src/test/java/RationalTest.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    private void assertRationalsEqual(Rational r1, Rational r2) {
+        assertEquals(r1.getNumerator(), r2.getNumerator());
+        assertEquals(r1.getDenominator(), r2.getDenominator());
+    }
+
+    // Tests
+
+    @Test
```
sjwarner-bp: As per our [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests), only the first test should be used by default. All other tests should be `@Ignore`, in order to allow users to build up a solution.

This occurs a few times here, but I won't leave a comment on all of the exercises.

exercises/rational-numbers/README.md
```diff
@@ -0,0 +1,27 @@
+A rational number is defined as the quotient of two integers `a` and `b`, called the numerator and denominator, respectively, where `b != 0`.
```
sjwarner-bp: This README looks a little incomplete.

If you use the [configlet](https://github.com/exercism/configlet/) tool, it will generate the README for you. Normally for our track, we have a couple of added bits that configlet will automatically pull in.

The [script](https://github.com/exercism/java/blob/master/bin/fetch-configlet) should pull down configlet for you.

exercises/rational-numbers/.meta/src/reference/java/Rational.java
```diff
@@ -0,0 +1,102 @@
...
+    public Rational abs() {
+        return new Rational(abs(this.numerator), abs(this.denominator));
+    }
+
+    private int abs(int x) {
+        final int mask = x >> 31;
+        return (mask ^ x) - mask;
+    }
+
+    public Rational add(Rational that) {
```
sjwarner-bp: There's a good few methods here that have a higher wider access modifier than they need to.

On the track here, we try to [adhere to best practise](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) as noted in our POLICIES.

This (and a few more) can be changed from `public` to package-private.

config.json
```diff
@@ -1117,6 +1117,18 @@
...
       "uuid": "52d11278-0d65-4b5b-b387-1374fced3243"
     },
+    {
+      "core": false,
+      "difficulty": 8,
+      "slug": "rational-numbers",
+      "topics": [
+        "floating_point_numbers",
+        "mathematics",
+        "classes"
```
sjwarner-bp: What is your reasoning for `classes`?

If you couldn't find it (I always have trouble locating it) the [list of topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) is super useful here!

exercises/rational-numbers/src/test/java/RationalTest.java
```diff
@@ -0,0 +1,260 @@
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class RationalTest {
+
+    // Helper methods
+
+    private static final double DOUBLE_EQUALITY_TOLERANCE = 1e-15;
```
sjwarner-bp: This seems like a sensible value for a tolerance.

I haven't run this locally yet - is the tolerance necessary? (I can see why it might be) :slightly_smiling_face:

config.json
clementi: @sjwarner-bp I saw that the topics of complex-numbers, which is similar to this problem, has classes. That's all. I'll take a look at that list.

exercises/rational-numbers/src/test/java/RationalTest.java
clementi: It may be, depending on the algorithm used to compute `double pow(double x)` or `double exp(double base)`. I included it because the complex-numbers project has it, and I thought it might be needed for these methods.

exercises/rational-numbers/README.md
clementi: That's great; I didn't know about that. I'll check it out.

config.json
sjwarner-bp: It's not the easiest list to find! :laughing: Your suggestions seem sensible so far, but feel free to add anything else you think might be suitable!

exercises/rational-numbers/src/test/java/RationalTest.java
sjwarner-bp: That makes sense to me - thanks :slightly_smiling_face:

exercises/rational-numbers/.meta/src/reference/java/Rational.java
```diff
@@ -0,0 +1,102 @@
+public class Rational {
+    private int numerator;
+    private int denominator;
+
+    Rational(final int numerator, final int denominator) throws IllegalArgumentException {
```
FridaTveit: `IllegalArgumentException` is an [unchecked exception](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/) therefore you don't have to declare it in the method signature 🙂

exercises/rational-numbers/src/test/java/RationalTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+    @Test
+    public void testAddTwoPositiveRationalNumbers() {
+        Rational expected = new Rational(7, 6);
+        Rational actual = new Rational(1, 2).add(new Rational(2, 3));
+        assertRationalsEqual(expected, actual);
+    }
+
+    @Test
+    @Ignore
```
FridaTveit: As described in our [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-testshttps://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests) this should be `@Ignore("Remove to run test")` and it should be above the `@Test` annotation 🙂

exercises/rational-numbers/src/test/java/RationalTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+    private static final double DOUBLE_EQUALITY_TOLERANCE = 1e-15;
+
+    private void assertDoublesEqual(double x, double y) {
+        assertEquals(x, y, DOUBLE_EQUALITY_TOLERANCE);
+    }
+
+    private void assertRationalsEqual(Rational r1, Rational r2) {
+        assertEquals(r1.getNumerator(), r2.getNumerator());
+        assertEquals(r1.getDenominator(), r2.getDenominator());
```
FridaTveit: Will this give a good error message if the test fails? That is will it clearly tell the user that it's the denominator or the numerator that's wrong or will it just say something like `Expected 5 but got 2`?

exercises/rational-numbers/.meta/src/reference/java/Rational.java
```diff
@@ -0,0 +1,102 @@
+public class Rational {
```
FridaTveit: This could be [`package private`](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) 🙂

exercises/rational-numbers/src/test/java/RationalTest.java
clementi: @FridaTveit It won't, no. It looks like this:
```
RationalTest > testAddTwoPositiveRationalNumbers FAILED
    java.lang.AssertionError: expected:<9> but was:<6>
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at RationalTest.assertRationalsEqual(RationalTest.java:18)
        at RationalTest.testAddTwoPositiveRationalNumbers(RationalTest.java:27)
```

I cribbed the `assertRationalsEqual` helper method from the `assertComplexNumbersEqual` method in `ComplexNumberTest.java` in `complex-numbers`. Would you like it to be more specific? I suppose we could implement `equals` and maybe `toString` on `Rational`, or require the exercise-taker to do it. That way, we could get better error messages.

FridaTveit: You could implement `equals`, that would help. In which case I would add some starter implementation with `equals` already implemented since that's not really part of the exercise. Or you could just add a message to your equals check saying something like `Incorrect numerator`. I think either would be fine 🙂 

You're right, it is the same in complex-numbers, thanks for pointing it out! I'll open an issue to improve the error message for complex-numbers 🙂

config.json
```diff
@@ -1117,6 +1117,19 @@
...
       "uuid": "52d11278-0d65-4b5b-b387-1374fced3243"
     },
+    {
+      "core": false,
+      "difficulty": 8,
+      "slug": "rational-numbers",
+      "topics": [
+        "floating_point_numbers",
+        "mathematics",
+        "classes",
```
FridaTveit: These look like good topics 🙂 But I'm not sure about `floating_point_numbers` and `classes`, how are they relevant to this exercise?

clementi: @FridaTveit Well, this exercise is similar to `complex-numbers`, and that exercise has those topics. Floating-point numbers (doubles) are involved in some of the methods of this exercise, and this exercise does require that one create a class with some data hiding. But if you don't think this exercise warrants those topics, I'll remove them.

sjwarner-bp: I think `classes` has recently been removed as a topic of `complex-numbers` :slightly_smiling_face:

clementi:
New commits:
* Remove 'classes' from topics
* Stub out `Rational` with `equals` and `toString`
* Use `assertEquals` in tests

exercises/rational-numbers/.meta/src/reference/java/Rational.java
```diff
@@ -0,0 +1,118 @@
...
+        return Math.pow(10.0, Math.log10(Math.pow(base, this.numerator)) / this.denominator);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%d/%d", this.getNumerator(), this.getDenominator());
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
FridaTveit: One small thing: if you're overriding `equals` then you should also override `hashCode` (see [this stackoverflow question](https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java) for an explanation of why 🙂 ). But you've done a lot of great work on this PR already so if you'd rather just get it merged then I can open a new issue for overriding `hashCode` 🙂

clementi: I think we can make an issue of it.

sjwarner-bp:
Opened #1296

PR #1282, opened 2018-02-02
rename plf HINTS and regenerate READMEs

sjwarner-bp:
When I went to regenerate the READMEs, it was removing the contents of `HINTS.md` from plf README.

Once this was renamed to `hints.md`, the problem stopped. I am not sure if this issue has always existed, or some change to configlet might now mean `hints.md` is case sensitive.

---

Reviewer Resources:


kytrinyx:
> I am not sure if this issue has always existed, or some change to configlet might now mean hints.md is case sensitive.

For what it's worth the old convention was `exercises/:slug/HINTS.md` and the new one is `exercises/:slug/.meta/hints.md` -- it was never supposed to be `exercises/:slug/.meta/HINTS.md`.

I wouldn't be surprised if configlet at some point allowed for the `.meta/HINTS.md` variant, though.

I've gone through and made sure no other tracks are suffering from this particular issue.

sjwarner-bp:
Thanks for clearing that one up for me! 🙂

PR #1283, opened 2018-02-02
update palindrome-products

jssander:
Added version file
added tests
updated .meta reference solution
---

Reviewer Resources:


exercises/palindrome-products/.meta/src/reference/java/PalindromeCalculator.java
```diff
@@ -3,6 +3,8 @@
 public class PalindromeCalculator {
 
     public SortedMap<Long, List<List<Integer>>> getPalindromeProductsWithFactors(int minFactor, int maxFactor) {
+    	if(minFactor > maxFactor) throw new IllegalArgumentException("invalid input: min is " + minFactor + " and max is " + maxFactor);
```
sjwarner-bp: This would read more nicely if it was presented as:

```
if (minFactor > maxFactor) {
    throw new IllegalArgumentException("invalid input: min is " + minFactor + " and max is " + maxFactor);
}
```

exercises/palindrome-products/.meta/src/reference/java/PalindromeCalculator.java
```diff
@@ -19,6 +21,9 @@
                 }
             }
         }
+        
+        if(palindromeSortedList.size() == 0) throw new NoSuchElementException("no palindrome with factors in the range " + minFactor + " to " + maxFactor);
```
sjwarner-bp: This would read more nicely if it was presented as:
```
if (palindromeSortedList.size() == 0)
    throw new NoSuchElementException("no palindrome with factors in the range " + minFactor + " to " + maxFactor);
```

exercises/palindrome-products/src/test/java/PalindromeCalculatorTest.java
```diff
@@ -1,25 +1,48 @@
...
+    public ExpectedException expectedException = ExpectedException.none();
 
     @Before
     public void setup() {
         palindromeCalculator = new PalindromeCalculator();
     }
+    
+    @Test
+    public void smallestPalindromeFromSingleDigitFactors() {
+        final List<List<Integer>> expected = Collections.unmodifiableList(
```
sjwarner-bp: At present, we try to avoid use of the keyword `final` as per our [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md#avoid-using-final)

exercises/palindrome-products/.meta/src/reference/java/PalindromeCalculator.java
FridaTveit: Please add curly brackets as well! 🙂 Otherwise it's easy to make mistakes (see https://stackoverflow.com/questions/8020228/is-it-ok-if-i-omit-curly-braces-in-java for more details) 🙂

jssander:
I have removed the final keyword and added curly brackets.

PR #1292, opened 2018-02-06
Diamond:update tests and version file

### LINKED ISSUE #1277: diamond: update tests and version file (1.0.0 -> 1.1.0)
The [canonical data for diamond](https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/canonical-data.json) has diverged from our test suite. Please update the tests, the version file and the reference implementation if necessary.

Haozhe321:
Fixes #1277 

1. Update test 
2. Update version file number

---

Reviewer Resources:


exercises/diamond/README.md
```diff
@@ -38,18 +38,16 @@ C···C
 ··A··
 ```
 
-Diamond for letter 'E':
+Diamond for letter 'D':
```
sjwarner-bp: I don't think that this change is part of the problem-specifications :slightly_smiling_face:

sjwarner-bp: I don't think this change is from the problem-specifications :slightly_smiling_face:

Haozhe321:
@sjwarner-bp The README was changed to be consistent with the changes done in the test suite. If the README and test suite, I can revert the change to the README.

sjwarner-bp:
The READMEs should be automatically generated through `configlet`.

This tool looks at [problem-specifications/exercises/diamond/description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/description.md) and essentially copies it in with a bit of extra Java stuff stuck around the edges.

It looks as though the description still uses examples of 'A', 'C' and 'E' diamonds, so we should be ok to keep the README as it is.

Does this make sense? :slightly_smiling_face: 

If you think that users could benefit from the README making use of your changes, please feel free to open an issue in the problem-specifications repo and see what the others think :slightly_smiling_face:

Haozhe321:
On a unrelated note, can I ask how is the test suite calling the main Class(in this case DiamondPrinter), when the DiamondPrinter is not being imported? In addition, there are two DiamondPrinter classes(one which the user is supposed to fill in, and one provided as an answer). How does `gradle test` know which file to take from? Thanks in advance!

Haozhe321:
Thanks for the explanation on README! I've put back the changes :)

sjwarner-bp:
Hi @Haozhe321 - we ensure that the reference solution compiles and passes all tests through the use of [this file](https://github.com/exercism/java/blob/master/exercises/build.gradle). Take a look, it should be fairly self-explanatory! :slightly_smiling_face:

sjwarner-bp:
Thanks @Haozhe321 :slightly_smiling_face:

PR #1297, opened 2018-02-07
update matrix tests

jssander:
I removed all of the old matrix tests which were parameterized and added new tests matching the canonical file which are not parameterized.

I also added a version file.
---

Reviewer Resources:


exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -10,198 +10,100 @@
...
-        }
-
-        @Test
-        public void getLastColumnTest() {
-            assertArrayEquals(lastColumn, matrix.getColumn(matrix.getColumnsCount() - 1));
-        }
-    }
+	
+	@Test
+	public void ExtractRowFromOneNumberMatrixTest() {
```
FridaTveit: In java method names should start with a lowercase letter 🙂

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -10,198 +10,100 @@
...
+		String matrixAsString = "1";
+		int rowIndex = 0;
+		int[] row = {1};
```
FridaTveit: A very minor thing, but I think if this is name `expectedRow` then it makes the test a bit easier to read 🙂

exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -10,198 +10,100 @@
...
-        }
-
-        public CountColumnsTest(String matrixAsString, int numberOfColumns) {
-            this.matrix = new Matrix(matrixAsString);
-            this.numberOfColumns = numberOfColumns;
-        }
-
-        @Test
-        public void countColumnsTest() {
-            assertEquals(numberOfColumns, matrix.getColumnsCount());
```
FridaTveit: Since you've removed these tests you can probably simplify the referece implementation a bit by removing `getColumnsCount` and `getRowsCount` 🙂

exercises/matrix/src/test/java/MatrixTest.java
sjwarner-bp: Check out [something like this](http://www.oracle.com/technetwork/java/codeconventions-135099.html) for a couple of cool tips about Java style! :slightly_smiling_face:

jssander:
I made the changes you requested.

PR #1299, opened 2018-02-08
rational-numbers: override hashCode

### LINKED ISSUE #1296: rational-numbers: override hashCode
In the reference solution for `rational-numbers`, the class overrides the equals method but does not override the hashCode method.

This violates the general contract for Object.hashCode(), and as a consequence hash-based structures could fail to work correctly.

See more at this [stackoverflow question](https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java).

Please add an override for the hashCode method to `rational-numbers`.

jackattack24:
Fix #1296 

---

Reviewer Resources:

### DIFF
--- a/exercises/rational-numbers/.meta/src/reference/java/Rational.java
+++ b/exercises/rational-numbers/.meta/src/reference/java/Rational.java
@@ -115,4 +115,14 @@ class Rational {
         return this.getNumerator() == other.getNumerator()
             && this.getDenominator() == other.getDenominator();
     }
+
+    @Override
+    public int hashCode() {
+        int prime = 31;
+        int result = 1;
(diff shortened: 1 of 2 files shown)


sjwarner-bp:
Thanks for the contribution @jackattack24 :fireworks: 

As an aside, I wonder if maybe we should add a case to the test suite which checks that equals and hashcode have been correctly overridden? Something like

```
@Test
    public void testEqualsAndHashCode() {
        EqualsVerifier.forClass(Rational.class).suppress(Warning.NONFINAL_FIELDS).verify();
}
```

Or would this be excessive / too complicated? @FridaTveit what do you think? :slightly_smiling_face:

PR #1301, opened 2018-02-10
rational-numbers: remove starter implementation

sjwarner-bp:
As per our POLICIES.md, exercises over a difficulty of 4 should not have starter implementations.

```
Exercises of difficulty 5 or higher: provide no stubs by default, but either:
(1) add hints to the hints.md file (which gets merged into the README.md for the exercise)
or
(2) provide stubs as above for exercises that demand complicated method signatures.
```

I think I just missed this in review, though if there has been specific conversation and either (1) or (2) have been decided as valid reasons to keep this, then let me know 🙂 

---

Reviewer Resources:


sjwarner-bp:
That makes sense to me! Thanks for clearing that up @FridaTveit - I'll close this now :slightly_smiling_face:

PR #1305, opened 2018-02-11
Update PigLatinTranslatorTest

### LINKED ISSUE #1293: Remove use of parameterized tests

- [x] raindrops
- [x] matrix
- [x] binary
- [ ] octal
- [x] scrabble-score
- [x] roman-numerals
- [x] prime-factors
- [x] atbash-cipher
- [x] pig-latin

LuLechuan:
Fixes #1293 
For exercise - PigLatinTranslator

---

Reviewer Resources:


exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -1,85 +1,95 @@
...
-    @Parameterized.Parameters(name = "{index}: expected \"{0}\" to translate to the pig latin phrase \"{1}\"")
-    public static Collection<Object[]> data() {
-        return Arrays.asList(new Object[][]{
-                // Ay is added to words that start with vowels
-                {"apple", "appleay"},
-                {"ear", "earay"},
-                {"igloo", "iglooay"},
-                {"object", "objectay"},
-                {"under", "underay"},
+    public PigLatinTranslatorTest() {
```
sjwarner-bp: I think you can probably drop this constructor, as Java will automatically add a default constructor for you!

See this [stackoverflow question](https://stackoverflow.com/questions/4488716/java-default-constructor) for more info! 🙂

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -1,85 +1,95 @@
...
+        pigLatinTranslator = new PigLatinTranslator();
+    }
 
-                // First letter and ay are moved to the end of words that start with consonants
-                {"pig", "igpay"},
-                {"koala", "oalakay"},
-                {"xenon", "enonxay"},
+    @Test
+    public void testWordsStartWithVowels() {
+        assertEquals("appleay", pigLatinTranslator.translate("apple"));
```
sjwarner-bp: It may make more sense to separate each of these `assertEquals` statements out into individual tests. @exercism/java what do you guys think? 🙂

sjwarner-bp:
Another minor comment (though it might fall outside of the scope of the related issue) is that the current `version` of this test suite is 1.2.0, and our `version` details 1.1.0.

It may be worth updating the version number with this exercise 🙂

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
LuLechuan: Actually I believe they should be group together since they all belong to the same type?

sjwarner-bp: It depends how we want to think about this - I personally prefer tests to be as granular as possible. For exercises like this, where simply checking if words start with vowels or consonants isn't actually sufficient to 'truly' convert to pig latin (there isn't one easy rule), I think it can be of benefit to guide the user a little more carefully.

For example, the assertions in `testWordsWithPrefixTreatedAsSingleConsonant()` are actually an amalgamation of a handful of exceptions to the 'easy' rules.

I'd be interested in hearing everyone else's views on this - I have a particular vendetta against this exercise, so maybe I am just being picky :wink:

FridaTveit: I agree with @sjwarner-bp, I think each test should only test one thing. This is for several reasons:
1. Since we're promoting TDD, if each test only tests one thing (in this case one word) it'll be much easier to keep track of your progress by how many tests pass.

2. I think it's quite rewarding when you make a test pass and you wouldn't  get that as often if each test has several assertions.

3. We should match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pig-latin/canonical-data.json) as closely as possible, and the canonical data only tests one word at a time. Furthermore, the people making the canonical data try to make sure that each word tests a different thing and that they're in a good order. That is, if you test each word individually, it shouldn't be the case that if the test for one word passes then the tests for some subsequent words will definitely pass. So the canonical data is deliberately constructed so that each word should have its own test.

Does that make sense? 🙂

LuLechuan: Hi, thanks for the clarification! Yes these are really reasonable. I will make the changes accordingly for each PR.

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -1,85 +1,157 @@
...
+        // Xr is treated like a single vowel
+        assertEquals("xrayay", pigLatinTranslator.translate("xray"));
+    }
 
-                // 'Y' as second letter in a two letter word
-                {"my", "ymay"},
+    @Ignore("Remove to run test")
+    @Test
+    public void testYTreatedLikeAConsonantAtTheBeginningOfAWord() {
+        // 'Y' is treated like a consonant at the beginning of a word
```
FridaTveit: From what I can tell this comment only repeats the test title so is it really needed? 🙂 

The same applies for the comments in the next two tests 🙂

exercises/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -1,85 +1,157 @@
...
+    public void testWordBeginningWithQWithoutAFollowingU() {
+        assertEquals("atqay", pigLatinTranslator.translate("qat"));
+    }
 
-                // Qu and a single preceding consonant are treated like a single consonant
-                {"square", "aresquay"},
+    @Ignore("Remove to run test")
+    @Test
+    public void testWordBeginningWithCh() {
+        // Ch is treated like a single consonant
```
FridaTveit: Instead of needed this comment you could try and fit the same information in the test title, e.g. ` testChTreatedLikeAConsonantAtTheBeginningOfAWord`. The test title will show up in the error message so it's useful if it describes what the test is testing as accurately as possible 🙂 

The same applies for the comments in the other tests. The comments were needed when the tests were parameterized since we didn't have individual test names. But now that we do I don't think we need the comments anymore. What do you think? 🙂

LuLechuan: Yeah that's reasonable, will make change accordingly

LuLechuan:
Changes made as requested. @sjwarner-bp @FridaTveit Thanks for teaching how to follow a good style! :)

PR #1307, opened 2018-02-11
Update AtbashTest

### LINKED ISSUE #1293: Remove use of parameterized tests

LuLechuan:
Fixes #1293 
For exercise: Atbash-cipher

---

Reviewer Resources:

### DIFF
--- a/exercises/atbash-cipher/src/test/java/AtbashTest.java
+++ b/exercises/atbash-cipher/src/test/java/AtbashTest.java
@@ -14,44 +14,74 @@ public class AtbashTest {
     }
 
     @Test
-    public void testEncodingSingleWord() {
+    public void testEncodeYes() {
         assertEquals("bvh", atbash.encode("yes"));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testEncodeNo() {
         assertEquals("ml", atbash.encode("no"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testEncodingWordsWithCapitalCases() {
+    public void testEncodeOmgInCapital() {
         assertEquals("lnt", atbash.encode("OMG"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testEncodingWordsWithSpace() {
+    public void testEncodeOmgWithSpaces() {
         assertEquals("lnt", atbash.encode("O M G"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testEncodingWordsLeadingToCodeWithSpace() {
+    public void testEncodeMindBlowingly() {
         assertEquals("nrmwy oldrm tob", atbash.encode("mindblowingly"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testEncodingSentences() {
+    public void testEncodeNumbers() {
         assertEquals("gvhgr mt123 gvhgr mt", atbash.encode("Testing,1 2 3, testing."));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testEncodeDeepThought() {
         assertEquals("gifgs rhurx grlm", atbash.encode("Truth is fiction."));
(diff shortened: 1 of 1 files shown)


exercises/atbash-cipher/src/test/java/AtbashTest.java
```diff
@@ -1,73 +1,44 @@
...
-            return Arrays.asList(new Object[][]{
-                    {"vcvix rhn", "exercism"},
-                    {"zmlyh gzxov rhlug vmzhg vkkrm thglm v", "anobstacleisoftenasteppingstone"},
-                    {"gvhgr mt123 gvhgr mt", "testing123testing"},
-                    {"gsvjf rxpyi ldmul cqfnk hlevi gsvoz abwlt", "thequickbrownfoxjumpsoverthelazydog"}
-            });
-        }
+    @Test
+    public void testEncodingSingleWord() {
+        assertEquals("bvh", atbash.encode("yes"));
```
sjwarner-bp: It might make more sense to separate each of the `assertEquals` statements into separate tests, as they test different features of the `encode()` method.

For example, case and spaces 🙂

exercises/atbash-cipher/src/test/java/AtbashTest.java
```diff
@@ -1,73 +1,59 @@
...
-        }
+    @Before
+    public void setup() {
+        atbash = new Atbash();
+    }
 
+    @Test
+    public void testEncodingSingleWord() {
+        assertEquals("bvh", atbash.encode("yes"));
+        assertEquals("ml", atbash.encode("no"));
```
FridaTveit: As discussed in PR #1305, I would prefer it if these were split into one assertion per test to match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/atbash-cipher/canonical-data.json) 🙂

exercises/atbash-cipher/src/test/java/AtbashTest.java
```diff
@@ -1,73 +1,59 @@
...
+    }
 
-        public DecodeTest(String ciphertext, String plaintext) {
-            this.ciphertext = ciphertext;
-            this.plaintext = plaintext;
-        }
+    @Ignore("Remove to run test")
+    @Test
+    public void testEncodingSentences() {
+        assertEquals("gvhgr mt123 gvhgr mt", atbash.encode("Testing,1 2 3, testing."));
```
FridaTveit: In this case, if someone solving this exercise has a solution which works perfectly apart from that it doesn't encode numbers properly then this test will fail. I think that might not be a very good use experience since the test name is `testEncodingSentences` which doesn't very well point out what the problem is.  So this is one of the reasons I think it would be good to split the tests out to only have one assertion per test 🙂 What do you think? 🙂

exercises/atbash-cipher/src/test/java/AtbashTest.java
```diff
@@ -1,73 +1,59 @@
...
+    }
 
-        @Ignore("Remove to run tests")
-        @Test
-        public void test() {
-            assertEquals(plaintext, new Atbash().decode(ciphertext));
-        }
+    @Ignore("Remove to run test")
+    @Test
+    public void testDecoding() {
```
FridaTveit: Since this test has all the decoding cases that means there's no TDD for implementing decoding

exercises/atbash-cipher/src/test/java/AtbashTest.java
LuLechuan: Ok sure, I will make the changes accordingly.

FridaTveit:
Haha, I didn't notice I had approved it twice 😛

PR #1313, opened 2018-02-13
binary: removed use of parameterized tests

c-thornton:
Removed the parameterized tests and ensured that the tests follow the current canonical data.

---

Reviewer Resources:

### DIFF
--- a/exercises/binary/src/test/java/BinaryTest.java
+++ b/exercises/binary/src/test/java/BinaryTest.java
@@ -1,49 +1,114 @@
 import org.junit.Test;
 import org.junit.Ignore;
-import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


FridaTveit:
Good idea but I think it might not be worth updating it since the exercise is deprecated 🙂 Though I guess contributors could look at it as an example exercise and get confused by the inconsistency... I don't mind, if you think it's worth fixing @sjwarner-bp then we can, but I don't think it matters too much since it's deprecated 🙂

sjwarner-bp:
That makes sense to me - I forgot this one was deprecated! Just got a bit confused :laughing: 

I'm happy to leave this as it is then :slightly_smiling_face:

PR #1315, opened 2018-02-13
new two-bucket exercise

jssander:
---

Reviewer Resources:


config.json
```diff
@@ -612,6 +612,18 @@
...
+    {
+      "core": false,
+      "difficulty": 5,
+      "slug": "two-bucket",
+      "topics": [
+        "algorithms",
+        "loops",
+        "conditionals"
+      ],
+      "unlocked_by": null,
```
sjwarner-bp: It would be great to add an `unlocked_by`, as we will want one for basically all the exercises for nextercism!

The only requirement is that it must be unlocked by a core exercise 🙂

exercises/two-bucket/src/main/java/TwoBucket.java
```diff
@@ -0,0 +1,19 @@
+
```
sjwarner-bp: As per [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations), because this is higher than a difficulty of 4 we don't need the starter implementation here 🙂

exercises/two-bucket/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Example:
+Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say bucket one, at a given step, is holding 7 liters, and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one "move". Instead, if you had poured from bucket one into bucket two until bucket two was full, leaving you with 4 liters in bucket one and 11 liters in bucket two (4,11), that would count as only one "move" as well.
+
+To conclude, the only valid moves are:
+- pouring from one bucket to another
+- emptying one bucket and doing nothing to the other
+- filling one bucket and doing nothing to the other
+
+Written with <3 at [Fullstack Academy](http://www.fullstackacademy.com/) by [Lindsay](http://lindsaylevine.com).
```
sjwarner-bp: Looks like this link is dead...

Actually, looking at the [metadata](https://github.com/exercism/problem-specifications/blob/master/exercises/two-bucket/metadata.yml), this README might be out of date.

Make sure you have the most recent version of `problem-specifications` before generating the README! From your problem-specifications clone:

```bash
git checkout master
git remote add upstream https://github.com/exercism/problem-specifications.git
git fetch upstream
git merge upstream/master
```

Then regenerate it 😄

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
...
+    	
+    	public State (int moves, int bucketOne, int bucketTwo) {
+    		this.moves = moves;
+    		this.bucketOne = bucketOne;
+    		this.bucketTwo = bucketTwo;
+    	}
+    	
+    	@Override
+    	public boolean equals(Object o) {
+    		State otherState = (State)o;
```
sjwarner-bp: Nitpick! I think a space could be added here:

```java
State otherState = (State) o;
```

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
...
+    	int bucketTwo;
+    	
+    	public State (int moves, int bucketOne, int bucketTwo) {
+    		this.moves = moves;
+    		this.bucketOne = bucketOne;
+    		this.bucketTwo = bucketTwo;
+    	}
+    	
+    	@Override
+    	public boolean equals(Object o) {
```
sjwarner-bp: I'm not too sure if we need to override `equals` here or not, as we don't test it.

If we do need to override `equals` then we might want to override `hashcode` too - otherwise we break the general contract for hashcode! [See here](https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java)

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
+import java.util.ArrayList;
+
+class TwoBucket {
+	
+    private class State {
+    	int moves;
+    	int bucketOne;
+    	int bucketTwo;
+    	
+    	public State (int moves, int bucketOne, int bucketTwo) {
```
sjwarner-bp: I think we could have this as package-private 🙂

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    private State finalState;
+    
+    int bucketOneCap;
+    int bucketTwoCap;
+    int desiredLiters;
+    String first;
+    
+    public TwoBucket(int bucketOneCap, int bucketTwoCap, int desiredLiters, String first) {
```
sjwarner-bp: package-private here too I think!

exercises/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -0,0 +1,85 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class TwoBucketTest {
+	
+	@Test
+	public void testBucketOneSizeThreeBucketTwoSizeFiveStartWithOne() {
+		
+		TwoBucket twoBucket = new TwoBucket(3, 5, 1, "one");
+		
+		assertEquals(twoBucket.getTotalMoves(), 4);
```
FridaTveit: [`assertEquals`](http://junit.sourceforge.net/javadoc/org/junit/Assert.html) takes expected as the first argument and actual as the second so this (and all following assertions) should be:
```
assertEquals(4, twoBucket.getTotalMoves());
```
Otherwise the error message will be confusing 🙂

exercises/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
```
FridaTveit: I don't think any of these imports are needed except for `Ignore`, `Test` and `assertEquals`?

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
FridaTveit: Agreed, you don't need to override `equals`. You only need to override it if you want to compare two objects of this class, for example in `assertEquals`. However, as far as I can tell you're not comparing two `State` objects at any point so I don't think it's needed 🙂

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
...
+    				this.bucketTwo == otherState.bucketTwo;
+    	}
+    }
+    
+    private State finalState;
+    
+    int bucketOneCap;
+    int bucketTwoCap;
+    int desiredLiters;
+    String first;
```
FridaTveit: These should all be `private` since it's good practice to [limit access to variables as much as possible](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) 🙂 

Also, what's `first` meant to represent? The names of the other variables make it very clear what they are, but this one is a bit more vague. If it's meant to represent the bucket you start with, maybe `firstBucket` or `startBucket` would be more descriptive? 🙂

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,139 @@
...
+    	adjacentStates.add(new State(state.moves + 1, state.bucketOne, 0));
+    	
+    	//Fill bucket one
+    	adjacentStates.add(new State(state.moves + 1, bucketOneCap, state.bucketTwo));
+    	
+    	//Fill bucket two
+    	adjacentStates.add(new State(state.moves + 1, state.bucketOne, bucketTwoCap));
+    	
+    	//pour from bucket one to bucket two
+    	if(state.bucketOne + state.bucketTwo > bucketTwoCap) {
```
FridaTveit: Minor nitpick: there should be a whitespace after `if` in this and the other `if` statements 🙂

config.json
```diff
@@ -612,6 +612,18 @@
...
       "uuid": "d36ce010-210f-4e9a-9d6c-cb933e0a59af"
     },
+    {
+      "core": false,
+      "difficulty": 5,
+      "slug": "two-bucket",
+      "topics": [
+        "algorithms",
+        "loops",
+        "conditionals"
```
FridaTveit: Maybe `mathematics` would be a good topic to add to this list? 🙂

config.json
jssander: Which exercise do you think it should be unlocked by?

sjwarner-bp: Hmm, I am not too sure. `triangle` is kind of mathematical with a couple of special rules applied - maybe this?

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
jssander: I only implemented equals because I use the contains method. Won't the contains method call equals?

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
jssander: package-private is no explicit modifier, right?

exercises/two-bucket/src/main/java/TwoBucket.java
jssander: I should just delete this file?

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
sjwarner-bp: That's correct :slightly_smiling_face:

exercises/two-bucket/src/main/java/TwoBucket.java
sjwarner-bp: If you can delete it, and in its place put an empty `.keep` file :slightly_smiling_face:

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
FridaTveit: Oh I see! I didn't spot that sorry. Thanks for explaining 🙂 In which case it's fine to override `equals` but I agree with @sjwarner-bp that you should override `hashCode` as well 🙂

exercises/two-bucket/README.md
jssander: I followed your instructions and regenerated the README, but I think it is the same as before.

jssander:
I think I've made all the changes you asked for, with the exception of the README file which I think is still being generated the same even though I updated the problem specifications repository. I'm not sure what to do about that.

exercises/two-bucket/README.md
Smarticles101: @jssander and @sjwarner-bp the README here is up to date, [the problem-specification description](https://github.com/exercism/problem-specifications/blob/master/exercises/two-bucket/description.md) has the broken link

config.json
```diff
@@ -612,6 +612,19 @@
...
     },
+    {
+      "core": false,
+      "difficulty": 5,
+      "slug": "two-bucket",
+      "topics": [
+        "algorithms",
+        "loops",
+        "conditionals",
+	"mathematics"
```
Smarticles101: this needs an indent :)

exercises/two-bucket/README.md
jssander: The README where?

Smarticles101: @jssander I was referring to the one you generated. It is up to date with the canonical data, it's just the canonical data had the broken link in the description for the exercise. I've submitted a pull request over there to resolve the issue.

exercises/two-bucket/.meta/src/reference/java/TwoBucket.java
```diff
@@ -0,0 +1,145 @@
...
+    		adjacentStates.add(new State(state.moves + 1, state.bucketTwo + state.bucketOne, 0));
+    	}
+    	
+    	return adjacentStates;
+    }
+    
+    private boolean isValid(State state) {
+    	if (state.bucketOne == bucketOneCap && state.bucketTwo == 0 && startBucket.equals("two")) {
+    		return false;
+    	} else if(state.bucketOne == 0 && state.bucketTwo == bucketTwoCap && startBucket.equals("two")) {
```
FridaTveit: This should also have a space after `if` 🙂 It's a very minor thing so you don't have to fix it, but if you want to fix it when you're fixing the indentation of the topics then that would be great 🙂

exercises/two-bucket/README.md
sjwarner-bp: Ah OK, I must have gotten confused because of the Lindsay link! I think I submitted a PR (maybe for a different exercise) with the same broken link a while ago 🙂

In that case once the indentation is fixed I am happy to merge it and will regenerate the README.md myself when the changes are merged to problem specifications!

PR #1316, opened 2018-02-14
Remove Parameterized tests for octal

### LINKED ISSUE #1293: Remove use of parameterized tests

Haozhe321:
Fixes #1293

---

Reviewer Resources:

### DIFF
--- a/exercises/octal/src/test/java/OctalTest.java
+++ b/exercises/octal/src/test/java/OctalTest.java
@@ -1,49 +1,99 @@
 import org.junit.Test;
 import org.junit.Ignore;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-
-import java.util.Arrays;
-import java.util.Collection;
 
 import static org.junit.Assert.assertEquals;
 
-@RunWith(Parameterized.class)
 public class OctalTest {
-
-    private String octalNumberAsString;
-    private int decimalNumber;
-
-    @Parameterized.Parameters(name = "{index}: expected {1} when converting \"{0}\" from octal to decimal.")
-    public static Collection<Object[]> data() {
-        return Arrays.asList(new Object[][]{
-                {"1", 1},
-                {"10", 8},
-                {"17", 15},
-                {"11", 9},
-                {"130", 88},
-                {"2047", 1063},
-                {"7777", 4095},
-                {"1234567", 342391},
-                {"carrot", 0},
-                {"8", 0},
-                {"9", 0},
-                {"6789", 0},
-                {"abc1z", 0},
-                {"011", 9}
-        });
-    }
-
-    public OctalTest(String octalNumberAsString, int decimalNumber) {
-        this.octalNumberAsString = octalNumberAsString;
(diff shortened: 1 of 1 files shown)


exercises/octal/src/test/java/OctalTest.java
```diff
@@ -1,49 +1,99 @@
...
-import org.junit.runners.Parameterized;
-
-import java.util.Arrays;
-import java.util.Collection;
 
 import static org.junit.Assert.assertEquals;
 
-@RunWith(Parameterized.class)
 public class OctalTest {
+    private Octal octal;
```
FridaTveit: This doesn't need to be a private variable since it's being assigned to a new object in each test. It would be just as easy (and would save you one little line of code 😛 ) to remove the private variable and change each test to have:
```
Octal octal = new Octal("1");
```
🙂

exercises/octal/src/test/java/OctalTest.java
```diff
@@ -1,49 +1,99 @@
...
+
+    @Test
+    public void testOne() {
```
FridaTveit: These test names could maybe describe what the test does a bit better. Looking at this it's easy to read it as `the first test` instead of `test converting the number one`, if that makes sense? So maybe something like `testCovertDecimalOneToOctal` or `testConvertOneToOctal` would be more descriptive? 🙂

exercises/octal/src/test/java/OctalTest.java
c-thornton: I checked some of the other exercise tests and the ones I checked have the same private variable declaration at the beginning as well. For consistency should it be looked at to change those as well to match if it is changed here?

Haozhe321:
@FridaTveit Sure, I don't mind since future developers can still look at this exercise! I've made the requested changes :)

exercises/octal/src/test/java/OctalTest.java
Haozhe321: +1 to what @c-thornton said. I looked at some other tests and found the same thing, so I followed what was done. I've made changes to declaring a new variable in each test case. In light of the inconsistency for the test cases, is one way of declaration always preferred over the other?

sjwarner-bp: It could be something worth looking at :slightly_smiling_face: 

I'll do some investigating later and see what the situation is like!

FridaTveit: That does seem worth looking at, thanks for raising it @c-thornton and @Haozhe321! 🙂 

I think it's good practice to have a private variable (initialised in a `@Before` method) if the class constructor takes no arguments. In that case it's useful since it saves you from having to repeat the line initialising the variable in each test.
If the constructor does take arguments then it would have to be initalised in each test in which case I don't think it's worth having a private variable. Does that sound sensible? 🙂

PR #1317, opened 2018-02-14
updated tests

### LINKED ISSUE #1264: raindrops: update version file (1.0.0 -> 1.1.0)
Please update the version specified in the [raindrops version file](https://github.com/exercism/java/blob/master/exercises/raindrops/.meta/version) from 1.0.0 to 1.1.0.

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of [the canonical data file for that exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/raindrops/canonical-data.json).

### LINKED ISSUE #1310: raindrops: make tests more closely match canonical data
In the [raindrops tests](https://github.com/exercism/java/blob/master/exercises/raindrops/src/test/java/RaindropConverterTest.java) we're currently testing the same cases as the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/raindrops/canonical-data.json). However, we deviate from the canonical data by:

1. The order of the tests. This is important since the order in the canonical data has been carefully chosen so that no test is useless. That is, making one test pass shouldn't necessarily make any of  the subsequent tests pass. That way each test is useful 🙂 

2.  How many assertions there are in each test. In the canonical data there is only one assertion per test. I think we should stick to that since [it's good practice to make tests as granular as possible](https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test). That way it's easy to track your progress by how many tests pass and you get the rewarding feeling of seeing a new test pass as often as possible. And it means that if a test fails then that should hopefully highlight only one specific issue, described in the test name.

So we should change the raindrops tests to have the same order and same number of assertions per test as the canonical data 🙂

tshradheya:
Fixes #1310 
Fixes #1264 

Updated Tests and Version number. Ready for Review
---

Reviewer Resources:


exercises/raindrops/src/test/java/RaindropConverterTest.java
```diff
@@ -17,28 +17,26 @@ public void soundFor1Is1() {
...
-    public void soundFor7OrNumberWithOnlyDistinctFactorOf7IsPlong() {
+    public void soundFor7IsPlong() {
         assertEquals("Plong", raindropConverter.convert(7));
-        assertEquals("Plong", raindropConverter.convert(14));
-        assertEquals("Plong", raindropConverter.convert(49));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void soundForNumber6WithOnlyDistinctFactorOf3IsPling() {
```
FridaTveit: The test names are a bit inconsistent. Some of them just state `the sound for this number is...` whereas others state `the sound for this number with factor... is...`. I think either is fine but it would be good to be consistent 🙂 Also some of them are of the form `soundForX` whereas others are `soundForNumberX`. Again I think either is fine but it would be good to be consistent 🙂 

If you want to stick to `the sound for this number with factor... is...` then I find the phrasing `with only distinct factor` a bit confusing. What exactly do you mean by that? The numbers do sometimes, like in this case, have other factors they're just not in the list of numbers that translate to raindrop sounds. I would maybe prefer something like `soundForXIsYAsItHasFactorZ` or something. What do you think? 🙂

tshradheya:
I have fixed both errors. Ready for Review

Issue #1279, opened 2018-02-01
phone-number: update version file (1.3.0 -> 1.4.0)


FridaTveit:
Please update the version specified in the [phone-number version file](https://github.com/exercism/java/blob/master/exercises/phone-number/.meta/version) from 1.3.0 to 1.4.0.

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of [the canonical data file for that exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json).

PR #1320, opened 2018-02-16
phone-number: update version number to 1.4.0

mau0727:
Updated

---

Reviewer Resources:


Smarticles101:
Thanks @mau0727 nice to see it's your first time contributing to the java track! Thanks for fixing this. For future reference a more descriptive pull request title would be better (like what i've modified the title to) and it's even better if you reference the issue you are solving: #1279 (although not required) just so there's more context as to what you're trying to solve in case the description and title are not clear.

I'll go ahead and merge this... Feel free to tackle another similar issue if you'd like, or maybe a more complex one! We're happy to help if you need!

Thanks again @mau0727

Issue #1326, opened 2018-02-19
New Exercise: Zipper


LuLechuan:
Hi, I would like to add a new exercise for this track - Zipper.
Problem description:
https://github.com/exercism/problem-specifications/blob/master/exercises/zipper/description.md

Smarticles101:
@LuLechuan Feel free! You're always welcome to add new exercises, you just need to open a pull request for it, an issue is not required.

PR #1327, opened 2018-02-20
Implement new exercise: Zipper

LuLechuan:
Implement Zipper #1326 
The reference file and test file are done.
However, I have some doubt for this exercise and would like to clarify before uploading README.md and config: 
From the problem-description: https://github.com/exercism/problem-specifications/blob/master/exercises/zipper/description.md, there are operations such as prev (move the focus to the previous child of the same parent), next(move the focus to the next child of the same parent), insert before, insert after and delete but they are not tested in the canonical file?
On the other hand, there is no operations called setLeft() or setRight() in the problem description but they are tested extensively in the canonical file.
I have created all these functions in the reference files, please correct me if I have misunderstood these operations. Should I add the tests for those operations that are not in the canonical files?
Thanks for the clarification!

---

Reviewer Resources:


LuLechuan:
@sjwarner-bp Ok.
For the difficulty level, I found the implemented zipper in the python track is an 8, so I just follow that?

sjwarner-bp:
@LuLechuan it depends! Some exercises will be very similar across all language tracks, but others will be particularly easy in some languages and much harder in others.

You don't have to just copy Python's difficulty. You wrote a reference solution - how hard did you find that compared to any other exercism Java exercises you've completed? :slightly_smiling_face:

LuLechuan:
@sjwarner-bp Well this question is an upgraded version for linked-list since they both use pointers in a list/tree. As linked-list is a 6, I would probably give this a 7.

exercises/zipper/.meta/src/reference/java/Zipper.java
```diff
@@ -0,0 +1,199 @@
+public class Zipper {
+
+    private int value;
+    BinaryTree tree; // Only the root node will bind to the BinaryTree
+    Zipper up;
+    Zipper left;
+    Zipper right;
+
+    public Zipper(int val) {
```
sjwarner-bp: As per our POLICIES document, we should limit access to minimal levels.

In that case, we should change this (and other public methods) to package private.

LuLechuan:
@sjwarner-bp Changes made as requested

LuLechuan:
@sjwarner-bp Ok sure, thanks for the review!

sjwarner-bp:
Thanks for the new exercise @LuLechuan!

It's always great to bulk the track out a bit :tada:

LuLechuan:
Thanks @sjwarner-bp and @Smarticles101 ! It's nice to work with you.

PR #1331, opened 2018-02-24
Add new exercise Diffie Hellman

CRivasGomez:
Hello.
I've implemented Diffie Hellman exercise.

Reviewer Resources:


config.json
```diff
@@ -1116,6 +1116,18 @@
...
+    {
+      "core": false,
+      "difficulty": 8,
+      "slug": "diffie-hellman",
+      "topics": [
+        "algorithms",
+        "integers",
+        "transforming"
+      ],
+      "unlocked_by": "book-store",
```
sjwarner-bp: The `unlocked_by` should be a core exercise, and `book-store` is not core.

Do you have a suggestion for a suitable value? 🙂

exercises/diffie-hellman/.meta/src/reference/java/DiffieHellman.java
```diff
@@ -0,0 +1,24 @@
+import java.math.BigInteger;
+import java.util.Random;
+
+public class DiffieHellman {
```
sjwarner-bp: As per our [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) we should restrict access as much as possible. Part of this would be making the class declaration package-private 🙂

exercises/diffie-hellman/.meta/version
```diff
@@ -0,0 +1,2 @@
+1.0.0
```
sjwarner-bp: Not sure if we should have a `version` for this, seeing as there is no canonical data for this exercise yet (this is where we normally source version numbers from).

exercises/diffie-hellman/README.md
```diff
@@ -0,0 +1,51 @@
...
+    s = A**b mod p
+
+The calculations produce the same result!  Alice and Bob now share
+secret s.
+
+## Hints
+This exercise requires you to perform calculations on large numbers. To correctly represent large numbers, the
+[BigInteger](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html) class is used.
+
+
```
sjwarner-bp: I'd normally expect a `Running the tests` section here, like we see in all of the other Java READMEs

exercises/diffie-hellman/src/main/java/DiffieHellman.java
```diff
@@ -0,0 +1,20 @@
+import java.math.BigInteger;
```
sjwarner-bp: As per our [POLICIES](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices) exercises with a difficulty of above 4 should not have starter implementations (except in extreme circumstances). As this has been rated as an 8 difficulty, we should remove this file and replace it with an empty `.keep`

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
+import org.junit.Before;
```
sjwarner-bp: Out of interest, did you make this test suite yourself?

I know there is no canonical data so far - did you think of these tests or port them from another exercise? 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+        BigInteger primeP = BigInteger.valueOf(23);
+        List<BigInteger> privateKeys = new ArrayList<>();
+        for (int i = 0; i < 10; i++) privateKeys.add(diffieHellman.privateKey(primeP));
+
+        for (BigInteger privateKey : privateKeys) {
+            assertTrue((privateKey.compareTo(BigInteger.ONE) >= 0) && (privateKey.compareTo(primeP) < 0));
+        }
+    }
+
+    // Note: due to the nature of randomness, there is always a chance that this test fails
```
sjwarner-bp: Not sure it would be ideal to have a test that could fail even if the implementation is correct.

This could cause issues with builds, and give us some grief. Could you explain to me why this test intermittently fails?

config.json
CRivasGomez: I would say that ```linked-list``` is the best candidate 🤔

exercises/diffie-hellman/.meta/src/reference/java/DiffieHellman.java
CRivasGomez: Yes, sorry. I'm going to remove the ```public``` modifier 😅

exercises/diffie-hellman/.meta/version
CRivasGomez: True, exercises ported to Java track with no canonical data don't have ```version```.

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: I port them from C# track [DiffieHellmanTest.cs](https://github.com/exercism/csharp/blob/master/exercises/diffie-hellman/DiffieHellmanTest.cs)

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: 5 different numbers are being generated (in a range of 0 to 7919) and we are checking that they are different to ensure that ```privateKey``` function generates random numbers. There is a very small chance of getting some number repeated and fail tests.

exercises/diffie-hellman/README.md
CRivasGomez: Added 🙂

exercises/diffie-hellman/src/main/java/DiffieHellman.java
CRivasGomez: Removed ```DiffieHellman.java``` and replaced with ```.keep```.

exercises/diffie-hellman/README.md
```diff
@@ -0,0 +1,59 @@
...
+## Step 1
+
+Alice picks a private key, a, greater than 1 and less than p.  Bob does
+the same to pick a private key b.
+
+## Step 2
+
+Alice calculates a public key A.
+
+    A = g**a mod p
```
FridaTveit: Has this README been generated using configlet? There are some discrepancies I think, for example this section should be a code block.

There a description in the section on adding a new exercise in our [contributing guide](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#adding-a-new-exercise) which explains how to generate the readme using configlet 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Before
+    public void setUp() {
+        diffieHellman = new DiffieHellman();
+    }
+
+    @Test
+    public void testPrivateKeyInRange() {
+        BigInteger primeP = BigInteger.valueOf(23);
+        List<BigInteger> privateKeys = new ArrayList<>();
+        for (int i = 0; i < 10; i++) privateKeys.add(diffieHellman.privateKey(primeP));
```
FridaTveit: Please use curly brackets for this for loop as well. That makes it more readable and it makes it less likely that someone modifying the code later will break it by trying to add a second line to the for loop 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private DiffieHellman diffieHellman;
+
+    @Before
+    public void setUp() {
+        diffieHellman = new DiffieHellman();
+    }
+
+    @Test
+    public void testPrivateKeyInRange() {
+        BigInteger primeP = BigInteger.valueOf(23);
```
FridaTveit: Why the name `primeP`? The name should describe what the variable is for as accurately as possible. If this variable is only meant to be some prime then just `prime` might be a more descriptive name? 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+        List<BigInteger> privateKeys = new ArrayList<>();
+        for (int i = 0; i < 10; i++) privateKeys.add(diffieHellman.privateKey(primeP));
+
+        for (BigInteger privateKey : privateKeys) {
+            assertTrue((privateKey.compareTo(BigInteger.ONE) >= 0) && (privateKey.compareTo(primeP) < 0));
```
FridaTveit: I would split this into two assertions. That makes it more readable and it makes it easy for the user to see exactly what's failing if the test fails 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
FridaTveit: There's a similar problem in the [robot-name tests](https://github.com/exercism/java/blob/master/exercises/robot-name/src/test/java/RobotTest.java). In that case the problem description says each robot should have a different name but most solutions will probably use randomness in which case there's a chance the names will be the same.

I think it's fine to keep this test, but I would only generate two keys and assert that they're not the same. That makes it a bit more obvious what you're testing 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+        for (int i = 0; i < 5; i++) privateKeys.add(diffieHellman.privateKey(primeP));
+
+        assertEquals(5, privateKeys.size());
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testPublicKeyCorrectlyCalculated() {
+        BigInteger primeP = BigInteger.valueOf(23);
+        BigInteger primeG = BigInteger.valueOf(5);
```
FridaTveit: Similarly to my earlier comment, I would name these something like `firstPrime` and `secondPrime` unless the letters `P` and `G` are significant? 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -0,0 +1,95 @@
...
+        BigInteger privateKey = BigInteger.valueOf(6);
+        BigInteger expected = BigInteger.valueOf(8);
+
+        assertEquals(expected, diffieHellman.publicKey(primeP, primeG, privateKey));
```
FridaTveit: Does `assertEquals` work well with `BigInteger`? Does it give a good error message if the test fails?

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: OK, I also think two keys will be enough 🙂

exercises/diffie-hellman/README.md
CRivasGomez: Ouch! Now i generated it using configlet 🙂
I noticed that configlet does't embed this section in a code block 🤔

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: I agree 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: True 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: I agree, just splitted into two assertions 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: Maybe ```primeA``` and ```primeB``` would be better names. Like ```publicKeyA```, ```publicKeyB```, ```privateKeyA```, ```privateKeyB```, ```secretA``` and ```secretB``` on the last test 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: Yes it does. It prints expected value and actual value in numeric format.
image

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -19,12 +19,15 @@ public void setUp() {
 
     @Test
     public void testPrivateKeyInRange() {
-        BigInteger primeP = BigInteger.valueOf(23);
+        BigInteger prime = BigInteger.valueOf(23);
         List<BigInteger> privateKeys = new ArrayList<>();
-        for (int i = 0; i < 10; i++) privateKeys.add(diffieHellman.privateKey(primeP));
+        for (int i = 0; i < 10; i++) {
+		privateKeys.add(diffieHellman.privateKey(prime));
```
FridaTveit: I think the indentation has gone slightly wrong here? This should be indented four spaces in from the `for` statement and the closing curly bracket should be indented the same as the `for` statement.

The second `assertTrue` statement in the next for loop has the same problem, as well as the for loop in `testPrivateKeyRandomlyGenerated` 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
```diff
@@ -33,61 +36,63 @@ public void testPrivateKeyInRange() {
     @Ignore("Remove to run test")
     @Test
     public void testPrivateKeyRandomlyGenerated() {
-        BigInteger primeP = BigInteger.valueOf(7919);
+        BigInteger prime = BigInteger.valueOf(7919);
         Set<BigInteger> privateKeys = new HashSet<>();
-        for (int i = 0; i < 5; i++) privateKeys.add(diffieHellman.privateKey(primeP));
+        for (int i = 0; i < 2; i++) {
+		privateKeys.add(diffieHellman.privateKey(prime));
+	}
```
FridaTveit: Since you're only doing two I don't think a loop and adding it to a set is necessary. I think it would be a bit more readable to just construct two private keys and assert that they're not equal (see the last [robot name test](https://github.com/exercism/java/blob/master/exercises/robot-name/src/test/java/RobotTest.java) for an example). That way it's more immediately obvious what you're doing without needing to know how a set works 🙂

exercises/diffie-hellman/README.md
FridaTveit: I see, I thought it was from the look of it in problem-specifications... But that's what configlet generates then it's probably right 🙂 Thanks for going to the trouble of doing that! 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: Sorry, I didn't notice it. Fixed! 🙂

exercises/diffie-hellman/src/test/java/DiffieHellmanTest.java
CRivasGomez: More clean, I totally agree 🙂

sjwarner-bp:
Thanks for smashing this @CRivasGomez :fireworks:

PR #1338, opened 2018-02-26
Add new exercise Error Handling

CRivasGomez:
Hello!
I added Error Handling exercise 🙂

As its [problem specification](https://github.com/exercism/problem-specifications/blob/master/exercises/error-handling/description.md) says I have to refer to the tests for Java track to see what's exactly required, but Java tests can't be generated, so i readed the tests for C# and Python and I come up with these.

---

Reviewer Resources:


config.json
```diff
@@ -395,6 +395,16 @@
       "unlocked_by": null,
       "uuid": "8dfc2f0d-1141-46e9-95e2-6f35ccf6f160"
     },
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "error-handling",
+      "topics": [
+        "exception_handling"
```
sjwarner-bp: Any other suggestions for possible topics? The list of topics can be found [here](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt)

exercises/error-handling/.meta/src/reference/java/ErrorHandling.java
```diff
@@ -0,0 +1,38 @@
+import java.util.Optional;
+
+class ErrorHandling {
+
+    public void handleErrorByThrowingException() throws Exception {
+        throw new Exception();
+    }
+
+    public void handleErrorByThrowingExceptionWithDetailMessage(String message) throws Exception {
```
sjwarner-bp: `public` methods could be changed to package-private 🙂

exercises/error-handling/README.md
```diff
@@ -0,0 +1,24 @@
+# Error Handling
```
sjwarner-bp: Has this been generated by `configlet`? I'd expect the metadata to show up

exercises/error-handling/src/main/java/ErrorHandling.java
```diff
@@ -0,0 +1,25 @@
+import java.util.Optional;
+
+class ErrorHandling {
+
+    public void handleErrorByThrowingException() throws Exception {
```
sjwarner-bp: Use of `public` in this file can also be changed to package private!

Also, I think we can remove `throws Exception` from these - it might give a little bit too much away to the users 😉 You should still be able to correctly throw `UnsupportedOperationException` even without this 🙂

sjwarner-bp:
Another thing to consider is that in Java we can make brand new exception classes - if it isn't too excessive (feel free to tell me it is overly-complex or out of specification), we could even ask a user to make their own custom exception!

@exercism/java what do you think of this suggestion?

config.json
CRivasGomez: Yes, ```optional_values```

exercises/error-handling/.meta/src/reference/java/ErrorHandling.java
CRivasGomez: True 🙂

exercises/error-handling/README.md
CRivasGomez: Yes, I generated it with ```configlet```.

exercises/error-handling/src/main/java/ErrorHandling.java
CRivasGomez: Ok 🙂

CRivasGomez:
As you say maybe is a bit excessive. Python and C# tracks don't ask the user to implement a new exception class.

exercises/error-handling/README.md
sjwarner-bp: Perfect! On a closer look, the normal metadata I would expect is not actually in the `problem-specifications` yet!

exercises/error-handling/.meta/src/reference/java/ErrorHandling.java
```diff
@@ -0,0 +1,38 @@
+import java.util.Optional;
+
+class ErrorHandling {
+
+    void handleErrorByThrowingException() throws Exception {
```
FridaTveit: In general it's considered bad practice to throw or catch just `Exception`. You should either use one of the built in [unchecked exceptions](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/) (e.g. `IllegalArgumentException`) or you should construct your own checked exception which extends `Exception`. This has been done for other exercises which don't just focus on error handling so I definitely don't think it's too much to ask in this exercise. On the contrary, I think it's an important thing to learn about error handling in Java 🙂 Since you've estimated this exercise as difficulty 4, you would need to provide starter implementation for the custom exception, so I don't think it should be that hard for the user to implement 🙂 What do you think? 🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
```diff
@@ -0,0 +1,66 @@
...
+public class ErrorHandlingTest {
+
+    private ErrorHandling errorHandling;
+
+    @Before
+    public void setUp() {
+        errorHandling = new ErrorHandling();
+    }
+
+    @Test(expected = Exception.class)
```
FridaTveit: We've decided to use [`ExpectedException`](https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) on this track instead of `@Test(expected = ...)`. This is because you can do more with `ExpectedException` such as assert what the error message is (which would be useful in your next test) 🙂 See [the phone number tests](https://github.com/exercism/java/blob/3631108c0d2fd245f603017fb86abdf462c7d10a/exercises/phone-number/src/test/java/PhoneNumberTest.java) for an example of how it's used 🙂

config.json
```diff
@@ -395,6 +395,17 @@
...
     },
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "error-handling",
+      "topics": [
+        "exception_handling",
+        "optional_values"
+      ],
+      "unlocked_by": "saddle-points",
```
FridaTveit: Why do you think this should be unlocked by `saddle-points`? I would expect it to be unlocked by an exercise that also deals with exceptions, for example `perfect-numbers`? 🙂

CRivasGomez: I think ```triangle``` would suit better (it also deals with exceptions) 🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
CRivasGomez: OK. I didn't notice that guideline 🙂 I will fix it.

exercises/error-handling/.meta/src/reference/java/ErrorHandling.java
CRivasGomez: Good point 🙂I will add a test case for throwing a custom exception and provide starter implementation 🙂

FridaTveit: Great, thank you! Those new methods look great! 🙂 

However, I would still disagree with keeping these methods which throw `Exception`. I realise I didn't explain why it's bad practice to throw or catch `Exception`, sorry about that! The reason is that `Exception` is the superclass of all exceptions. So if you say your method throws `Exception` then you don't give any details about the exception your method might throw, it could throw any exception at all. Even worse, when you catch `Exception` you catch all exceptions including runtime exceptions you might not want to handle. Therefore in Java you should always throw and catch specific exception classes that accurately describe the error that has occurred or the error that you're trying to handle. [This stackoverflow question](https://stackoverflow.com/questions/2416316/why-is-the-catchexception-almost-always-a-bad-idea) has some more information. Does that make sense? 🙂 

It might actually be a good idea to add some hints to this exercise (in a hints.md file in the .meta directory, see [flatten-array](https://github.com/exercism/java/blob/master/exercises/flatten-array/.meta/hints.md) for an example) explaining the difference between checked and unchecked exceptions (possibly linking to https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/), when you should use unchecked exceptions (possibly linking to https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html) and why you shouldn't throw or catch `Exception` (possibly linking to the stackoverflow question). Do you want to do that @CRivasGomez? That's absolutely not a requirement to get this exercise merged, if you think you have enough work with the rest of the exercise then I can just open a separate issue for adding a hints file 🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
```diff
@@ -0,0 +1,152 @@
...
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("This is the detail message.");
+        errorHandling.handleErrorByThrowingIllegalArgumentExceptionWithDetailMessage("This is the detail message.");
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testThrowCustomCheckedException() {
+        try {
+            errorHandling.handleErrorByThrowingCustomCheckedException();
```
FridaTveit: You should be able to deal with this in the same way as the other tests, i.e.
```
thrown.expect(CustomCheckedException.class);
```
The same goes for the next three tests🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.junit.Assert.assertTrue;
+
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+public class ErrorHandlingTest {
+
+    private ErrorHandling errorHandling;
+    private Set<Class> checkedExceptions;
```
FridaTveit: Why do you need these two sets? You've already provided the starter implementation for the exception you want them to throw (`CustomCheckedException` and `CustomUncheckedException`), so you know which class it should be 🙂

sjwarner-bp:
:+1: to adding some hints - I think that could be super useful

exercises/error-handling/src/test/java/ErrorHandlingTest.java
CRivasGomez: The purpose of the first test is to test the users capacity for throwing an Exception. The user will have to become aware about the need of adding ```throws Exception``` to the method signature. There is no way to bypass the declaration of ```throws Exception``` for the user as long as the test is expecting an instance of Exception.class. I could remove this test and replace it with other two ```testThrowCheckedException``` and ```testThrowUncheckedException```, what do your think? 🙂

I think is a good idea to add these hints 🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
CRivasGomez: I added them before providing the implementation and I forgot them 😅 Sets removed 🙂

exercises/error-handling/src/test/java/ErrorHandlingTest.java
FridaTveit: Yes, my problem is with having the test expect the user to throw just `Exception` and not a more specific class. Replacing this test with the other two sounds like a great idea 🙂 Though maybe the `IllegalArgumentException` tests should be before the custom exception tests? That way the user can first use an exception class which already exists before having to worry about custom exceptions. I don't think it matters too much though, either is fine 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,11 @@
+This exercise requires you to handle exceptions. An [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/definition.html) is an event, which occurs during the execution of a program, that disrupts the normal flow of the program's instructions.
+
+In Java, there are two types of exceptions: checked and unchecked exceptions.
+
+Checked exceptions are the exceptions that are checked at compile time.
```
FridaTveit: Maybe you could expand a bit on this since it might not be obvious to a beginner what "checked at compile time" means? 🙂 I think it's important to make clear:
1. The practical implications of a checked exception, i.e. that you have to declare it in the method signature of any method that can throw a checked exception, and handle or rethrow it when calling any method that can call a checked exception.
2. That this is because checked exceptions are meant to be handled at runtime, i.e. they are errors you can recover from.
3. Examples of where they are used. They're often used when a method can't return any valid result, for example a search method which hasn't found the item it was searching for. It's an alternative to returning `null` or an error code. A checked exception is better than those alternatives because it forces the user of the method to consider the error case.

Does that make sense? 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,11 @@
...
+
+Unchecked exceptions are the exceptions that are not checked at compiled time.
```
FridaTveit: Again it might be worth expanding on this a bit to explain what this means (that you don't have to add the exception to method signatures and you don't have to handle it) and explain what kind of errors it's meant to be used for (any that can't be handled at runtime, e.g. running out of memory) 🙂

exercises/error-handling/.meta/hints.md
CRivasGomez: OK 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+
+- [Unchecked Exceptions — The Controversy](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html)
+
+## Checked exceptions
+
+Checked exceptions are the exceptions that are checked at [compile time](https://en.wikipedia.org/wiki/Compile_time).
+
+### Practical implications
+
+You have to declare them in the [method signature](https://docs.oracle.com/javase/tutorial/java/javaOO/methods.html) of any method that can [throw](https://docs.oracle.com/javase/tutorial/essential/exceptions/throwing.html) a checked exception and handle or rethrow them when calling any method that can call a checked exception.
```
FridaTveit: I would probably say "that can throw a checked exception" rather than "call" 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+
+This is because checked exceptions are meant to be handled at runtime, ie., they are errors you can recover from.
```
FridaTveit: I think it should be "i.e." with a full stop between i and e, and without a comma after it 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+
+It's an alternative to returning null or a error code. A checked exception is better than those alternatives because it forces the user of the method to consider the error case.
+
+## Unchecked exceptions
+
+Unchecked exceptions are the exceptions that are not checked at [compile time](https://en.wikipedia.org/wiki/Compile_time).
+
+### Practical implications
+
+You don't have to declare them in the [method signature](https://docs.oracle.com/javase/tutorial/java/javaOO/methods.html) of methods that can [throw](https://docs.oracle.com/javase/tutorial/essential/exceptions/throwing.html) an unchecked exception and handle or rethrow them when calling any method that can call an unchecked exception.
```
FridaTveit: Again, I would say "throw" rather than "call" an exception 🙂

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Examples of where they are used
+
+Unchecked exceptions are mean to be used for any error than can't be handled at runtime, e.g., running out of memory.
```
FridaTveit: I don't think there should be a comma after "e.g." 🙂

exercises/error-handling/.meta/hints.md
CRivasGomez: Ups 🙂

exercises/error-handling/.meta/hints.md
CRivasGomez: Ok 🙂

PR #1348, opened 2018-02-27
Add new exercise Variable Length Quantity

jackattack24:
I've added a new exercise, Variable Length Quantity. 

---

Reviewer Resources:


config.json
```diff
@@ -422,6 +422,18 @@
       "unlocked_by": "hello-world",
       "uuid": "c3e89c7c-3a8a-4ddc-b653-9b0ff9e1d7d8"
     },
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "variable-length-quantity",
+      "topics": [
+        "lists",
```
sjwarner-bp: I like these topics!

Any others? The list can be found [here](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt)

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
```diff
@@ -0,0 +1,248 @@
...
+
+        assertEquals(expected, variableLengthQuantity.decode(bytes));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testCannotDecodeIncompleteSequence() {
+        List<Long> bytes = Arrays.asList(0xffL);
+
+        assertEquals(null, variableLengthQuantity.decode(bytes));
```
FridaTveit: Since this is essentially an error case I would maybe expect it to throw an exception instead of returning `null`?

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
```diff
@@ -0,0 +1,248 @@
...
+      List<Long> bytes = Arrays.asList(0x80L);
+
+      assertEquals(null, variableLengthQuantity.decode(bytes));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testDecodeMultipleBytes() {
+        List<String> expected = Arrays.asList("0x2000", "0x123456", "0xfffffff", "0x0", "0x3fff", "0x4000");
+        List<Long> bytes = Arrays.asList(0xc0L, 0x0L, 0xc8L, 0xe8L, 0x56L, 0xffL, 0xffL, 0xffL, 0x7fL, 0x0L, 0xffL, 0x7fL, 0x81L, 0x80L, 0x0L);
```
FridaTveit: This might be easier to read if it's broken up over several lines? 🙂

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
```diff
@@ -0,0 +1,248 @@
...
+        List<String> expected = Arrays.asList("0x81", "0x80", "0x0", "0xc8", "0xe8", "0x56");
+        List<Long> numbers = Arrays.asList(0x4000L, 0x123456L);
+
+        assertEquals(expected, variableLengthQuantity.encode(numbers));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testManyMultiByteValues() {
+        List<String> expected = Arrays.asList("0xc0", "0x0", "0xc8", "0xe8", "0x56", "0xff", "0xff", "0xff", "0x7f", "0x0", "0xff", "0x7f", "0x81", "0x80", "0x0");
```
FridaTveit: This might be easier to read if it's broken up over several lines? 🙂

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
sjwarner-bp: This is a really good point to raise actually - in problem specifications, most of the time when the canonical data expects `null`, this is just an indicator to language tracks that we should handle this error case in whichever way is most natural to the language.

config.json
jackattack24: How about conditionals and loops? And maybe exception handling once I refactor the error cases. Any other suggestions?

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
jackattack24: That's a good point; I'll go ahead and change that. Maybe something like this?
```java
throw new IllegalArgumentException("Invalid variable-length quantity encoding");
```

FridaTveit: Sounds good! 🙂

config.json
sjwarner-bp: I think those sound sensible - `exception-handling` especially if we refactor the tests! :slightly_smiling_face:

exercises/variable-length-quantity/src/test/java/VariableLengthQuantityTest.java
```diff
@@ -0,0 +1,262 @@
...
+        List<Long> numbers = Arrays.asList(0x4000L, 0x123456L);
+
+        assertEquals(expected, variableLengthQuantity.encode(numbers));
+    }
+
+    @Ignore("Remove to run test")
+    @Test
+    public void testManyMultiByteValues() {
+        List<String> expected = Arrays.asList("0xc0", "0x0", "0xc8", "0xe8",
+        "0x56", "0xff", "0xff", "0xff", "0x7f", "0x0", "0xff", "0x7f", "0x81",
```
FridaTveit: Very minor nitpick: this should be indented so that it's obvious that it's part of the same command as the previous line, if that makes sense 🙂 Common best practice is to put each argument on a new line if they don't all fit on one line. But that would make for a lot of lines in this case, so it's up to you whether you want to stick to that or not 🙂 

The same goes for the other places with line breaks 🙂

jackattack24:
Looks like the last build timed out.

PR #1351, opened 2018-03-06
go-counting: add new exercise

jssander:
This is what I have right now for a new exercise, go-counting. One issue I see right now is that in config.json it has the changes I made previously for the new two-bucket exercise. I'm not sure what to do about that. I'd like your feedback on the general code though and if it looks like a possible addition to the repository.
---

Reviewer Resources:


FridaTveit:
Thanks for adding this exericse @jssander! 😃 The conflicts are probably because your new-go-counting branch isn't up to date with the upstream master branch for this repo, so there are changes to config.json that you don't have. To fix this please [fetch the most recent upstream changes](https://help.github.com/articles/syncing-a-fork/) and merge them into your branch or rebase your branch. 

Does that make sense? Let me know if you need any help with that 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
```diff
@@ -0,0 +1,157 @@
+import java.awt.Point;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
+
+
+public class GoCounting {
```
FridaTveit: This can be [package private](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html) :)

exercises/go-counting/src/test/java/GoCountingTest.java
```diff
@@ -0,0 +1,175 @@
...
+		Set<Point> territory = new HashSet<>();
+		territory.add(new Point(0, 0));
+		territory.add(new Point(0, 1));
+		territory.add(new Point(1, 0));
+		
+		assertEquals(gocounting.BLACK, gocounting.getTerritoryOwner(0, 1));
+		assertEquals(territory, gocounting.getTerritory(0, 1));
+	}
+	
+	@Test
```
FridaTveit: As described in our [policies doc](https://github.com/exercism/java/blob/master/POLICIES.md#ignore-noninitial-tests) all but the first test should be ignored 🙂

exercises/go-counting/README.md
```diff
@@ -0,0 +1,50 @@
...
+In the game of go (also known as baduk, igo, cờ vây and wéiqí) points
+are gained by completely encircling empty intersections with your
+stones. The encircled intersections of a player are known as its
+territory.
+
+Write a function that determines the territory of each player. You may
+assume that any stones that have been stranded in enemy territory have
+already been taken off the board.
+
+Write a function that determines the territory which includes a specified coordinate.
```
FridaTveit: Did you generate this readme using configlet? It doesn't appear to be consistent with the one in [problem specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/go-counting/description.md). Specifically this line is not included.

We want our exercises to be consistent with the description and canonical data (if present) in problem specifications. That way the exercises are consistent across tracks 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
```diff
@@ -0,0 +1,157 @@
...
+		for(int i = 0; i < edges.size(); i++) {
+			if(owner != board[edges.get(i).x][edges.get(i).y]) {
+				owner = NONE;
+			}
+		}
+		
+		return owner;
+	}
+	
+	Set<Point> getTerritory(int x, int y) {
```
FridaTveit: From what I can tell from the [description of the exercise in problem specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/go-counting/description.md) you only want one method which given a player returns that player's territory. You don't appear to have a function like that from what I can tell? 🙂

config.json
```diff
@@ -612,6 +612,18 @@
       "unlocked_by": "rotational-cipher",
       "uuid": "d36ce010-210f-4e9a-9d6c-cb933e0a59af"
     },
+    {
+      "core": false,
+      "difficulty": 5,
```
FridaTveit: As described in our [policies doc](https://github.com/exercism/java/blob/master/POLICIES.md#add-hint-for-the-first-exercises-without-starter-implementation) all exercises with difficulty 5 should have a hint added to them since they are the first exercises without starter implementation 🙂 

However, this exercise is fairly complicated so I think it should maybe have a difficulty higher than 5. What do you think @jssander, @exercism/java? 🙂

config.json
```diff
@@ -612,6 +612,18 @@
...
+      "slug": "go-counting",
+      "topics": [
+        "algorithms",
+        "loops",
+        "conditionals"
+      ],
+      "unlocked_by": "triangle",
```
FridaTveit: What about adding "games" as a topic since go is a game? 🙂 

Also, scrabble-score might be a good exercise for this to be unlocked by since that is also a game and all the other game exercises are unlocked by it 🙂

exercises/go-counting/README.md
jssander: Previously I had submitted a PR to the problem specifications repository which includes an update to the readme. I think that is why it is a little different. I think soon my PR will be merged, but it hasn't been yet. You can take a look at my PR though if you like.

exercises/go-counting/.meta/src/reference/java/GoCounting.java
jssander: The getTerritories function does this in a similar manner to how it is done in the Python implementation. I think in C# it is probably similar as well. I based my java implementation on that.

config.json
jssander: What number do you think it should have?

exercises/go-counting/README.md
FridaTveit: Okay, that's great that you've suggested improvements to the problem specifications repo 🙂

However since changes to that repo affect all tracks, PRs can take a while to merge. So I think this reame should be consistent with what problem specifications currently has. 

Feel free to open a new PR to change it on this track once your PR in problem specifications has been merged though 🙂 And we do try to regularly sync our reames with the ones in problem specifications so it should be brought up to date eventually anyway 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
FridaTveit: Okay, that's fair enough. I think we should try to stick to the description in the readme as much as possible though, especially since there's no canonical data. That way it should be easier for people to understand what they have to do to implement the exercise. So in my opinion we should require the user to implement one function which takes a player (i.e. black or white I think in this game) and returns a set of points.

What do you think @jssander, @sjwarner-bp? 🙂

config.json
FridaTveit: Maybe 7 or 8? What do you think @sjwarner-bp? 🙂

exercises/go-counting/README.md
FridaTveit: Though I see now that [your PR to problem specifications](https://github.com/exercism/problem-specifications/pull/1195) also includes canonical data which is great! And it looks like it's close to being merged 🙂 So maybe it would be best to wait until that PR has been merged and then we can review this exericse while having the canonical data and the updated description in problem specifications? 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
sjwarner-bp: I think that seeing as the PR in problem-specifications contains canonical data and is looking close to being merged, we can review this again with a clearer idea of what we want to achieve as soon as it is merged 🎆

config.json
sjwarner-bp: I think this could be a quite difficult task... Maybe a 7? 🙂

Smarticles101: I think an 8 might be a little high, so I'd go with a 7

FridaTveit: Sounds like 7 would be a good idea then! 🙂

jssander:
Ok, all checks pass now and the PR in the problem specifications has been merged.

config.json
```diff
@@ -651,6 +651,19 @@
       "unlocked_by": "triangle",
       "uuid": "210bf628-b385-443b-8329-3483cc6e8d7e"
     },
+    {
+      "core": false,
+      "difficulty": 5,
```
FridaTveit: I still think this should be changed to 7 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
+
+class GoCounting {
+	int NONE = 0;
+	int WHITE = 1;
+	int BLACK = 2;
+	
+	int[][] board;
```
FridaTveit: These four variable could (and therefore should since we want to [minimize the accessibility of classes and members](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices)) be private 🙂 

If you want to use the `NONE`, `WHITE` and `BLACK` variables in the tests then please define them as a separate enum, also included as starter implementation (see [the robot-simulatot exercise](https://github.com/exercism/java/blob/master/exercises/robot-simulator/src/main/java/Orientation.java) as an example) 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
```diff
@@ -0,0 +1,156 @@
...
+	
+	GoCounting (String boardString) {
+		String[] lines = boardString.split("\n");
+		
+		board = new int[lines[0].length()][lines.length];
+		
+		for(int i = 0; i < lines.length; i++) {
```
FridaTveit: Minor nitpick: there should be a space after each `for` and after each `if`, e.g.
```
for (int i = 0; i < lines.length; i++) {
```
🙂

exercises/go-counting/src/test/java/GoCountingTest.java
```diff
@@ -0,0 +1,178 @@
...
+	public void blackCorner5x5BoardTest() {
+		GoCounting gocounting = new GoCounting(board5x5);
+		
+		Set<Point> territory = new HashSet<>();
+		territory.add(new Point(0, 0));
+		territory.add(new Point(0, 1));
+		territory.add(new Point(1, 0));
+		
+		assertEquals(gocounting.BLACK, gocounting.getTerritoryOwner(0, 1));
+		assertEquals(territory, gocounting.getTerritory(0, 1));
```
FridaTveit: Does this give a good error mesage when the test fails? That is, does it point out the exact difference between the two sets?

jssander: It seems to. For example if I remove a point from the expected set I get:

GoCountingTest > blackCorner5x5BoardTest FAILED
    java.lang.AssertionError: expected:<[java.awt.Point[x=0,y=0], java.awt.Point[x=0,y=1]]> but was:<[java.awt.Point[x=0,y=1], java.awt.Point[x=0,y=0], java.awt.Point[x=1,y=0]]>

config.json
jssander: Ok, but I think it is actually quite similar to two-bucket in difficulty in that all you have to do is a breadth first search.

jssander:
I made a new commit. I think with at least most of the changes you requested. I also commented on the assert error message issue. Let me know what you think.

config.json
FridaTveit: That's fair enough, they do look pretty similar. However, I think they're both quite a bit harder than most of the other difficulty 5 exercises, e.g. bob and pascals-triangle. Maybe the solution would be to also increase the difficulty of two-bucket? What do you think @jssander, @exercism/java? 🙂

exercises/go-counting/.meta/src/reference/java/GoCounting.java
```diff
@@ -5,55 +5,55 @@
 import java.util.Set;
 
 class GoCounting {
-	int NONE = 0;
-	int WHITE = 1;
-	int BLACK = 2;
+	//private int NONE = 0;
+	//private int WHITE = 1;
+	//private int BLACK = 2;
```
FridaTveit: This commented out code can be removed 🙂

config.json
jssander: I'm not as familiar with the other exercises as you guys, so I'll leave this up to you.

PR #1379, opened 2018-03-15
dominoes: add new exercise

jssander:
This is what I have so far for the dominoes exercise. I haven't done config.json or the readme yet.
---

Reviewer Resources:


jssander:
Now I've added changes to config.json and the readme, but there are some conflicts.

FridaTveit:
Hi @jssander, thanks for adding this exercise 🙂 Is your branch definitely up to date with exercism master? I would try fetching upstream master and either rebasing on that or merging it into your branch 🙂

jssander:
I rebased, but now when I try to push it says:

To https://github.com/jssander/java.git
 ! [rejected]        new-dominoes -> new-dominoes (non-fast-forward)
error: failed to push some refs to 'https://github.com/jssander/java.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

FridaTveit:
Sorry for taking a while to get back to you @jssander!

You're getting this error because when you rebase you change the history of your branch, therefore git is warning you that you're essentially rewriting history. If you're working on the same branch as other people then this is generally a bad idea, because you could accidentally undo someone else's work or just make it very confusing for them. However, in this case this is a branch only you're working on, so it's fine to rewrite history. To do that you just need to add the force param to the push command:
```
git push -f origin new-dominoes
```

Does that make sense? This [merge vs rebase tutorial](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) should have some more information about it 🙂

jssander:
Excellent, that worked and it passes the checks now.

exercises/dominoes/.meta/src/reference/java/Domino.java
```diff
@@ -0,0 +1,21 @@
+import java.util.Objects;
+
+class Domino {
+    int a;
+    int b;
```
FridaTveit: These should be private 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,75 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+
+class Dominoes {
+	
+	public static void main(String[] args) {
```
FridaTveit: As described in our [policies doc](https://github.com/exercism/java/blob/master/POLICIES.md#prefer-instance-methods), we prefer exercise implementations to have instance methods. It would be great if you could change this to use instance methods to be consistent with the other exercises 🙂 

Also, why do you need a main method which prints to the console?

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,75 @@
...
+		Domino[] dominoesList = {new Domino(1, 2), new Domino(3, 1), new Domino(2, 3)};
+		ArrayList<Domino> dominoes = new ArrayList<>(Arrays.asList(dominoesList));
+		
+		ArrayList<Domino> chain = formPartialChain(new ArrayList<Domino>(), dominoes);
+		
+		for(int i = 0; i < chain.size(); i++) {
```
FridaTveit: Please add whitespace after each `for` and each `if`, e.g. 
```
for (int i = 0; i < chain.size(); i++) {
```
to be consistent with the other code in this repo 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,75 @@
...
+			System.out.print("(" + chain.get(i).a + ", " + chain.get(i).b + "), ");
+		}
+	}
+	
+	static ArrayList<Domino> formChain(ArrayList<Domino> inputDominoes) {
+		if(inputDominoes.size() == 0) {
+			return null;
```
FridaTveit: Please throw an exception on error cases. That's generally considered best practice in Java since it forces the user to consider the error case 🙂

config.json
```diff
@@ -1102,6 +1102,19 @@
...
     },
+    {
+      "core": false,
+      "difficulty": 7,
+      "slug": "dominoes",
+      "topics": [
+        "games",
+        "algorithms",
+        "lists",
+        "recursion"
```
FridaTveit: Do you have to use recursion to solve this exercise? If not, then I would probably not add it to the topics. The topics should be suggestions of things the user might or must use to solve the exercise. It's normally better to use iteration than recursion in Java since it's more suitable to the language. So I would only add recusion to the topics if the exercise has to, or definitely should, be solved using recursion. Otherwise it can be seen as encouraging the users to implement a solution which isn't necessarily suitable for Java 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,189 @@
...
+		assertValidChain(dominoesArray, chain);
+	}
+	
+	private void assertValidChain(ArrayList<Domino> inputDominoes, ArrayList<Domino> outputDominoes) {
+		assertSameDominoes(inputDominoes, outputDominoes);
+		assertEndDominoesMatch(outputDominoes);
+		assertConsecutiveDominoes(outputDominoes);
+	}
+	
+	private void assertEndDominoesMatch(ArrayList<Domino> outputDominoes) {
```
FridaTveit: For this and the other two helper methods it might be useful to provide a useful error message to assertEquals? Otherwise the user will just get a message along the lines of `Expected 3 but got 4`. Which isn't very useful. The error message should contain all the details the user needs to see exactly what's gone wrong 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,189 @@
...
+	
+	@Test
+	public void emtpyInputEmptyOutputTest() {
+		Dominoes dominoes = new Dominoes();
+		
+		ArrayList<Domino> dominoesArray = new ArrayList<>();
+		
+		ArrayList<Domino> chain = dominoes.formChain(dominoesArray);
+		
+		assertNull(chain);
```
FridaTveit: These tests don't appear to match up exactly with the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/dominoes/canonical-data.json). The first few tests should just check a `canChain` method which should return a boolean 🙂

exercises/dominoes/.meta/src/reference/java/Domino.java
jssander: If they are private, won't I not be able to access them as I am currently doing?

exercises/dominoes/.meta/src/reference/java/Dominoes.java
jssander: The main method was just for testing. I'll remove that.

By instance methods you mean to make them not static, right?

exercises/dominoes/.meta/src/reference/java/Dominoes.java
jssander: Ok. I'm sorry I keep doing this. I guess old habits die hard.

config.json
jssander: Is it ok that I use recursion in the .meta reference solution?

Can you think of any other topics I could/should include?

exercises/dominoes/src/test/java/DominoesTest.java
jssander: In the canonical data file it says there are two levels for this exercise. Level 1 is to determine if a chain exists and level 2 is to find a possible chain if one exists. I guess I was trying to implement level 2.

exercises/dominoes/.meta/src/reference/java/Dominoes.java
jssander: Is it really an error though? It is just that there is no valid chain. Also the canonical data file doesn't seem to mention any exceptions. What do you think?

sjwarner-bp: Hi @jssander - the canonical data refrains from mentioning exceptions etc. to allow the track maintainers to suggest the best course of action in handling these cases :slightly_smiling_face:

config.json
sjwarner-bp: Solving the problem any way you see fit is completely fine :slightly_smiling_face: 

We should however refrain from using labels that aren't a core part of the exercise

exercises/dominoes/.meta/src/reference/java/Domino.java
FridaTveit: If you need to access them then you can add getter methods 🙂 That's generally thought of as better encapsulation (see [this stack overflow question](https://stackoverflow.com/questions/10407877/what-is-the-point-of-getters-and-setters)). You could also add setters but you shouldn't need them since you shouldn't need to change the value of a domino after creating it 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
FridaTveit: Yes, that's right 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
FridaTveit: That's fine, like you said it's hard to change habits like this 🙂 And it's easy to miss when you're focusing on the code logic 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
FridaTveit: Yes, and in Java the best course of action is usually to throw an exception. If you return null then the user of your method has to read your code (or your documentation if you have any) to find out if the code can return null, which is easy to forget. And people shouldn't need to look at the implementation of a method to be able to use it. Also, checking for null all the time can make code look messy. So I'm generally in favour of throwing an exception, even if it's not really an error but just a way of alerting the user to the fact that you haven't been able to return the normal output of the method 🙂 

Another solution could be to return an [`Optional`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html). I think they're also a bit messy and not as nice as exceptions in Java, but they're definitely better than returning null. So you could use an `Optional` if you prefer 🙂

config.json
FridaTveit: Yes, I agree with @sjwarner-bp, you can solve the exercise any way you want 🙂 That won't get shown to the users, it's just a way for us to check that the tests work 🙂 

If you decide to throw exceptions in your solution and expect exceptions in the tests then you could add `exception_handling` to the topics. If you decide to go with optionals then you could add `optional_values` to the topics 🙂 Apart from that I think the topics you have are a good selection 🙂

exercises/dominoes/src/test/java/DominoesTest.java
FridaTveit: I see, I didn't look at the canonical data carefully enough! Thanks for explaining 🙂 In which case what you're doing is fine 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
jssander: Ok. What kind of exception should it be?

jssander:
I think I've addressed everything, at least in part.

exercises/dominoes/.meta/src/reference/java/Dominoes.java
FridaTveit: It's up to you 🙂 You could use an exception which already exists, e.g. `IllegalArgumentException` seems like it would be quite suitable in this case. Or you could make a custom exception like in the triangle exercise 🙂

sjwarner-bp: I completely agree with @FridaTveit - you could also check out the [reference solution for error-handling](https://github.com/exercism/java/blob/master/exercises/error-handling/.meta/src/reference/java/ErrorHandling.java) which gives a handful of possibilities :slightly_smiling_face:

exercises/dominoes/.meta/src/reference/java/Domino.java
```diff
@@ -0,0 +1,30 @@
+import java.util.Objects;
+
+class Domino {
+    private int a;
```
FridaTveit: Since these are accessed using getLeft and getRight, naming them `left` and `right` might be more consistent 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,69 @@
...
+
+class Dominoes {
+	
+	ArrayList<Domino> formChain(ArrayList<Domino> inputDominoes) {
+		if (inputDominoes.size() == 0) {
+			throw new IllegalArgumentException("Empty input domino list.");
+		} else {
+			ArrayList<Domino> dominoChain = formPartialChain(new ArrayList<Domino>(), inputDominoes);
+			if(dominoChain == null) {
+				throw new IllegalArgumentException("No domino chain found.");
```
FridaTveit: Please add a whitespace after `if` 🙂 

Also I'm not sure if `IllegalArgumentException` is the correct exception to throw here. It's not really an illegal argument, it's just that it's not possible to form a chain with these dominoes. Maybe it would be better to use  a custom exception? What do you think @jssander, @exercism/java?

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,69 @@
...
+		
+		return true;
+	}
+	
+	ArrayList<Domino> formPartialChain(ArrayList<Domino> current, ArrayList<Domino> remaining) {
+		if (remaining.size() == 0) {
+			if (isValidChain(current)) {
+				return current;
+			} else {
+				return null;
```
FridaTveit: Even in internal methods I would prefer to throw exceptions over returning null. That makes the code neater and it means people maintaining your code don't have to remember to check for null 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,201 @@
...
+	@Test
+	public void emtpyInputEmptyOutputTest() {
+		Dominoes dominoes = new Dominoes();
+		
+		ArrayList<Domino> dominoesArray = new ArrayList<>();
+		
+		expectedException.expect(IllegalArgumentException.class);
+		expectedException.expectMessage("Empty input domino list.");
+		
+		ArrayList<Domino> chain = dominoes.formChain(dominoesArray);
```
FridaTveit: You don't have to assign to a variable here  (and in the other tests that expect an exception) since it's never used 🙂

And from looking at the canonical data it doesn't look like this should throw an error, it should just return an empty list I think?

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,69 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+
+class Dominoes {
+	
+	ArrayList<Domino> formChain(ArrayList<Domino> inputDominoes) {
```
FridaTveit: Do you need the input to be an `ArrayList`? I would keep it just a general `List` to make the tests a bit neater and to give the users a bit more freedom in their implementation 🙂 

And you can change the return type to just `List` as well 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,201 @@
...
+		
+		assertValidChain(dominoesArray, chain);
+	}
+	
+	@Ignore("Remove to run test")
+	@Test
+	public void separateLoopsTest() {
+		Dominoes dominoes = new Dominoes();
+		
+		Domino[] dominoesList = {new Domino(1, 2), new Domino(2, 3), new Domino(3, 1), new Domino(1, 1), new Domino(2, 2), new Domino(3, 3)};
```
FridaTveit: This would be a bit easier to read if you split it over multiple lines 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,201 @@
...
+	}
+	
+	private void assertValidChain(ArrayList<Domino> inputDominoes, ArrayList<Domino> outputDominoes) {
+		assertSameDominoes(inputDominoes, outputDominoes);
+		assertEndDominoesMatch(outputDominoes);
+		assertConsecutiveDominoes(outputDominoes);
+	}
+	
+	private void assertEndDominoesMatch(ArrayList<Domino> outputDominoes) {
+		assertEquals("The end dominoes are not the same.", outputDominoes.get(0).getLeft(), outputDominoes.get(outputDominoes.size() - 1).getRight());
```
FridaTveit: Maybe something like:
```
int leftValueOfFirstDomino = outputDominoes.get(0).getLeft();
int rightValueOfLastDomino = outputDominoes.get(outputDominoes.size() -1).getRight();
String errorMessage = "The left value of the first domino ("
    + leftValueOfFirstDomino
    + ") needs to match the right value of the last domino ("
    + rightValueOfLastDomino
    + ")."
```
That way the user gets told exactly which values are wrong and why 🙂 

When a test fails the user will want to know:
- What is their code returning and what should it be returning -> the normal assert error message will tell them this
- Why is the current value wrong -> your current error message will tell them this
- Which variables are wrong -> this is what's currently missing. For example in this case they'll know that the end dominoes are not the same and they'll know the values of both ends, but not which value is which end. They could figure it out by inspecting the test code and knowing how `assertEquals` formats its error messages, but they shouldn't need to do that 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,201 @@
...
+	}
+	
+	private void assertSameDominoes(ArrayList<Domino> inputDominoes, ArrayList<Domino> outputDominoes) {
+		assertEquals("The number of dominoes in the output chain doesn't match the number in the input.", inputDominoes.size(), outputDominoes.size());
```
FridaTveit: Similarly to the assert above, and the same goes for the ones below, it would be great if you could include exactly which values are wrong in your error message 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
sjwarner-bp: I'd agree that a custom exception would be more sensible! It's a completely valid argument I think - it's more a dominoes-specific rule, right? :slightly_smiling_face:

exercises/dominoes/.meta/src/reference/java/Dominoes.java
sjwarner-bp: :+1: Exceptions over null!! :fireworks:

jssander:
I think I've addressed all of your comments. Let me know what you think.

exercises/dominoes/.meta/src/reference/java/ChainNotFoundException.java
```diff
@@ -0,0 +1,5 @@
+class ChainNotFoundException extends RuntimeException {
```
FridaTveit: This should extend `Exception` not `RuntimeException`. When you extend `RuntimeException` it means your custom exception will be [unchecked](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/) so the user doesn't have to deal with it. We want the user to have to deal with the error, therefore our exception should be [checked](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/).

Unchecked exceptions are used for errors that you in most circumstances can't recover from, e.g. out of memory exception. It makes sense to not make the user deal with those kinds of error because usually the best you can do in those cases is just to let the program crash. I hope that makes sense, let me know if not 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
```diff
@@ -0,0 +1,67 @@
...
+			ArrayList<Domino> newChainA = new ArrayList<Domino>(current);
+			newChainA.add(remaining.get(i));
+			
+			ArrayList<Domino> newChainB = new ArrayList<Domino>(current);
+			newChainB.add(new Domino(remaining.get(i).getRight(), remaining.get(i).getLeft()));
+			
+			if (isValidPartialChain(newChainA)) {
+				try {
+					return formPartialChain(newChainA, newRemaining);
+				} catch (ChainNotFoundException e) {
```
FridaTveit: Why are you catching the exception here but leaving the catch block empty? Leaving a catch block empty is generally bad practice as it can easily hide errors and make your program hard to debug. If an exception has been thrown you should need to do something to handle it or just rethrow it. There might be a special case where you're sure you need to leave the catch block empty in which case you should leave a comment explaining why this is necessary.

In your case you don't need to catch the exception at all as far as I can tell, you just want the exception to propagate up and leave it up to the caller of this method to handle or rethrow it. Does that make sense? 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,228 @@
...
+		assertEquals("The output list should be empty.", 0, chain.size());
+	}
+	
+	@Ignore("Remove to run test")
+	@Test
+	public void singletonInputSingletonOutput() {
+		Dominoes dominoes = new Dominoes();
+		
+		Domino[] dominoesList = {new Domino(1, 1)};
+		ArrayList<Domino> dominoesArray = new ArrayList<>(Arrays.asList(dominoesList));
```
FridaTveit: You don't need to make this an `ArrayList` anymore which should simplify this code a bit 🙂

exercises/dominoes/src/test/java/DominoesTest.java
```diff
@@ -0,0 +1,228 @@
...
+		List<Domino> chain = dominoes.formChain(dominoesArray);
+		
+		assertValidChain(dominoesArray, chain);
+	}
+	
+	@Ignore("Remove to run test")
+	@Test
+	public void nineElementsTest() {
+		Dominoes dominoes = new Dominoes();
+		Domino[] dominoesList = {new Domino(1, 2), new Domino(5, 3), new Domino(3, 1), new Domino(1, 2), new Domino(2, 4), new Domino(1, 6), new Domino(2, 3), new Domino(3, 4), new Domino(5, 6)};
```
FridaTveit: It would be more readable if you add spread this over several lines like you did for the previous test 🙂

exercises/dominoes/.meta/src/reference/java/Dominoes.java
jssander: I've added a comment, but I think it needs an empty catch block. It is just supposed to continue on if there is no chain found in a particular branch of the tree. Am I right?

jssander:
I believe I have addressed everything. The only issue is the two empty catch blocks. I think they should be empty so I've added a comment. It is just supposed to continue on if an exception is thrown.

PR #1388, opened 2018-03-28
queen-attack: update tests(#1357)

### LINKED ISSUE #1357: queen-attack: update tests
The [queen-attack tests](https://github.com/exercism/java/blob/master/exercises/queen-attack/src/test/java/QueenAttackCalculatorTest.java) are slightly out of sync with the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/queen-attack/canonical-data.json). Please update the tests, the reference solution (if needed) and the version file (2.0.0 -> 2.1.0).

michaelspets:
Fix #1357

---

Reviewer Resources:


exercises/queen-attack/.meta/src/reference/java/Queen.java
```diff
@@ -0,0 +1,39 @@
+final class Queen {
+
+    private final int row;
+
+    private final int column;
+
+    Queen(final int row, final int column) throws IllegalArgumentException {
```
FridaTveit: The `throws IllegalArgumentException` isn't needed since `IllegalArgumentException` is an [unchecked exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html). I know this isn't your fault since it was like this before, so you don't have to fix it, we can open a separate issue for it. But if you do feel like fixing it now then that would be great 🙂

sjwarner-bp:
Thanks for getting around to fixing that extra bit so quickly too! :tada:

PR #1394, opened 2018-04-12
Update rectangles version file and add test. Closes #1372

### LINKED ISSUE #1372: rectangles: update version file (1.0.0 -> 1.1.0) and add test
Please update the version specified in the [rectangles](https://github.com/exercism/java/blob/master/exercises/rectangles/.meta/version) version file from 1.0.0 to 1.1.0.

Also please add the second test in the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/rectangles/canonical-data.json) which is missing from [our tests](https://github.com/exercism/java/blob/master/exercises/rectangles/src/test/java/RectangleCounterTest.java). This test should be exactly the same as the first test except that the input grid should be an array with one empty string (`new String[]{""}`) instead of an empty array of strings (`new String[]{}`).

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of [the canonical data file for that exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/rectangles/canonical-data.json).

tendaitt:
- Update version file to 1.1.0
- Add test to verify no rectangle when there are no columns
- Fix test naming to match canonical data

---

Reviewer Resources:


exercises/rectangles/src/test/java/RectangleCounterTest.java
```diff
@@ -14,12 +14,20 @@ public void setUp() {
...
 
         assertEquals(0, rectangleCounter.countRectangles(inputGrid));
     }
 
+    @Ignore("Remove to run test")
+    @Test
+    public void testInputWithNoColumnsContainsNoRectangles() {
+    	String[] inputGrid = new String[]{""};
+    	
+    	assertEquals(0, rectangleCounter.countRectangles(inputGrid));
```
FridaTveit: I think the indentation might be a bit off here? It should be indented with four spaces, same as the other tests 🙂

PR #1401, opened 2018-04-17
Update contributing.md

FridaTveit:
Make link to policies more general so it doesn't need updating every
time the policies change.

Add hint to only fix one issue per PR and do all changes on a branch.

---

Reviewer Resources:


sjwarner-bp:
Thanks for the great update :fireworks:

PR #1402, opened 2018-04-19
Fix canonical data check script

### LINKED ISSUE #1261: Improve canonical data check script
The `canonical_data_check.sh` script is sometimes reporting that exercises need to have their version files updated when they actually don't (e.g. it outputs `"needs update (v1.0.0 -> v1.0.0)."`). I'm guessing this is because one or the other of the files ends with a new line or has an extra whitespace or something.

This should hopefully be relatively easy to fix by adding some trimming of whitespace or similar to the script.

FridaTveit:
The script was reporting versions as different because of line endings. These line endings are now being removed before the version comparison.

I've tested it in WSL on windows. There's a chance it could cause problems on mac or linux (I don't think it would but you never know) so if anyone wants to try it to check then feel free 🙂

Fixes #1261. 

---

Reviewer Resources:

### DIFF
--- a/scripts/canonical_data_check.sh
+++ b/scripts/canonical_data_check.sh
@@ -93,7 +93,7 @@ do
         continue
     fi
 
-    track_data_version=$(cat $track_exercise_version_file_path)
+    track_data_version=$(cat $track_exercise_version_file_path | sed 's/\r$//')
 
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
Nice one @FridaTveit!

I'll pull it down over lunch and check it out on my mac :slightly_smiling_face:

Issue #1389, opened 2018-04-05
Do we need to make READMEs for earlier exercises simpler / add more hints?


sjwarner-bp:
Some feedback taken from the exercism/java Gitter room:

> I'm having trouble even starting the exercism course. I'm on the 2nd exercise and I just have no idea where to begin.

...

> I actually managed to get the first two tests to pass now. But yeah in the beginning it's a bit overwhelmingly contextless and I had no idea how to start
> That task has a huge README on TDD which is great, really helpful to understand how to work your way through these tasks in general. But at that stage I wasn't ready for the task itself. I needed a nudge in the right direction.
>
> As it was, it turned out I just needed to understand that if I simply deleted the exception and return the string in the test, that the test would pass.
>
> Once I understood what exactly the test was doing and what it was expecting, (and that I needed to return the value rather than print it), I've been pretty much fine ever since.

Maybe we need to append some hints to make it more obvious what is expected for brand new users (or make it more explicit that the tests are the place to consult for implementation questions?

What do we think @exercism/java? I'd be interested in hearing opinions and getting them implemented if we decide anything needs to be done! :slightly_smiling_face:

FridaTveit:
That's really useful to know, thanks @sjwarner-bp! 🙂 

The first exercise, hello-world, has [a long tutorial](https://github.com/exercism/java/blob/master/exercises/hello-world/TUTORIAL.md) which I think explains everything you have to do to make the tests pass (including deleting the exception and returning something instead). We could add that tutorial to more of the earlier exercises? Or maybe just reference it in the other exercises so that people know where to look for more instructions?

sjwarner-bp:
That's perfect! It's been so long since I did the track, I'd forgotten that was there.

I don't want to bloat the readmes unnecessarily because I think we shouldn't overwhelm users with too much info! Additionally, it was mentioned...

> That task has a huge README on TDD

... so we probably shouldn't add too much! I like the idea of referencing the tutorial :slightly_smiling_face:

FridaTveit:
Okay, should we reference the tutorial in e.g. the first three core exercises? Or all exercises with difficulty 2 or below? Or any other suggestions? 🙂

FridaTveit:
[...]

PR #1410, opened 2018-04-20
Add reference tutorial policy

FridaTveit:
We have decided that every exercise with diffulty less than 3 should reference the hello world tutorial to make it easier for people to solve those exercises. This has now been added to our policies doc.

We also now reference those policies in the contributing doc in the section for adding new exercises.

See issue #1389 for discussion.

---

Reviewer Resources:


sjwarner-bp:
Thanks for doing that so quickly!! :slightly_smiling_face:

Issue #1006, opened 2017-11-19
Add Linting to the Java Track


masters3d:
I would be interested in seeing linting support as part of the build process. 

I know Android provides some of these facilities. https://developer.android.com/studio/write/lint.html

I am not familiar with lint-mojo but it also seems to be an option. 
https://simpligility.github.io/android-maven-plugin/lint-mojo.html

I'd be happy to help implement this but I was hoping for some feedback.

masters3d:
Here are other options:
https://pmd.github.io/

Built in?
https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javac.html#BHCEECJF
```
-Xlint:all  
Enables all recommended warnings. In this release, enabling all available warnings is recommended.
```


Sample PR of somebody adding multiple Java Linters. I do not think we need all of these but maybe we would stick with the google code style?

https://github.com/commercetools/commercetools-sync-java/pull/32/files


I guess the goal would be to run the style check on the CI. 

We already have solutions to run it locally:

https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml
https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

masters3d:
https://github.com/checkstyle/checkstyle
https://github.com/jacoco/jacoco
https://github.com/spotbugs/spotbugs
https://github.com/pmd/pmd

Which one of these have you used before?

stkent:
Good thought!

I've used all of the above before (well, FindBugs in place of SpotBugs) and actually help maintain a Gradle plugin that simplifies applying several different checkers to Java (soon: +Kotlin) projects: https://github.com/btkelly/gnag/. It could work well here. My concern would be that linting would push `journey-test` times back up towards the Travis timeout limit. We could open a PR and find out!

stkent:
That plugin also has built-in PR commenting for feedback.

sjwarner-bp:
I was also thinking about this, as I'd recently contributed to the Python track and found linting to be a part of their process.

I haven't used any Java linters before, but I think it could be a nice way to enforce consistency across the track. I agree with @stkent that this could cause Travis builds to take too long, but we won't know if we don't try it! :slightly_smiling_face: 

On the subject of Travis, have we looked in to why the build failed in #1004 ? This might be the wrong place to discuss it, but should this be opened as an issue too?

stkent:
[...]

PR #1416, opened 2018-04-20
Update policies

FridaTveit:
Added some more best practices that I've seen come up on reviews quite often.

Added a section on not returning null since the canonical data often causes confusion about this.

Added a section on using ExpectedException since this was something we agreed a while ago but it's not documented anywhere.

---

Reviewer Resources:

### DIFF
--- a/POLICIES.md
+++ b/POLICIES.md
@@ -17,7 +17,7 @@ Our policies are not set-in-stone. They represent directions chosen at a point i
 
 | Track Event | Policies to review |
 |:------------|:-----------------|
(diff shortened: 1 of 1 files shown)


FridaTveit:
Softwire lets us have some charity holiday to do any kind of volunteer work, so I'm using mine to get some Exercism stuff done which is why I'm doing so much today 🙂

Using CheckStyle sounds like a good idea 🙂 I've only used it on one project so I'm not too familiar with it, but I'm sure it would be useful for people contributing. I've definitely found having a linter very useful when writing typescript 🙂

Are you familiar with CheckStyle? Do you want to have a go at setting it up? If not then I'm happy to have a go 🙂

sjwarner-bp:
That's cool! What a nice way to spend it :slightly_smiling_face: 

I've only used CheckStyle on one project too - it looks like we can write our own XML (or just adopt a popular one, for example Google or Sun) and configure it to work with our IDEs.

Similarly to you, I've had more experience with linters in Javascript and Typescript but it could be a cool way to enforce consistency - especially if we integrate it with our Travis builds :slightly_smiling_face: 

I'm happy to give it a look at some point if you want - I am a bit fuzzy on how this might be implemented but I can give it a go! Otherwise if you want to pick it up I don't mind either :slightly_smiling_face:

FridaTveit:
Sounds good! Well, we have an issue for it (#1006) so whoever gets round to it first can put a comment on there to say that they're working on it 🙂

PR #1429, opened 2018-04-20
grep: add exercise

FridaTveit:
* [Description](https://github.com/exercism/problem-specifications/blob/master/exercises/grep/description.md)
* [Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/grep/canonical-data.json)

---

Reviewer Resources:


exercises/grep/.meta/src/reference/java/GrepTool.java
```diff
@@ -0,0 +1,62 @@
...
+        String patternToCompare = isCaseInsensitive ? pattern.toLowerCase() : pattern;
+
+        boolean shouldMatchEntireLine = flags.contains("-x");
+        boolean matchesPattern = shouldMatchEntireLine
+            ? lineToCompare.equals(patternToCompare)
+            : lineToCompare.contains(patternToCompare);
+
+        boolean shouldInvertMatch = flags.contains("-v");
+        //noinspection SimplifiableConditionalExpression
+        return shouldInvertMatch ? !matchesPattern : matchesPattern;
```
FridaTveit: I know this can be simplified, but I think it's a lot easier to read like this. Happy to change it if people want though! 🙂

config.json
```diff
@@ -638,6 +638,20 @@
...
+      "difficulty": 5,
+      "slug": "grep",
+      "topics": [
+        "pattern_matching",
+        "strings",
+        "files",
+        "filtering",
+        "searching"
+      ],
+      "unlocked_by": "robot-name",
```
FridaTveit: Feel free to suggest a different difficulty or a different exercise for it to be unlocked by! 🙂 

I thought 5 might be a good difficulty since it has some complexity but not too much. And `robot-name` is also about strings and pattern matching (it has difficulty 5 as well) so I thought it might be a good fit for unlocked by 🙂

exercises/settings.gradle
```diff
@@ -101,3 +101,5 @@ include 'word-count'
 include 'word-search'
 include 'wordy'
 include 'zipper'
+include 'grep'
```
sjwarner-bp: Are these exercises meant to be alphabetised? :slightly_smiling_face:

exercises/grep/.meta/src/reference/java/GrepTool.java
sjwarner-bp: I'm happy to keep this as it is - I like readability! :slightly_smiling_face:

config.json
sjwarner-bp: I agree - this seems like a good fit!

exercises/settings.gradle
FridaTveit: Yes they are, thanks for reminding me! Will move `grep` so that it's in the correct order 🙂

sjwarner-bp:
Thanks for the new exercise @FridaTveit - great job! :fireworks:

Issue #1408, opened 2018-04-20
binary-search-tree: update tests and add version file


FridaTveit:
The [binary-search-tree tests](https://github.com/exercism/java/blob/master/exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java) should be updated to exactly match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/canonical-data.json).

Also a version file should be added to match the canonical data version. This file should be added in https://github.com/exercism/java/tree/master/exercises/binary-search-tree/.meta. See [the acronym version file](https://github.com/exercism/java/blob/master/exercises/acronym/.meta/version) for an example version file.

hgvanpariya:
@FridaTveit , please provide your review comments , the version should be 1.0.0 ?

FridaTveit:
Thanks @hgvanpariya! Yes, the version should be 1.0.0 🙂

PR #1431, opened 2018-04-22
binary-search-tree: update tests and add version file #1408

hgvanpariya:
binary-search-tree - update tests and add version file

---

Reviewer Resources:


hgvanpariya:
@sjwarner-bp , fixed the assertNull issue ... please provide your review comments ...

hgvanpariya:
@sjwarner-bp , but as per the specifications, left and right node should be null.

exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
```diff
@@ -122,26 +122,41 @@ public void sortsSingleElement() {
     public void sortsCollectionOfTwoIfSecondInsertedIsSmallerThanFirst() {
```
FridaTveit: The test above should insert `2`, not `4` I think according to the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/canonical-data.json) 🙂 

I also don't think the `createsComplexTree` test is quite correct according to the canonical data 🙂

hgvanpariya: @FridaTveit : as follow :

{
                    "description": "can sort if second number is smaller than first",
                    "property": "sortedData",
                    "input": {
                        "treeData": ["2", "1"]
                    },
                    "expected": ["1", "2"]
                }

Am I checking at wrong location ?
Will you please reverify ?

sjwarner-bp: I think the GitHub comment is just showing up in a weird place - @FridaTveit is right, the test `sortsSingleElement` should not use `4`, instead it should be `2`.

The test `sortsCollectionOfTwoIfSecondInsertedIsSmallerThanFirst` has been correctly updated :slightly_smiling_face:

hgvanpariya:
@FridaTveit , updated ... please let me know if now, I've updated at correct place ... 😀

exercises/binary-search-tree/src/test/java/BinarySearchTreeTest.java
```diff
@@ -108,10 +108,10 @@ public void createsComplexTree() {
     public void sortsSingleElement() {
```
FridaTveit: I still think the test above this one (`createsComplexTree`) doesn't quite match the canonical data 🙂

hgvanpariya: @sjwarner-bp

PR #1438, opened 2018-04-26
removed final keyword from test

jsertel:
per issue 1424

---

Reviewer Resources:


sjwarner-bp:
Thanks for the fix @jsertel :fireworks:

PR #1480, opened 2018-06-10
Reformat config files

kytrinyx:
This runs the configlet fmt command, which normalizes the
contents and ordering of keys in the track config and
maintainers config.

This will let us script changes to the config files without
having unnecessary noise in the diffs when submitting pull
requests.

See exercism/meta#95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,165 +1,174 @@
 {
+  "language": "Java",
   "active": true,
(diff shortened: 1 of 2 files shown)


sjwarner-bp:
Merged! :fireworks:

PR #1482, opened 2018-06-12
hello-world: set auto-approve to true

FridaTveit:
The first core exercise on the track should be auto-approved. This means that mentor approval isn't needed in order to move forward when the exercise has been completed.

Currently any hello-world exercises are auto-approved in the backend of v2 but this special handling might be removed later which is why this also needs to be set in config.json.

This fixes the last point in issue #1481.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,7 +17,8 @@
       "difficulty": 1,
       "topics": [
         "strings"
-      ]
+      ],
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
Thanks for the quick fix @FridaTveit 🎆

PR #1484, opened 2018-06-12
luhn: Updated tests and added version file

### LINKED ISSUE #1132: luhn: update tests and add version file
The luhn tests should be updated to exactly match the canonical data. Also a version file should be added to match the canonical data version.

c-thornton:
Updated the tests to match the canonica data in the problem specifications and added a version file. Fixes #1132.

---

Reviewer Resources:


exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -14,55 +14,55 @@ public void setUp() {
...
+    public void testSimpleValidSINReversedRemainsValid() {
+        assertTrue(luhnValidator.isValid("059"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testThatAValidCanadianSocialInsuranceNumberIsIdentifiedAsValidV1() {
-        assertTrue(luhnValidator.isValid("046 454 286"));
+    public void testSimpleValidSINReversedBecomesInvalid() {
+	assertTrue(luhnValidator.isValid("59"));
```
FridaTveit: Looks like the indentation is a bit wrong here? 🙂

exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -73,26 +73,38 @@ public void testThatStringContainingPunctuationIsInvalid() {
...
-        assertFalse(luhnValidator.isValid("34&"));
+    public void testStringContainingSymbolsInvalidJavaTrackSpecific() {
+        assertFalse(luhnValidator.isValid("85&"));
     }
 
     @Ignore("Remove to run test")
     @Test
-    public void testThatTheStringConsistingOfASpaceAndASingleZeroIsInvalid() {
+    public void testStringContainingSymbolsInvalid() {
+	assertFalse(luhnValidator.isValid("055£ 444$ 285"));
```
FridaTveit: I think the indentation here is a bit off as well? 🙂

exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -73,26 +73,38 @@ public void testThatStringContainingPunctuationIsInvalid() {
     */
     @Ignore("Remove to run test")
     @Test
-    public void testThatStringContainingSymbolsIsInvalid() {
-        assertFalse(luhnValidator.isValid("34&"));
+    public void testStringContainingSymbolsInvalidJavaTrackSpecific() {
```
FridaTveit: I think this should be moved so that it's the last test. Otherwise I think fixing this test might also fix later tests which we'd rather avoid. What do you think @c-thornton @sjwarner-bp? 🙂

exercises/luhn/src/test/java/LuhnValidatorTest.java
```diff
@@ -73,26 +73,38 @@ public void testThatStringContainingPunctuationIsInvalid() {
...
     @Test
-    public void testThatDoublingNineIsHandledCorrectly() {
+    public void testDigitNineConvertedToOutputNine() {
         assertTrue(luhnValidator.isValid("091"));
     }
 
+    @Ignore("Remove to run test")
+    @Test
+    public void testStringsWithNonDigitsInvalid() {
+	assertFalse(luhnValidator.isValid(":9"));
```
FridaTveit: I think the indentation is a bit off here as well? 🙂

c-thornton:
Good catch on the indentation @FridaTveit I will get that fixed tonight so that they match the rest.

exercises/luhn/src/test/java/LuhnValidatorTest.java
c-thornton: I am not sure if it should be moved to the bottom, but I see where you are coming from with the ordering. I kept all of the tests in the order that they appear in the canonical data and I slid this one in where there was a similar test in there.

Putting the tests in an order that will check in an increasing depth may provide better incremental improvements to the users program. I could put them in an order like the below if you think that would be better.
* testSingleDigitStringInvalid
* testSingleZeroIsInvalid
* testSingleSpaceWithZeroInvalid
* testMoreThanSingleZeroValid
* testDigitNineConvertedToOutputNine
* testSimpleValidSINReversedRemainsValid
* testSimpleValidSINReversedBecomesInvalid
* testValidCanadianSINValid
* testInvalidCanadianSINInvalid
* testInvalidCreditCardInvalid
* testStringsContainingNonDigitInvalid
* testStringContainingPunctuationInvalid
* testStringContainingSymbolsInvalid
* testStringsWithNonDigitsInvalid
* testStringContainingSymbolsInvalidJavaTrackSpecific

FridaTveit: In general we prefer to stick to the order of the test in the canonical data (like you've done in this PR) so that it's consistent across the different tracks 🙂 But if you think of any improvements to the order feel free to open an issue or a PR in [problem specifications](https://github.com/exercism/problem-specifications) to change the order in the canonical data 🙂 

However, since this specific test is just for the Java track we can decide where it should go 🙂 I think it would make sense for it to go at the end since it's testing for quite a specific bug and can therefore come after the tests for the core functionality. What do you think @c-thornton? 🙂

c-thornton: That makes sense to me as to why it should be at the end instead of in the middle where it currently is. I will move that down tonight while I am taking care of the indentation as well and then it should be ready to go. Thanks for the breakdown @FridaTveit I appreciate it. :smile:

PR #1492, opened 2018-07-13
Replace fetch command with download

PTrottier:
Update command to reflect the cli command options by providing an example command for downloading the next exercise.

---

Reviewer Resources:

### DIFF
--- a/exercises/hello-world/TUTORIAL.md
+++ b/exercises/hello-world/TUTORIAL.md
@@ -257,10 +257,10 @@ Cheers!
 
 ## Move on to the next exercise
 
-There are many more exercises you can practice with.  Grab the next one!
+There are many more exercises you can practice with.  Grab the next one! (example for `two-fer`)
 
 ```
-$ exercism fetch java
+$ exercism download --exercise=two-fer --track=java
 ```
 
 ## Review (and comment on) others' submissions to this exercise


FridaTveit:
Thanks for this update @PTrottier! 🙂

It looks like the build failed because `--config` is no longer an option for exercism cli:
```
+./exercism --config .journey-test.exercism.json configure --dir=/home/travis/build/exercism/java/build/exercism
Error: unknown flag: --config
```

Does anyone know if this has been removed permanently @PTrottier @sjwarner-bp @kytrinyx? Are there any other cli changes we should know about and should prepare the java track for now? Or was this just a temporary change due to the maintenance over the weekend? Sorry if I've missed any information about things we should be updating! 🙂

sjwarner-bp:
There's a whole new handful of cli-goodness for the nextercism project!

Here's the link to the download (just looking for some documentation around it). I've been trying to keep up to date with it, but absolutely rammed at the minute!

FridaTveit:
Yes, I thought there would be changes but I didn't think they were being used live already. But if they are then I guess we need to make changes! 🙂 Some documentation would be great, but I guess I can download the latest version and just play around with it to hopefully see what's changed. I'll open an issue for it 🙂 

Sorry @PTrottier your PR might be held up by fixing this issue for a little while, hope that's okay! 🙂 We'll merge your PR as soon as we've fixed the build 🙂

Issue #1493, opened 2018-07-16
Update docs and build with latest exercism cli changes


FridaTveit:
There have been some exercism cli changes recently that need to be reflected in our build process and documentation. To do this it might be easiest to [download the latest version of exercism cli](https://github.com/exercism/cli/releases) and play around with it to find out what the changes are.

sjwarner-bp:
There are some notes in the nextercism Slack group - I'll condense and copy over here :tada:

sjwarner-bp:
Basic guide:

1) Download it here: http://bit.ly/2tuyoI1
2) Grab your token from https://mentors-beta.exercism.io/my/settings
3) Configure it using `mentorcism configure --token=TOKEN`
4) Using the UI, find the exercise you want to try. Note the download link
5) Run the download command like `mentorcism download hello-world --track=ruby`
6) Make the tests pass and refactor
7) Submit using `mentorcism submit DIR` (edited)

This is from a month ago though - there doesn't seem to be much more detailed right this second. I guess now it is live that new info will begin to come out?

FridaTveit:
Thanks! 🙂 That release is tagged as `pre-release` and `alpha` though, so has it really been released? Is that really what our CI build will be using? I thougth this was the link to getting the latest exercism cli release: https://github.com/exercism/cli/releases?

Obviously it's good to be prepared anyway, even if it's just with a list of things that will need changing once we start using the new cli. But I'm still a bit confused as to whether that has happened yet or not...

sjwarner-bp:
Me either... The new version of the site went live a few days ago though, so I would have imaged that the new cli would be used now too? Not sure - I'll have to do some digging :slightly_smiling_face:

FridaTveit:
That makes sense, I'll try to look into it to 🙂 Working on the needed updates to journey-test now! 🙂

FridaTveit:
Okay the issues I'm currently seeing with the update:
- It doesn't seem to be possible to run `exercism configure` without a token anymore. My current workaround is to use my own token but that's probably not a good longterm solution.

- The script doesn't appear to be able to reach x-api. I get the error `The base API URL 'http://localhost:9292' cannot be reached`, even though just before then the script logs:
```
Puma 2.15.3 starting...
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://localhost:9292
+ echo 'x-api is running (pid=54961).'
x-api is running (pid=54961).
```

[...]

PR #1494, opened 2018-07-19
Make journey-test up to date with exercism cli updates

FridaTveit:
The exercism cli has been updated, therefore we need to change our build script so that it still works.

Problems we still need to fix:
- `exercism configure` can't be called without a token so I'm currently using my own token. We should find out what the standard procedure is for this, e.g. creating a token for the track or using some option that allows us not to use a token.

- I get the error `The base API URL 'http://localhost:9292' cannot be reached` even though xapi should be running. Apparently other tracks have had this issue as well.

- `exercism --config` doesn't appear to be a valid option anymore. I can't find the config file we were using so I'm hoping it's fine to not use it anymore?

See issue #1493 for details.

---

Reviewer Resources:

### DIFF
--- a/bin/journey-test.sh
+++ b/bin/journey-test.sh
@@ -3,341 +3,141 @@
 TRACK=java
 TRACK_REPO="$TRACK"
 TRACK_SRC_EXT="java"
-CPU_CORES=`getconf NPROCESSORS_ONLN 2>/dev/null || getconf _NPROCESSORS_ONLN 2>/dev/null || echo 1`
 EXERCISES_TO_SOLVE=$@
 
 on_exit() {
-  echo ">>> on_exit()"
-  if [[ "$xapi_pid" != "" ]] ; then
-    kill $xapi_pid
-    echo "stopped x-api (pid=${xapi_pid})"
-  fi
-  cd $EXECPATH
-  echo "<<< on_exit()"
+    echo ">>> on_exit()"
+    cd $EXECPATH
+    echo "<<< on_exit()"
 }
 
 assert_installed() {
-  local binary=$1
-  echo ">>> assert_installed(binary=\"${binary}\")"
-
-  if [[ "`which $binary`" == "" ]]; then
-    echo "${binary} not found; it is required to perform this test."
-    echo -e "Have you completed the setup instructions at https://github.com/exercism/${TRACK_REPO} ?\n"
-    echo "PATH=${PATH}"
-    echo "aborting."
-    exit 1
-  fi
-  echo "<<< assert_installed()"
-}
-
-assert_ruby_installed() {
-  local ruby_app_home="$1"
-  echo ">>> assert_ruby_installed(ruby_app_home=\"${ruby_app_home}\")"
-
-  pushd "${ruby_app_home}"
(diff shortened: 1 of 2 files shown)


bin/journey-test.sh
```diff
@@ -305,7 +304,6 @@ main() {
   local trackler_home="${build_path}/trackler"
   local exercism_home="${build_path}/exercism"
 
-  local exercism_configfile=".journey-test.exercism.json"
```
FridaTveit: Does anyone know anything about this file? I can't find it anywhere so I'm not sure if it's important or not...

FridaTveit:
Hm, the build passed... That's unexpected...

I think it's because `run-journey-test-from-ci` doesn't detect any changes to the exercises or `config.json` and therefore doesn't run `journey-test`. We probably want it to solve all exercises if any of the scripts have changed or if `config.json` or any `.gradle` file has changed?

FridaTveit:
I'm working on a change to `run-journey-test-from-ci` to fix this. I think it works, just need to test it a bit more! Will push it to this PR when I'm done 🙂

FridaTveit:
Done 🙂 I did a lot of testing (though of course I might have missed something) and I think the script works now 🙂 I've assumed that we want to run all exercises if a `.gradle` file or a `.sh` file or `config.json` has changed, let me know if anyone disagrees with that 🙂

sjwarner-bp:
Woohoo! Thanks for that @FridaTveit - I'll take a closer look when I get a chance.

Just as a quick update, Katrina has left the message:

> We have an open issue for figuring out a more robust set of tests for tracks. The journey tests are (as you discovered) not going to work with the new CLI directly. I'm sorry I forgot to think about this during the rush of the launch!

It's really great to have those changes in to allow us to run Travis on those changes too :tada:

FridaTveit:
I see, thanks for the update! 🙂

FridaTveit:
Given that there's not been much progress on https://github.com/exercism/exercism/issues/3955 the best temporary solution might be to just iterate over the exercises on `config.json` and not use the exercism cli. As long as we don't squash on merge for this PR we'll still have the nearly working commit to go back to if we decide to use exercism cli again. What do you think @sjwarner-bp, @exercism/java? 🙂

Smarticles101:
@FridaTveit that sounds like a good solution to me, even if it's just temporary.  What value does using the cli add exactly rather than just iterating over the exercises in `config.json` exactly?

Sorry I haven't been active recently

sjwarner-bp:
@FridaTveit I think that sounds like a good temp solution :slightly_smiling_face: 

Until there is a nice way to use the new CLI, I'm all for steering clear of it :slightly_smiling_face:

FridaTveit:
Thanks, you too! 🙂

FridaTveit:
This is ready to be reviewed now 🙂 

@Smarticles101, sorry I didn't see your comment earlier! Using the cli meant that the tests were more similar to what the users will do so should allow us to spot more issues. But I can't think of any examples in particular of issues I think using the cli would find that our now more traditional test runner wouldn't. Apart from spotting if an exercise doesn't have a uuid since that would be required to use the cli. But hopefully we would spot that during the review anyway 🙂

Issue #1471, opened 2018-06-05
series: update tests and add version file


FridaTveit:
The [series tests](https://github.com/exercism/java/blob/master/exercises/series/src/test/java/SeriesTest.java) should be updated to exactly match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/series/canonical-data.json). Also a version file should be added to match the canonical data version.

miranikaran01:
I would like to work on this one.

hgvanpariya:
@FridaTveit : Test case  `slice length cannot be zero` is invalid in usecase in https://github.com/exercism/problem-specifications/blob/master/exercises/series/canonical-data.json 

When the slice is Zero it is not throwing Exception, Can you please confirm ?

AndresAcuna:
If this is still open, I'd like to take care of it.

FridaTveit:
@AndresAcuna thanks for wanting to contribute! 🙂 However, @hgvanpariya has got an open pull request for this issue. @hgvanpariya are you still working on it? 🙂

hgvanpariya:
@FridaTveit : yes, I'll fix this ...

FridaTveit:
Great, thanks @hgvanpariya! 🙂

FridaTveit:
Fixed in PR #1504. Thanks @hgvanpariya 🙂

PR #1504, opened 2018-08-30
series: update tests and add version file #1471

hgvanpariya:
---

Reviewer Resources:

### DIFF
--- a/exercises/series/.meta/src/reference/java/Series.java
+++ b/exercises/series/.meta/src/reference/java/Series.java
@@ -7,22 +7,23 @@ import java.util.stream.Collectors;
 class Series {
 
     private final int digitsSize;
-    private List<Integer> digits;
+    private List<String> digits;
 
     Series(String string) {
-        this.digits = Arrays.stream(string.split(""))
-                        .map(Integer::parseInt)
-                        .collect(Collectors.toList());
-        this.digitsSize = this.digits.size();
+        this.digits = Arrays.stream(string.split("")).collect(Collectors.toList());
+        this.digitsSize = string.isEmpty()?0:this.digits.size();
     }
 
-    List<List<Integer>> slices(int num) {
+    List<List<String>> slices(int num) {
+        if (num <= 0) {
+            throw new IllegalArgumentException("Slice size is too small.");
+        }
         if (num > this.digitsSize) {
             throw new IllegalArgumentException("Slice size is too big.");
         }
         final int limit = this.digitsSize - num + 1;
-        List<List<Integer>> result = new ArrayList<>(limit);
-        List<Integer> tmp;
+        List<List<String>> result = new ArrayList<>(limit);
+        List<String> tmp;
         for (int i = 0; i < limit; i++) {
             tmp = this.digits.subList(i, i + num);
             result.add(tmp);
@@ -30,7 +31,7 @@ class Series {
         return result;
     }
 
-    List<Integer> getDigits() {
+    List<String> getDigits() {
         return digits;
     }
 
--- a/exercises/series/src/test/java/SeriesTest.java
+++ b/exercises/series/src/test/java/SeriesTest.java
@@ -17,145 +17,114 @@ public class SeriesTest {
     public ExpectedException expectedException = ExpectedException.none();
 
     @Test
-    public void hasDigitsShort() {
-        Series series = new Series("01234");
-        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4);
-        List<Integer> actual = series.getDigits();
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
+    public void slicesOfOneFromOne() {
+        Series series = new Series("1");
+        List<List<String>> expected = Arrays.asList(Arrays.asList("1"));
+        List<List<String>> actual = series.slices(1);
         assertEquals(expected, actual);
     }
-
+    
     @Ignore("Remove to run test")
     @Test
-    public void hasDigitsLong() {
-        Series series = new Series("0123456789");
-        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
-        List<Integer> actual = series.getDigits();
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
-    }
-
-    @Ignore("Remove to run test")
-    @Test
-    public void keepsTheDigitOrderIfReversed() {
-        Series series = new Series("9876543210");
-        List<Integer> expected = Arrays.asList(9, 8, 7, 6, 5, 4, 3, 2, 1, 0);
-        List<Integer> actual = series.getDigits();
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
-    }
-
-    @Ignore("Remove to run test")
-    @Test
-    public void keepsArbitraryDigitOrder() {
-        Series series = new Series("936923468");
-        List<Integer> expected = Arrays.asList(9, 3, 6, 9, 2, 3, 4, 6, 8);
-        List<Integer> actual = series.getDigits();
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
-    }
-
-    @Ignore("Remove to run test")
-    @Test
-    public void canSliceByOne() {
-        Series series = new Series("01234");
-        List<List<Integer>> expected = Arrays.asList(
-		 Collections.singletonList(0),
-                Collections.singletonList(1),
-                Collections.singletonList(2),
-                Collections.singletonList(3),
-                Collections.singletonList(4)
+    public void slicesOfOneFromTwo() {
+        Series series = new Series("12");
+        List<List<String>> expected = Arrays.asList(
+                Arrays.asList("1"), 
+                Arrays.asList("2")
         );
-        List<List<Integer>> actual = series.slices(1);
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
+        List<List<String>> actual = series.slices(1);
         assertEquals(expected, actual);
     }
 
     @Ignore("Remove to run test")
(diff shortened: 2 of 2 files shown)


hgvanpariya:
@FridaTveit : added comments on the commit instead of PR ....

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,117 @@
     public ExpectedException expectedException = ExpectedException.none();
 
     @Test
-    public void hasDigitsShort() {
-        Series series = new Series("01234");
```
FridaTveit: The indentation appears to have gone a bit wrong in this file. Please make sure everything is indented correctly, using four spaces for each indentation 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,117 @@
...
-        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4);
-        List<Integer> actual = series.getDigits();
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
+    public void slicesOfOneFromOne() {
+		Series series = new Series("1");
+		List<List<Integer>> expected = Arrays.asList(Collections.singletonList(1));
```
FridaTveit: From the canonical data it looks like the expected return type of `slices` is a list of strings instead of a list of lists of integers 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,117 @@
...
+		List<Integer> actual = series.slice(1);
```
FridaTveit: From the [reference implementation](https://github.com/exercism/java/blob/master/exercises/series/.meta/src/reference/java/Series.java) it looks like this method is called `slices` not `slice` which is why the build is failing 🙂

hgvanpariya:
@FridaTveit : Test case slice length cannot be zero is invalid in following file https://github.com/exercism/problem-specifications/blob/master/exercises/series/canonical-data.json

When the slice is Zero it is not throwing Exception, Can you please confirm ?

hgvanpariya:
@sjwarner-bp : If the slice is `0`, shouldn't it return the entire String instead of InvalidParameterException ?

sjwarner-bp:
I've not had a proper chance to read over the README for that exercise recently - is there any guidance over there? When I find a free moment I'll check it out myself 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,108 @@
...
-                Arrays.asList(3, 1, 0),
-                Arrays.asList(1, 0, 0),
-                Arrays.asList(0, 0, 1)
-        );
-        List<List<Integer>> actual = series.slices(3);
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
+    public void sliceLengthIsToolarge() {
+        expectedException.expect(IllegalArgumentException.class);
```
FridaTveit: You can check the exception message here as well if you want? 🙂 Some exercises do and some don't, it gives the user a bit of an extra challenge and encourages good practice for error handling. What do you think @hgvanpariya @sjwarner-bp? 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,108 @@
...
+        Series series = new Series("12345");
+        List<List<Integer>> actual = series.slices(6);
```
FridaTveit: You don't have to save this as a variable since it's never used 🙂

exercises/series/.meta/src/reference/java/Series.java
```diff
@@ -17,7 +17,7 @@
     }
 
     List<List<Integer>> slices(int num) {
-        if (num > this.digitsSize) {
+        if (num > this.digitsSize || num == 0) {
             throw new IllegalArgumentException("Slice size is too big.");
```
FridaTveit: I don't think this is the correct exception message for `num == 0`? 🙂

FridaTveit:
For the error handling I think throwing an exception would be the best way to handle it in Java. That's what's already implemented in the reference solution and it's what we use in most other exercises 🙂

sjwarner-bp:
Hi @hgvanpariya - how are you progressing with this one? 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,108 @@
...
-
-    @Ignore("Remove to run test")
-    @Test
-    public void hasDigitsLong() {
-        Series series = new Series("0123456789");
-        List<Integer> expected = Arrays.asList(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
-        List<Integer> actual = series.getDigits();
+    public void slicesOfOneFromOne() {
+        Series series = new Series("1");
+        List<List<Integer>> expected = Arrays.asList(Arrays.asList(1));
```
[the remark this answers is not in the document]
hgvanpariya: @FridaTveit : `series.slices()` returns `List<List<Integer>>` so if I convert it to the `List<List<String>>` then it won't match . Do you want me to change the return type of `series.slices()` also ?

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,108 @@
...
+        List<List<Integer>> actual = series.slices(1);
         assertNotNull(actual);
         assertFalse(actual.isEmpty());
         assertEquals(expected, actual);
```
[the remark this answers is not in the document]
hgvanpariya: Done

exercises/series/src/test/java/SeriesTest.java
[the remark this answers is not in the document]
FridaTveit: Yes, that's what I meant sorry! 🙂

exercises/series/.meta/src/reference/java/Series.java
hgvanpariya: done

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,108 @@
...
     }
-
-    @Ignore("Remove to run test")
```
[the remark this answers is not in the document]
hgvanpariya: Done

exercises/series/src/test/java/SeriesTest.java
hgvanpariya: Added Exception Message

exercises/series/.meta/src/reference/java/Series.java
```diff
@@ -7,30 +7,35 @@
...
 
     Series(String string) {
-        this.digits = Arrays.stream(string.split(""))
-                        .map(Integer::parseInt)
-                        .collect(Collectors.toList());
-        this.digitsSize = this.digits.size();
+        this.digits = Arrays.stream(string.split("")).collect(Collectors.toList());
+        if(!string.isEmpty()) {
+            this.digitsSize = this.digits.size();
+        }else {
```
FridaTveit: Very minor thing: there should be a space before `else` here 🙂 You could also write this using a ternary operator if you want:

```
this.digitsSize = string.isEmpty() ? this.digitz.size() : 0;
```

🙂

exercises/series/.meta/src/reference/java/Series.java
```diff
@@ -7,30 +7,35 @@
...
+            this.digitsSize = 0;
+        }
     }
 
-    List<List<Integer>> slices(int num) {
+    List<List<String>> slices(int num) {
```
FridaTveit: Sorry, I probably wasn't quite clear about what I was suggesting this method should return. Instead of returning a list of lists of strings, where each string represents a digit, you could return a list of strings where each string is a number. That way it would match the canonical data 🙂 Does that make sense? 

If you don't want to do that as part of this review, that's fine. I can open another issue for it 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,122 @@
...
-    public void keepsArbitraryDigitOrder() {
-        Series series = new Series("936923468");
-        List<Integer> expected = Arrays.asList(9, 3, 6, 9, 2, 3, 4, 6, 8);
-        List<Integer> actual = series.getDigits();
+    public void slicesOfTwo() {
+        Series series = new Series("35");
+        List<List<String>> expected = Arrays.asList(Arrays.asList("3", "5"));
+        List<List<String>> actual = series.slices(2);
         assertNotNull(actual);
         assertFalse(actual.isEmpty());
```
FridaTveit: As for the previous tests, the two assetions above can be removed I think. But I'm happy to open a new issue for that if you prefer 🙂 That applies for all the other tests below as well 🙂

exercises/series/src/test/java/SeriesTest.java
```diff
@@ -17,145 +17,122 @@
...
-        assertFalse(actual.isEmpty());
-        assertEquals(expected, actual);
+    public void slicesIncludeDuplicates() {
+        Series series = new Series("777777");
+        List<List<String>> expected = Arrays.asList(
+                Arrays.asList("7", "7", "7"),
+                Arrays.asList("7", "7", "7"), 
+                Arrays.asList("7", "7", "7"), 
+                Arrays.asList("7", "7", "7")
+		);
```
FridaTveit: This line and the ones below for this test appear to be indented four spaces too much 🙂

exercises/series/.meta/src/reference/java/Series.java
hgvanpariya: @FridaTveit , I got the request. And I want to fix that but, I would like to do it as separate issue.

hgvanpariya: Please let me know once you open an issue for that, I'll raise a new PR.

FridaTveit: I've openened issue #1598 🙂

PR #1512, opened 2018-09-18
Fix issue-1507

hugueschabot:
Remove 'generics' and use only List<Integer> for tests input.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -809,7 +809,6 @@
       "difficulty": 6,
       "topics": [
         "arrays",
-        "generics",
         "recursion",
         "searching"
       ]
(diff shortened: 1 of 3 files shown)


FridaTveit:
@hugueschabot There's also a hints file in this exercise which mentions how this exercise introduces generics. This hints file should be moved to the next core exercise which introduces generics, which I believe is `linked-list`. You can do that as part of this PR if you want or I can open a new issue for it. Let me know what you would prefer 🙂

FridaTveit:
Thanks @hugueschabot! 😄

One last thing, that hints file will be used to generate the readme. Therefore now that it's been moved the reames for binary-search and linked-list will both need to be regenerated. There are instructions for how to do that in our contributing guide: https://github.com/exercism/java/blob/master/CONTRIBUTING.md#updating-the-readmes.

Sorry, I should have mentioned that before!

It would be great if you could do that as part of this PR, but as before I can open a separate issue for it so let me know if you'd rather I do that! 🙂

PR #1551, opened 2018-10-01
update gradle to the latest release (v4.10.2)

vpondala:
Gradle 4.0.1 does not understand the new JDK versions (9 and beyond) well and build fails with an error (ex: Could not determine java version from '9.0.1')

---

Reviewer Resources:


sjwarner-bp:
I think this is a great idea - I did something similar for the Kotlin track a little while ago

PR #1555, opened 2018-10-01
phone-number: update version file (v1.4.0 -> v1.6.0) and update tests

### LINKED ISSUE #1535: phone-number: update version file (v1.4.0 -> v1.6.0) and update tests
Please update the version specified in the [phone-number](https://github.com/exercism/java/blob/master/exercises/phone-number/.meta/version) version file from 1.4.0 to 1.6.0.

Please also make sure the [phone-number tests](https://github.com/exercism/java/blob/master/exercises/phone-number/src/test/java/PhoneNumberTest.java) are up to date with the [canonical data for phone-number](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json). They should contain the same tests in the same order.

After any new tests have been added make sure that the [reference solution](https://github.com/exercism/java/blob/master/exercises/phone-number/.meta/src/reference/java/PhoneNumber.java) still passes the tests. If it doesn't then please update the reference solution so that it passes all the tests.

mirkoperillo:
proposal to fix  #1535 

---

Reviewer Resources:


exercises/phone-number/src/test/java/PhoneNumberTest.java
```diff
@@ -25,6 +25,7 @@ public void cleansTheNumber() {
         assertEquals(
                 expectedNumber, actualNumber
         );
+
```
FridaTveit: I don't think this extra empty line is needed? Not very important of course, but it would be great if you could remove it if you have time 🙂

mirkoperillo: done

Issue #1516, opened 2018-09-24
rna-transcription: update version file (v1.2.0 -> v1.3.0) and add one test


FridaTveit:
Please update the version specified in the [rna-transcription](https://github.com/exercism/java/blob/master/exercises/rna-transcription/.meta/version) version file from 1.2.0 to 1.3.0.

Please also add one test to the [rna-transcription tests](https://github.com/exercism/java/blob/master/exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java). This should be the test "Empty RNA sequence" as described in the [canonical data for rna-transcription](https://github.com/exercism/problem-specifications/blob/master/exercises/rna-transcription/canonical-data.json). See [this diff file](https://github.com/exercism/problem-specifications/commit/294c831e290670d2932365754317e92f09baacbc#diff-3910f220e27365c92ddd7517cae5c9ad) to make it more obvious which test has been added.

Please add the test as the first test in the test file.

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of the canonical data file for that exercise.

alis0nc:
I'm working on this issue.

FridaTveit:
Fixed in PR #1574. Thanks @alis0nc! 🙂

PR #1574, opened 2018-10-01
Update rna-transcription to 1.3.0 and add empty sequence test (#1516)

alis0nc:
Update ran-transcription to 1.3.0 and add the empty sequence test, as mentioned in #1516 

---

Reviewer Resources:


exercises/rna-transcription/src/test/java/RnaTranscriptionTest.java
```diff
@@ -4,6 +4,9 @@
 
 import static org.junit.Assert.assertEquals;
 
+/*
+ * Based on problem-specifications/exercises/rna-transcription/canonical-data.json 1.3.0
+ */
```
FridaTveit: This comment isn't needed. We've got the version file to tell us which version the tests and reference solution are based on 🙂

alis0nc: Removed!

PR #1578, opened 2018-10-02
atbash-cipher: update version file (v1.1.0 -> v1.2.0) and update tests

### LINKED ISSUE #1527: atbash-cipher: update version file (v1.1.0 -> v1.2.0) and update tests
Please update the version specified in the [atbash-cipher](https://github.com/exercism/java/blob/master/exercises/atbash-cipher/.meta/version) version file from 1.1.0 to 1.2.0.

Please also make sure the [atbash-cipher tests](https://github.com/exercism/java/blob/master/exercises/atbash-cipher/src/test/java/AtbashTest.java) are up to date with the [canonical data for atbash-cipher](https://github.com/exercism/problem-specifications/blob/master/exercises/atbash-cipher/canonical-data.json). They should contain the same tests in the same order.

After any new tests have been added make sure that the [reference solution](https://github.com/exercism/java/blob/master/exercises/atbash-cipher/.meta/src/reference/java/Atbash.java) still passes the tests. If it doesn't then please update the reference solution so that it passes all the tests.

mirkoperillo:
fix #1527 

---

Reviewer Resources:


exercises/atbash-cipher/src/test/java/AtbashTest.java
```diff
@@ -85,5 +85,18 @@ public void testDecodeAllTheLetters() {
         assertEquals("thequickbrownfoxjumpsoverthelazydog",
                 atbash.decode("gsvjf rxpyi ldmul cqfnk hlevi gsvoz abwlt"));
     }
-
+    
+    @Ignore("Remove to run test")
+    @Test
+    public void testDecodeWithTooManySpaces() {
+        assertEquals("exercism",
+                atbash.decode("vc vix    r hn"));
```
FridaTveit: Not very important but this could probably all fit on one line? We usually only split things over multiple lines if it's too long for one line. Exactly what people think "too long" is can vary, but in general 120 characters can be a good cut off point 🙂

mirkoperillo: done

PR #1592, opened 2018-10-04
Fixed #1590

### LINKED ISSUE #1590: grade-school: use assertEquals instead of assertThat
In [the grade-school tests](https://github.com/exercism/java/blob/master/exercises/grade-school/src/test/java/SchoolTest.java) `assertThat` is used for the test assertions. To be more consistent with the other exercises we should use `assertEquals` instead. For example:

```
assertThat(school.roster(), is(expected));
```

Should be changed to:

```
assertEquals(expected, school.roster());
```

yuval105:
Fixed #1590  
Used these 2 bash commands

`grep "assertThat" -r * -l | xargs sed -r "/assertThat[(].*, *is[(].*[)][)]/s/assertThat/assertEquals/" -i`  

`grep "assertEquals" -r * -l | xargs sed -r "/assertEquals[(].*, *is[(].*[)][)]/s/is[(](.*?)[)]/\1/" -i`

---

Reviewer Resources:

### DIFF
--- a/exercises/circular-buffer/src/test/java/CircularBufferTest.java
+++ b/exercises/circular-buffer/src/test/java/CircularBufferTest.java
@@ -26,7 +26,7 @@ public class CircularBufferTest {
         CircularBuffer<Integer> buffer = new CircularBuffer<>(1);
 
         buffer.write(1);
-        assertThat(buffer.read(), is(1));
+        assertEquals(buffer.read(), 1);
     }
 
     @Ignore("Remove to run test")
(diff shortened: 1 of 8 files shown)


FridaTveit:
Thanks for wanting to contribute @yuval105! 🙂 However, issue #1590 was only for changing grade-school. And @AndresAcuna had already commented on the ticket and opened a PR for the issue. Please check that no one else is working on the issue before opening a PR for it 🙂 

If you want to change this in any other exercises using `assertThat` then that's great! 🙂 But please only change one exercise per PR. That makes it easier for us to review 🙂 Does that make sense?

FridaTveit:
@yuval105, I'll close this PR for now. Feel free to open PRs for each exercise you've modified and thanks again for contributing! 🙂

PR #1593, opened 2018-10-05
Java code linting

mirkoperillo:
This is a early proposal for issue #1006.
I used `robot-simulator` exercise to this spike adding findbugs plugin to gradle.
Actually the build fails every findbugs best practise warning during the check task
Yo can try it adding something bad to `RobotTest.java`, i.e
```
List badList = new ArrayList();
badList.add(new Integer(12));
```
A HTML report file is produces in `build/reports/findbugs` folder.
The findbugs analyzes the main e test code (not the reference implementation for the moment)

Is this the right way ?

---

Reviewer Resources:


FridaTveit:
Thanks for this work @mirkoperillo! It's great that someone wants to help us with this 😄 

I think I agree with @sjwarner-bp that linting (e.g. something like CheckStyle) is what we'd prefer to focus on. And I think the code we'd prefer to focus on having checked is the test code and the reference solution. After all the user impemented solution will be reviewed by mentors and it might be confusing for beginners to have to deal with linting. Do you agree @sjwarner-bp @mirkoperillo? 🙂

sjwarner-bp:
That's exactly right @FridaTveit - I think that we only have to worry about things internal to this repo so we end up with a bit more consistency (and hopefully it will pick up common small issues in code to give maintainers better focus on the actual logic)! :slightly_smiling_face:

mirkoperillo:
@FridaTveit @sjwarner-bp now I understand what you need.
I'll move from findbugs to checkstyle plugin, I think to propose something to you in a couple of days, stay tuned :laughing:

mirkoperillo:
@Smarticles101 the plugin does both things, do you intend that the report should not be created ?

mirkoperillo:
@FridaTveit ok, I'll work on this.
Another thing: actually the configuration of the style is in the directory of the exercise, maybe it is better to move it in a more centralized place. Do you think that [config repo folder](https://github.com/exercism/java/tree/master/config) could be a good place ?

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+            <property name="applyToPrivate" value="false" />
+            <property name="format" value="^([A-Z][A-Z0-9]*(_[A-Z0-9]+)*|FLAG_.*)$" />
+            <message key="name.invalidPattern"
+                value="Variable ''{0}'' should be in ALL_CAPS (if it is a constant) or be private (otherwise)." />
+            <property name="severity" value="error" />
+        </module>
+
+        <module name="StaticVariableNameCheck">
+            <!-- Validates static, non-final fields against the supplied
+            expression "^[a-z][a-zA-Z0-9]*_?$". -->
```
FridaTveit: Why the underscore at the end? I wouldn't expect static variable names to end in an underscore?

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+            <property name="ignorePattern" value="${com.puppycrawl.tools.checkstyle.checks.sizes.LineLength.ignorePattern}"
+                default="^(package .*;\s*)|(import .*;\s*)|( *\* *https?://.*)$" />
+        </module>
+
+        <module name="LeftCurly">
+            <!-- Checks for placement of the left curly brace ('{'). -->
+            <property name="severity" value="error" />
+        </module>
+
+        <module name="RightCurly">
```
FridaTveit: Great, this rule should be really useful! 😄

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+              }
+              else
+            </pre>
+            -->
+            <property name="option" value="same" />
+            <property name="severity" value="error" />
+        </module>
+
+        <!-- Checks for braces around if and else blocks -->
+        <module name="NeedBraces">
```
FridaTveit: This should also be really useful! 😄

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+            <property name="illegalPattern" value="true" />
+        </module>
+
+        <!-- No printStackTrace -->
+        <module name="Regexp">
+            <!-- . matches any character, so we need to escape it and use \. to match dots. -->
+            <property name="format" value="e\.printStackTrace\(\)" />
+            <property name="illegalPattern" value="true" />
+        </module>
+    </module>
```
FridaTveit: I don't think we need the rules above? I don't think it'll come up, no one should need to add them. But I also don't think there should be rules against it. Thoughts @mirkoperillo @Smarticles101 @sjwarner-bp? 🙂

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+
+    <!--module name="SuppressionFilter">
+        <property name="file" value="suppressions.xml"/>
+    </module-->
```
FridaTveit: Please remove the rule above instead of commenting it out 🙂

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<!-- This is a checkstyle configuration file. For descriptions of
+what the following rules do, please see the checkstyle configuration
+page at http://checkstyle.sourceforge.net/config.html -->
+
+<module name="Checker">
+
+    <module name="FileTabCharacter">
```
FridaTveit: Is it possible to add a rule to check that files are indented correctly with four spaces? 🙂

mirkoperillo: I'll give a look to rule syntax, but I imagine that it's possible

mirkoperillo:
I have a proposal where to move `config/checkstyle` folder. 
What do you think about move it in exercises folder named as `.config` to show clearly that folder isn't an exercise ?

mirkoperillo:
Yes thinking in a lean mode you are right: it's a single file so there is no need of a folder, thinking in a more structured way a folder can be used for future shared plugins configurations. I think there isn't a best solution, you are the mantainers so you have to choose the more suitable solution for you :smile:

exercises/robot-simulator/config/checkstyle/checkstyle.xml
mirkoperillo: done

FridaTveit:
Good point 🙂 I think just a single file in the exercises folder for now and then we can always create a separate folder for it later if we realise we need more config files 🙂

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+        <property name="format" value="((//.*)|(\*.*))FIXME" />
+        <property name="message" value='TODO is preferred to FIXME.  e.g. "TODO: (ENG-123) -  Refactor when v2 is released."' />
+    </module>
+
+    <module name="RegexpSingleline">
+        <!-- Checks that TODOs are named with some basic formatting. Checks for the following pattern  TODO: (
+        -->
+        <property name="format" value="((//.*)|(\*.*))TODO[^: (]" />
+        <property name="message" value='All TODOs should be named.  e.g. "TODO: (ENG-123) - Refactor when v2 is released."' />
+    </module>
```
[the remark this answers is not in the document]
mirkoperillo: yes you are right, I think it is too strict to the example I found online

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+        <!-- required for SupressionCommentFilter and SuppressWithNearbyCommentFilter -->
+        <module name="FileContentsHolder" />
+
+        <!--
+
+        IMPORT CHECKS
+
+        -->
+
+        <module name="AvoidStarImport">
```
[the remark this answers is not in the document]
mirkoperillo: yes you are right

exercises/robot-simulator/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,369 @@
...
+            <property name="allowClassImports" value="false" />
+            <property name="severity" value="error" />
+        </module>
+
+        <module name="RedundantImport">
+            <!-- Checks for redundant import statements. -->
+            <property name="severity" value="error" />
+        </module>
+
+        <module name="ImportOrder">
```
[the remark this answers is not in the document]
mirkoperillo: done

exercises/robot-simulator/config/checkstyle/checkstyle.xml
mirkoperillo: me too

exercises/robot-simulator/config/checkstyle/checkstyle.xml
mirkoperillo: done

exercises/checkstyle.xml
```diff
@@ -0,0 +1,334 @@
...
+            <property name="format" value="System\.out\.println" />
+            <property name="illegalPattern" value="true" />
+        </module>
+
+        <!-- No System.err -->
+        <module name="Regexp">
+            <!-- . matches any character, so we need to escape it and use \. to match dots. -->
+            <property name="format" value="System\.err\.println" />
+            <property name="illegalPattern" value="true" />
+        </module>
```
FridaTveit: I think the two rules above can be removed as well? 🙂

FridaTveit:
Also is it possible at the moment for contributors to run just the checkstyle linter locally @mirkoperillo? If so then we should probably document how to do that somewhere 🙂 But I'm happy to open another issue for adding that functionality (if it doesn't exist) and documenting it 🙂

mirkoperillo:
The checkstyle plugin adds three tasks to gradle: `checkstyleMain`, `checkstyleTest`, `checkstyleSourceSet` and it adds a dependency to the check phase from these 3 tasks.
So everyone that can run `gradle check` run consequently the style verification.
Have I answered to your question ?

FridaTveit:
That's great, thanks @mirkoperillo! 🙂 In that case we should probably describe that somewhere in our [contributing doc](https://github.com/exercism/java/blob/master/CONTRIBUTING.md). Feel free to do that as part of this PR or I'm happy to open a separate issue for it 🙂

exercises/checkstyle.xml
mirkoperillo: done

exercises/flatten-array/src/test/java/FlattenerTest.java
```diff
@@ -18,123 +18,57 @@ public void setUp() {
...
-                '8',
-                100,
-                "four",
-                singletonList(
-                  singletonList(
-                    singletonList(
-                      50)))),
-              "-2")));
+            asList(0, '2', 2, "three", '8', 100, "four", 50, "-2"),
+            flattener.flatten(asList(0, '2', asList(asList(2, "three"), '8', 100, "four",
```
FridaTveit: When a line is too long a would prefer for either all the arguments to go on a new line if they fit, or for each argument to go on a new line. So in this case either: 
```
flattener.flatten(
    asList(0, '2', asList(asList(2, "three"), '8', 100, "four", singletonList(singletonList(singletonList(50)))), "-2")
)
```

or
```
flattener.flatten(
    asList(
        0,
        '2',
        asList(asList(2, "three"), '8', 100, "four", singletonList(singletonList(singletonList(50)))),
        "-2"
    )
)
```

Especially in an exercise like this when it's very easy to get confused about which arguments go with which functions. Does that make sense? 🙂

mirkoperillo: done

mirkoperillo: done

mirkoperillo: done

mirkoperillo: done

mirkoperillo:
@FridaTveit I have some issues with some validations. Actually `StaticVariableNameCheck` with regExp `^[a-z][a-zA-Z0-9]*$` doesn't accept values as

`private static double[] DISCOUNT_TIERS = {0, 5, 10, 20, 25};`

Validation wants something like this
`private static double[] discountTiers = {0, 5, 10, 20, 25};`
So actually checkstyle permits a variable name as DISCOUNT_TIERS only for constants (static final non-private variables), in all other cases something like this breaks the validation.
Some other violations:

LocalFinalVariableName validation
https://github.com/exercism/java/blob/e2c22df918dcf7f94fb12d277645ee81cdddaf9a/exercises/nucleotide-count/.meta/src/reference/java/NucleotideCounter.java#L29

MemberNameCheck
https://github.com/exercism/java/blob/e2c22df918dcf7f94fb12d277645ee81cdddaf9a/exercises/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java#L9

What I have to do ?

FridaTveit:
Good question @mirkoperillo 🙂 Is there any reason all those variables can't be `private final static` variables? 🙂

mirkoperillo:
Any reasons, I changed them as `private static final`

Last problem: one test in `ParallelLetterFrequencyTest` throws me OutOfMemory exception as you can see.
```
ParallelLetterFrequencyTest > multipleThreadsGetUsed FAILED
    java.lang.OutOfMemoryError: Java heap space
        at java.util.Arrays.copyOf(Arrays.java:3332)
        at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
        at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:448)
        at java.lang.StringBuilder.append(StringBuilder.java:136)
        at ParallelLetterFrequencyTest.multipleThreadsGetUsed(ParallelLetterFrequencyTest.java:150)
```

```
I fix it on my machine putting `maxHeapSize` on build.gradle of the test(actually I haven't pushed it). Do you have the same problem or it's caused only by my local settings ?

test {
  testLogging {
    exceptionFormat = 'full'
    events = ["passed", "failed", "skipped"]
    maxHeapSize = "2048m"
  }
}
```

mirkoperillo:
@FridaTveit ok, I'll merge master in a couple of days

mirkoperillo:
merged the master

PR #1594, opened 2018-10-08
zebra-puzzle: Add exercise

lemoncurry:
---

Reviewer Resources:


config.json
```diff
@@ -961,6 +961,18 @@
         "loops"
       ]
     },
+    {
+      "slug": "zebra-puzzle",
```
FridaTveit: If this has difficulty 6 then it should probably be placed before two-bucket? 🙂 But I think it might be more difficult than a 6? Especially since the exercise format offers very little guidance so they can't do it as "step by step" as most other exercises. So maybe it should be a 7 or 8? What do you think @sjwarner-bp @lemoncurry? 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static boolean isJustRightOf(int a, int b) {
+        return a - 1 == b;
+    }
+
+    private static boolean nextTo(int a, int b) {
+        return isJustRightOf(a, b) || isJustRightOf(b, a);
+    }
+
+    private void solve() {
```
FridaTveit: This method is a bit big. Would it be possible to extract some of it into individual methods? For example extract each method passed to `forEach`? 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
```diff
@@ -0,0 +1,108 @@
...
+        int[] houses = new int[]{1,2,3,4,5};
+        int first = 1;
+        int middle = 3;
+
+        perms(houses).forEach(colors -> {
+            int red = colors[0], green = colors[1], ivory = colors[2], yellow = colors[3], blue = colors[4];
```
FridaTveit: I'd prefer each variable to be declared on a new line. That makes it more obvious how many variables there are, and some of the lines below that declare many variables like this are very long 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
```diff
@@ -0,0 +1,108 @@
...
+
+            if (!isJustRightOf(green, ivory)) return;
```
FridaTveit: I prefer to always include curly brackets with `if` statements. Otherwise they're harder to spot and it's easier for people modifying the code later to accidentally introduce bugs (see our [policies doc](https://github.com/exercism/java/blob/master/POLICIES.md#adhere-to-best-practices)) 🙂

exercises/zebra-puzzle/src/test/java/ZebraPuzzleTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ZebraPuzzleTest {
+
+    @Test
+    public void residentWhoDrinksWater() {
+        ZebraPuzzle zebraPuzzle = new ZebraPuzzle();
+        assertEquals("Norwegian", zebraPuzzle.getWaterDrinker());
```
FridaTveit: An exercise which features a Norwegian! How exciting! 🙂

config.json
```diff
@@ -961,6 +961,18 @@
...
+      "uuid": "b1e2bd39-3f4b-44c1-b7e2-258d4ee241f8",
+      "core": false,
+      "unlocked_by": "saddle-points",
+      "difficulty": 6,
+      "topics": [
+        "arrays",
+        "conditionals",
```
FridaTveit: I'm not sure I would add `arrays` and `conditionals`. You don't necessarily need to use `arrays` and while you probably would use `conditionals` I'm not sure it's the main focus of the exercise. So I think it's fine just to have `logic` as the topic as it is very much an exercise about logic 🙂 What do you think @sjwarner-bp @lemoncurry? 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
lemoncurry: The structure of this solution relies on lambda capture. Restructuring that into multiple methods seems impractical to me, since there would be 5*5 parameters for the last method.
The solution was heavily influenced by the solution provided in the Go and Python track, which was adapted from a solution by Peter Norvig.

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
```diff
@@ -0,0 +1,156 @@
...
+    private static boolean isJustRightOf(int a, int b) {
+        return a - 1 == b;
+    }
+
+    private static boolean nextTo(int a, int b) {
+        return isJustRightOf(a, b) || isJustRightOf(b, a);
+    }
+
+    private void solve() {
+        int[] houses = new int[]{1, 2, 3, 4, 5};
```
FridaTveit: Sicne you're not actually modifying this, could it be a static variable? 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
FridaTveit: I don't think it'll be quite as bad as that 🙂 From a little playing around I think it looks like the last method will need at most 11 parameters. Not ideal but I think I might still prefer it to having one massive method. And it could be decreased by storing `houses` as a static variable and passing parameters as arrays, e.g. `int[] colorPermutation` instead of `int red, int green, int yellow...` if that makes sense? 🙂 

So I still think it would be a nice improvement. But you absolutely don't have to do it if you don't want to! 🙂 If you'd rather have it merged as is then that's fine, I might open an issue for it which can be picked up by anyone later 🙂 Does that sound okay? 🙂

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
```diff
@@ -0,0 +1,156 @@
...
+        int first = 1;
+        int middle = 3;
+
+        perms(houses).forEach(colors -> {
```
FridaTveit: Since this is iterating through the permutations maybe `colorPermutation` might be a more descriptive name? 🙂 Same for the other iterations through permutations 🙂

lemoncurry:
@FridaTveit Thank you for your great suggestions, I will work on it 😃

exercises/zebra-puzzle/.meta/src/reference/java/ZebraPuzzle.java
FridaTveit: It looks great, thank you! 😄 

I see what you mean, jumping from method to method can make it harder to see the flow. However, I also think it's very hard to take in and understand a really long method. When it's broken up into smaller methods then you can try to understand each piece individually and then use that to understand what it all does together. I find that easier, but I guess it varies from person to person 🙂 

And as the solution to this exercise is a bit complicated (it being a logic puzzle) I guess no matter what we do the code will probably be a little bit difficult to understand 🙂

PR #1600, opened 2018-10-11
affine-cipher: Add new exercise

lemoncurry:
---

Reviewer Resources:


exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testEncodeYes() {
+        AffineCipher affineCipher = new AffineCipher();
+
+        assertEquals("xbt", affineCipher.encode("yes", 5, 7)
+        );
```
FridaTveit: I think this should be on the previous line? 🙂

exercises/affine-cipher/.meta/src/reference/java/AffineCipher.java
```diff
@@ -0,0 +1,69 @@
...
+                                    ALPHABET_START_UNICODE + Math.floorMod(inv * (orig - keyB), ALPHABET_SIZE)
+                            );
+                        }
+                    }
+                });
+
+        return stringBuilder.toString();
+    }
+
+    String encode(String plainMessage, int keyA, int keyB) {
```
FridaTveit: I would put these two methods first  since they're package private 🙂 Generally I put more accessible methods first so that they're the first thing you see when opening the file 🙂

exercises/affine-cipher/.meta/src/reference/java/AffineCipher.java
```diff
@@ -0,0 +1,69 @@
...
+        List<String> words = new ArrayList<>();
+
+        for (int i = 0; i < text.length(); i += GROUP_SIZE) {
+            words.add(text.substring(i, Math.min(text.length(), i + GROUP_SIZE)));
+        }
+
+        return String.join(" ", words);
+    }
+
+    private static String translate(String text, int keyA, int keyB, int mode) {
```
FridaTveit: It looks like you're using `mode` essentially as a switch that should only have one of two values? In which case maybe using a boolean would be better, e.g. something like `shouldDecode`? Or you could use an enum? Not very important, but if there should only be two possible values then it's preferrable to use a type with only two values 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Ignore("Remove to run test.")
```
FridaTveit: Not very important, but to be consistent with the other exercises there shouldn't be a full stop after "test" 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+    @Test
+    public void testEncodeAllTheLetters() {
+        AffineCipher affineCipher = new AffineCipher();
+        assertEquals("swxtj npvyk lruol iejdc blaxk swxmh qzglf",
+                affineCipher.encode("The quick brown fox jumps over the lazy dog.", 17, 33));
+    }
+
+    @Ignore("Remove to run test.")
+    @Test
+    public void testEncodeThrowsMeaningfulException() throws IllegalArgumentException {
```
FridaTveit: Since `IllegalArgumentException` is an unchecked exception you don't need to include `throws` here 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class AffineCipherTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testEncodeYes() {
+        AffineCipher affineCipher = new AffineCipher();
```
FridaTveit: From what I can tell this is the same for all tests? So it might be nice to pull it out into a `@Before` block? 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+    public void testDecodeNumbers() {
+        AffineCipher affineCipher = new AffineCipher();
+        assertEquals("testing123testing", affineCipher.decode("odpoz ub123 odpoz ub", 25, 7));
+    }
+
+    @Ignore("Remove to run test.")
+    @Test
+    public void testDecodeAllTheLetters() {
+        AffineCipher affineCipher = new AffineCipher();
+        assertEquals("thequickbrownfoxjumpsoverthelazydog", affineCipher.decode("swxtj npvyk lruol iejdc blaxk swxmh qzglf", 17, 33));
```
FridaTveit: This line is a bit long, maybe break it up over two lines? 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Ignore("Remove to run test.")
+    @Test
+    public void testDecodeWithNoSpaces() {
+        AffineCipher affineCipher = new AffineCipher();
+        assertEquals("thequickbrownfoxjumpsoverthelazydog", affineCipher.decode("swxtjnpvyklruoliejdcblaxkswxmhqzglf", 17, 33));
```
FridaTveit: This line is also a bit long 🙂

exercises/affine-cipher/.meta/src/reference/java/AffineCipher.java
lemoncurry: Great suggestion, thank you!

FridaTveit: 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -7,130 +8,123 @@
 
 public class AffineCipherTest {
 
+    private AffineCipher affineCipher;
+
+    @Before
+    public void genAffineCipfer() {
```
FridaTveit: Very minor thing, should this be `genAfficeCipher`? 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -0,0 +1,130 @@
...
+        affineCipher = new AffineCipher();
+    }
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testEncodeYes() {
+        assertEquals("xbt", affineCipher.encode("yes", 5, 7)
+        );
```
FridaTveit: I still think this should be on the previous line? So that it becomes: 
```
assertEquals("xbt", affineCipher.encode("yes", 5, 7));
```
What do you think? 🙂

exercises/affine-cipher/src/test/java/AffineCipherTest.java
lemoncurry: yes, of course, thank you! :smile:

exercises/affine-cipher/src/test/java/AffineCipherTest.java
lemoncurry: yes, seems I didn't exactly understand your previous comment. Of course I will correct it! 😃

PR #1605, opened 2018-10-23
darts: add new exercise

lemoncurry:
@wip ready for review

---

Reviewer Resources:


lemoncurry:
Exercise is ready for review. Removed WIP from title of PR, but is still showing as WIP in the checks?

FridaTveit:
Hm, that's strange @lemoncurry. I tried adding "@wip ready for review" to the pull request body as according to the [wip app docs](https://github.com/wip/app) that should clear it, but it didn't help. Any ideas @sjwarner-bp @Smarticles101?

exercises/darts/.meta/src/reference/java/Darts.java
```diff
@@ -0,0 +1,32 @@
+class Darts {
+
+    private static final int R_CIRCLE_OUTER = 10;
```
FridaTveit: Does `R` stand for radius? In which case I would prefer to use the whole word to make it clearer 🙂

config.json
```diff
@@ -41,6 +41,17 @@
...
     },
+    {
+      "slug": "darts",
+      "uuid": "4d400a44-b190-4a0c-affb-99fad8ea18da",
+      "core": false,
+      "unlocked_by": "two-fer",
+      "difficulty": 1,
+      "topics": [
+        "games",
+        "math"
```
FridaTveit: According to the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) I think this should be `mathematics` 🙂

config.json
```diff
@@ -41,6 +41,17 @@
         "strings"
       ]
     },
+    {
+      "slug": "darts",
+      "uuid": "4d400a44-b190-4a0c-affb-99fad8ea18da",
+      "core": false,
+      "unlocked_by": "two-fer",
+      "difficulty": 1,
```
FridaTveit: I think this involves a bit more logic than the other difficulty 1 exercises so maybe it should have difficulty 2? 🙂 

And since it has difficulty 4 or lower it should have a starter implementation (see the [POLICIES doc](https://github.com/exercism/java/blob/master/POLICIES.md#starter-implementations)) 🙂

exercises/darts/.meta/src/reference/java/Darts.java
FridaTveit: Does R stand for radius? In which case I would prefer to use the whole word to make it clearer 🙂

config.json
lemoncurry: According to #1499 this was changed from 'mathematics' to 'math' in August and 'math' is used throughout the config file. So I would keep it if there have not been any new changes since then and suggest to adapt the topics list accordingly. I could open an issue on the resp. rep after checking there about this decision.

config.json
lemoncurry: Thank you, I will add the starter implementation and change the difficulty level.

exercises/darts/.meta/src/reference/java/Darts.java
lemoncurry: Yes of course I can add that to the field name.

config.json
lemoncurry: Oh, I didn't see this list of exercises that should have math topics. I will check for it in the problem-spec repo. 
Of course I would then remove the math topic from this new exercise. The darts exercise was added just a few days ago to the problem-spec repo.

FridaTveit: I see, I didn't realise it was so new 🙂 I don't think there's a definite list, there was just a list they made at the time of raising https://github.com/exercism/exercism/issues/4110. So from my understanding of that we can add maths as a topic to new exercises but we should avoid doing so unless maths is really important to the exercise as it might put people off 🙂

exercises/darts/src/main/java/Darts.java
```diff
@@ -0,0 +1,14 @@
...
+
+    Darts(double x, double y) {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    int score() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
+    }
+
+    double distanceDartToOrigin() {
```
FridaTveit: You don't need to add this method since it's private 🙂 You only need to add starter implementation for any methods used in the tests. Does that make sense? 🙂

lemoncurry: I see. Will change :smile:

PR #1612, opened 2018-10-31
sum-of-multiples: Update to v1.5.0 of canonical-data

lemoncurry:
---

Reviewer Resources:


lemoncurry:
Sry did not have the linting option yet in my copy, will fix it.

exercises/sum-of-multiples/src/test/java/SumOfMultiplesTest.java
```diff
@@ -6,11 +6,11 @@
 public class SumOfMultiplesTest {
 
     @Test
-    public void testSumOfMultiplesOf3and5UpToOne() {
+    public void testNoMultiplesWithinLimit() {
```
FridaTveit: Just for future reference, our test names don't have to correspond exactly with the canonical data. We only want the test values and test order to be the same. And we want the names to be good descriptive names for the test 🙂 The names you've changed them to seem fine though 🙂

exercises/sum-of-multiples/src/test/java/SumOfMultiplesTest.java
```diff
@@ -162,11 +162,40 @@ public void testSumOfMultiplesOfOneUpToOneHundred() {
...
+    @Test
+    public void testSumOfMultiplesOfZeroIsZero() {
+
+        int[] set = {
+            0
+        };
+        int output = new SumOfMultiples(1, set).getSum();
+        assertEquals(0, output);
+
+    }
```
FridaTveit: It looks like there's now been added another test after this one `the factor 0 does not affect the sum of multiples of other factors` and the version number has been bumped to 1.5.0. It's up to you if you want to add that as part of this PR or not! 🙂

lemoncurry: I will add this new test and update to version 1.5.0

exercises/sum-of-multiples/src/test/java/SumOfMultiplesTest.java
lemoncurry: The reason why I changed the method names was that the change from version 1.4.0->1.4.1 only included changes in the description for all tests so it is clearer. that's why I thought it would be good to mirror that effort in the test method names.

FridaTveit: I see, that makes sense 🙂

exercises/sum-of-multiples/.meta/src/reference/java/SumOfMultiples.java
```diff
@@ -4,21 +4,18 @@
 public class SumOfMultiples {
     private final int sum;
 
-    public SumOfMultiples(int number, int[] set) {
-        sum = calculateSum(number, set);
+    public SumOfMultiples(int limit, int[] set) {
```
FridaTveit: Much better variable name, thanks! 🙂

PR #1613, opened 2018-11-05
acronym: update v1.5.0 to v1.6.0

lemoncurry:
---

Reviewer Resources:

### DIFF
--- a/exercises/acronym/.meta/src/reference/java/Acronym.java
+++ b/exercises/acronym/.meta/src/reference/java/Acronym.java
@@ -9,16 +9,17 @@ class Acronym {
         this.acronym = generateAcronym(phrase);
     }
 
+
+    String getAcronym() {
+        return acronym;
+    }
+
     private String generateAcronym(String phrase) {
         return Arrays.stream(phrase.split("(?!')\\W"))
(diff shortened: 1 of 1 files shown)


exercises/acronym/.meta/src/reference/java/Acronym.java
```diff
@@ -1,24 +1,24 @@
...
+            .toUpperCase();
     }
 
-    private String generateAcronym(String phrase) {
-        return Arrays.stream(phrase.split("[^a-zA-Z]"))
-                .filter(word -> !word.isEmpty()) // Remove empty strings from the result of phrase.split
-                .map(word -> word.substring(0, 1)) // Get the first character of each word
-                .collect(Collectors.joining()) // Concatenate the characters
-                .toUpperCase();
+    String getAcronym() {
```
FridaTveit: Why did you change the order of these methods? I would expect this one to come first since it's package private while `generateAcronym` is private 🙂

lemoncurry: Oh, I'm sorry, I think I mixed up your suggestion from a previous PR. I will change it!

FridaTveit: No problem, my comment might have been confusing! Thanks! 🙂

PR #1623, opened 2019-01-24
DnD Character: Added tests for randomness and distribution of abilities

jippiee:
The idea for these tests is that currently you can simply return a constant value within range of 3 to 18 in the ability method and all test will still pass. (See [my solution](https://exercism.io/tracks/java/exercises/dnd-character/solutions/7471492bf2d44ddb95c813c12070d4aa))

The first test `testRandomAbilityIsRandom` asserts that within 100 calls to `ability()` there are at leased 2 different values returned.

The second test `testDistributionMatchesReference` asserts that the resulting distribution matches the reference distribution. This reference is simply the count of all distinct abilities for all possible dice rolls. I generated this with:
```
int[] distribution = new int[20];
for (int i = 1; i <= 6; i++) {
    for (int j = 1; j <= 6; j++) {
        for (int k = 1; k <= 6; k++) {
            for (int l = 1; l <= 6; l++) {
                int min = Math.min(i, Math.min(j, Math.min(k, l)));
                int sum = i + j + k + l - min;
                distribution[sum]++;
                // System.out.format("%d %d %d %d (min=%d) (sum=%d)%n", i, j, k, l, min, sum);
            }
        }
    }
}
System.out.println(Arrays.toString(distribution));
// [0, 0, 0, 1, 4, 10, 21, 38, 62, 91, 122, 148, 167, 172, 160, 131, 94, 54, 21, 0]
```
To be considered as matching, a distribution populated by 10000 random values must differ less than 2% from the reference. These two values (cycles=10000, threshold=0.02) were determined by experimentation. The larger this first value (cycles) is, the more precise the second value (threshold) can be. To determine the threshold I let the entire test loop 1000000 times and keep track of the max difference. It was 0.0166. I created this test to catch the implementations that only return the sum of 3 dice rolls or (even wore) return a random number between 3 and 18. 

---

Reviewer Resources:


sshine:
Hi @jippiee, and thanks for your PR!

Here comes a very delayed response, but better late than never I hope!

With this response I hope that @exercism/java will find interest in eventually approving it.

I came up with this exercise (in exercism/problem-specifications#1397) and @lemoncurry created the exercise on the Java track (in exercism/java#1621). One of my motivations behind this exercise was that I was working with random numbers at work and had to test that some function returned random data according to the proper probability distribution.

Testing that a solution doesn't use `rand(3, 18)` but uses `rand(1,6) + rand(1,6) + rand(1,6)` was also brought up [here](https://github.com/exercism/problem-specifications/pull/1397#issuecomment-443710119) where some maintainers (myself included) deferred from testing this extensively. But since my original motivation was to stir conversation that could guide me towards a good solution to testing randomness (using declarative and/or probabilistic approaches), I still think this is a subject worth spreading learning material on.

Let us know what @lemoncurry thinks also.

As for your two proposed properties, here is some feedback for them:

- `testRandomAbilityIsRandom`: We should justify this test by saying that the probability of false negatives is astronomically small (exactly how small takes a bit of figuring out), and that the cost of the test is relatively small (100 random calls). But we should also be honest and say that we don't actually test that `ability()` is random. We test that it returns more than one thing over a few amount of calls. So perhaps rename this to something like `testAbilityIsNotConstant`? Or maybe to humor the student, `testAbilityIsProbablyNotConstant`. ;-)

- `testDistributionMatchesReference`: I am very impressed with this approach. We should justify this test by saying that it overcomes [coriolinus'es warning](https://github.com/exercism/problem-specifications/pull/1397#issuecomment-443716867) (quoted below) by reducing computational complexity on each run;

  > *[...]* Doing this the mathematically rigorous way would involve a few million iterations; at that scale, the mathematical properties of the generated numbers will be reliably measurably different from the mathematical properties of `rand(3, 18)`. That's too big for a unit test, though. *[...]*

  We can then similarly reason about the probability of false negatives during tests, and we can lower this probability by either increasing the number of cycles or the threshold for the difference between the reference distribution and the tested distribution.

  But perhaps it would comfort the reviewer to give a ballpark estimate of how often this test would produce a false negative given some parameters such as the ones submitted.

In summary,

- The implementation strategy of these probabilistic tests seems sound to me.
- The probability of false negatives can in both cases be calculated, experimentally or theoretically, for fun or for moral justification.
- We should clarify if the purpose of a probabilistic test (in particular the one that tests the distribution) is only to catch the `rand(3, 18)` vs. `rand(1, 6) + rand(1, 6) + rand(1, 6)` bug, or if we also aim to teach how to test something like this. If the latter is the case, perhaps the probabilistic tests should carry some of the justifications in this pull request with them? In particular, I don't believe that being able to justify that false negatives that occur with astronomical frequency is something that everyone will immediately derive from looking at the test code.

jmrunkle:
@jippiee - are you still actively working on this PR? If not, we might close it to clean up our backlog. It can always be reopened if needed.

jippiee:
Ok, I've updated the test and also included comments describing how the test with the distribution works. I've also included a Note on why `rand(1,6) + rand(1,6) + rand(1,6)` also gives a wrong distribution, but I'm not quite sure if that should belong in the test source.

Here I've also plotted the distributions. The green is the correct one.

exercism_dnd_character_distributions

jmrunkle:
Closing and reopening to trigger travis CI to retry this...

jmrunkle:
Looks like there is a slight style violation at line 189:

> Task :dnd-character:checkstyleTest FAILED
[ant:checkstyle] [ERROR] /home/travis/build/exercism/java/exercises/dnd-character/src/test/java/DnDCharacterTest.java:189: 'array initialization' child has incorrect indentation level 16, expected level should be 12. [Indentation]

So just reduce the indentation there to 12 (ie. delete 4 spaces). Other than that, I will try to take a look at this tomorrow morning.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -128,7 +130,144 @@ public void testRandomCharacterIsValid() {
...
+         *     Returning a random number between 3 and 18.
+         *     Returning the sum of three random numbers each between 1 and 6.
+         *
+         * The way we check if the 'ability' method is distributed correctly is
+         * by creating a histogram of the distribution by calling ability
+         * repeatedly. Then we compare it to the reference distribution within
+         * a certain tolerance.
+         */
+
+        // These two values were chosen empirically.
```
jmrunkle, line 179: I'm not sure empirically chosen values are going to be sufficient and may lead to test flakiness. Running this test on the reference solution needs to consistently pass (pretty much every time since we gate PRs on passing all tests). Can you make sure that this at least passes on at least 1000 runs of this test?

Another option is - can you modify the reference solution to have a test-only constructor which fixes the Random object with a specific seed that definitely passes the tests?

That is, instead of creating a Random object [here](https://github.com/exercism/java/blob/master/exercises/dnd-character/.meta/src/reference/java/DnDCharacter.java#L67) which will use time to help seed the object, we set it in the constructor and make a test-only constructor that takes in the Random object so that we can seed it specifically for this test case. Thoughts?

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -128,7 +130,144 @@ public void testRandomCharacterIsValid() {
...
+            double actualProbability    = (double) actualDistribution[ability] / cycles;
+            double difference           = Math.abs(referenceProbability - actualProbability);
+            if (difference > tolerance) {
+                fail();
+            }
+        }
+    }
+
+    /*
+     * Extra Notes:
```
jmrunkle, line 213: This is a very interesting statistical commentary, but I'm not sure it belongs here. Perhaps this sort of thing should go in the mentoring notes for this exercise?

exercises/dnd-character/src/test/java/DnDCharacterTest.java
sshine: @jmrunkle: The test is flaky, but at an astronomically small rate.

If we fix the seed, the next progression is this:

random_number

The novel approach here, which is not something you see in a lot of places, is to make a probabilistic property test: Every time you run the test suite, you strenghten your belief, and the chance of a false positive is quantifiably insignificant. It'd be neat if it said in a comment exactly / approximately what the odds are, except, that would put a lot of strain on the contributor.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
sshine: Yes, this really remains to be decided.

The purpose of the exercise (I can say because I came up with it) is to explore probabilistic property testing.

But any maintainer is fit to choose how to make any exercise fit some track-specific goal.

I think this stuff is ingenious -- and it is, it seems, one of the things people do wrong: Not understanding that just because the range is correct, the distribution isn't necessarily.

Alternative suggestions for placements:

 - In a hint: Would increase exposure.
 - In a git commit message: Would remove exposure.
 - In a blog post linked to from a hint: Would require someone to want to write a blog post.

sshine:
@jmrunkle wrote:
> are you still actively working on this PR? If not, we might close it

Ever since reviewing this, my impression has been that the solution is sufficient, but that it lacks explanation for maintainers to want to touch it. So my review was an encouragement to convince those less statistics savvy (myself included) that it won't, in practice, fail, without resorting to "trust me, I know math", because I think such an argument is extremely fragile in the context of Exercism, where we rely heavily on commit history.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
jmrunkle: Fixing a seed value would not actually break the testing logic at all. It just guarantees that your test *always* passes. Rather than having someone quantify the "astronomically small rate" why don't we just make our setup resilient to test failures?

I'm suggesting having a setup like this where the object takes a seed (or even the Random object itself) in the constructor. Then we have some margin of control over the randomness:

```java
private final Random random;

DnDCharacter() {
  this(new Random());
}

/** test-only constructor used for ensuring stability of tests results */
DnDCharacter(Random random) {
  this.random = random;
}
```

Most of the tests would still be able to use the empty constructor, but this one test would be able to take advantage of a fixed seed to ensure that our tests on the reference solution always pass. The only cost is that some "astronomically unlikely" student might have a test that always fails because it just happens to fail that test with the seed we chose. I much prefer that.

sshine: > Fixing a seed value would not actually break the testing logic at all.

Not fixing a seed value would not actually break the testing logic, either.

> Rather than having someone quantify the "astronomically small rate" why don't we just make our setup resilient to test failures?

Because property-based testing with arbitrary, random variables means that running the unit test again and again tests the property from new angles every single time.

> I'm suggesting having a setup like this where the object takes a seed (or even the Random object itself) in the constructor.

This is a very good suggestion. I'm saying this because that's exactly what I suggested in #1671. ;-) This was downvoted by two people about a year ago. Perhaps it is better to keep the two subjects of probabilistic testing and dependency injection separate since they're both being contended?

jmrunkle: > This is a very good suggestion. I'm saying this because that's exactly what I suggested in #1671. ;-) This was downvoted by two people about a year ago. Perhaps it is better to keep the two subjects of probabilistic testing and dependency injection separate since they're both being contended?

I would have argued in favor of this back when that debate was active. If you want to see the real effects of this thing on repeated trials, you can simply modify the test locally to not set a seed on the Random object. Automated tests that we rely on for actual development should pass, reliably, every time. I honestly do not accept any other opinion on that (especially when there are a lot of tests being run at once).

As such, I guess I am really saying that we either (1) need to add the dependency injection or (2) get rid of a distribution test that might be flaky. I prefer adding DI, but maybe that's just me.

Back in #1671, it was argued that this is only a "level 2" difficulty exercise. Maybe it should have the difficulty set a bit higher if we think DI is too difficult for early exercises?

sshine: > we either (1) need to add the dependency injection or (2) get rid of a distribution test that might be flaky

Why? The distribution test currently supports the global state of `Random`'s side-effects.

It's not flaky in practice. You can decide exactly how often you will get a false negative and make that frequency so large we won't experience it in our lifetimes. This cannot mean the same as if the frequency were observable.

While I'm a big fan of DI, it isn't necessary to make a probabilistic test. In fact, if the exercise had DI, this would instead encourage mock testing, which is argued in #1671.

At this point I sense "too many chefs", so I'll let you decide and will not object to any decision made.

jmrunkle: > this would instead encourage mock testing

I think that's highly debatable and well within our control as maintainers. It's not easy to mock the Random object precisely because you have no idea how they will use it. I just want to fix the seed so that we can guarantee that the test passes consistently for our reference implementation.

jippiee: > Can you make sure that this at least passes on at least 1000 runs of this test?

My solution has passed the updated test 1000000 times without any failure.

> this one test would be able to take advantage of a fixed seed to ensure that our tests on the reference solution always pass. The only cost is that some "astronomically unlikely" student might have a test that always fails because it just happens to fail that test with the seed we chose. I much prefer that.

I disagree. If the test should fail for the "astronomically unlikely" student with a conforming solution, they can just rerun the tests, which then should succeed, as well as all following runs. Further there may also be an "astronomically unlikely" false positive solution for that seed, which would consistently pass the tests.

The way we circumvent these issues is precisely by having the `Random` Object being seeded with a different seed every time.

As the solution has already passed the test 1000000 times, I think it is safe to assume that it will continue to do so for the foreseeable future. If there are still concerns about this I might try to figure out the actual probability of this test failing on a conforming solution (as well as passing a non-conforming one).

If we choose a fixed seed we can only control our reference solution to consistently pass the tests. Not the solutions of our students. Thus if we want to make this exercise enjoyable to them, we must hold ourselves to the same standard as them.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
jippiee: The idea for this comment was, that if a student has a wrong solution which fails the `testAbilityIsDistributedAccordingToRules` test, but they don't jet know why it is failing, they might want to check out the source of the tests and thus stumble on this note. I admit that probably not every student will find it, but for some it might help them solve the exercise.

exercises/dnd-character/.meta/src/reference/java/DnDCharacter.java
```diff
@@ -22,12 +23,14 @@
...
-        }
-        return scoreSum;
+        // Throw four dice
+        int a = random.nextInt(6) + 1;
+        int b = random.nextInt(6) + 1;
+        int c = random.nextInt(6) + 1;
+        int d = random.nextInt(6) + 1;
+        // Sum the largest three
+        int min = Math.min(Math.min(a, b), Math.min(c, d));
+        return a + b + c + d - min;
```
jippiee, line 33: The was nothing inherently wrong with the previous implementation (other than the `Random` object being created every time `ability` is called, as discussed previously here https://github.com/exercism/java/pull/1623#discussion_r410287158). This simply was the one I was using during all of my tests.

I did start out with the reference as well as a few other possible implementations, but as I started testing with more and more iterations, runtime does start to matter. This solution has about a 40% speedup compared to the sorting-array method, and about a 400% speedup compared to a stream based approach.

Now I know that performance isn't probably the most important thing for our reference solutions, but I would argue that this solution is actually more readable and easier to grasp quickly than the previous one. For example it is not clear from the name of the method `rollDices` that it returns _four_ dice in _ascending_ order. Further it isn't necessary to sort the entire array we only need the smallest element. Thus the intention is clearer.

So I would argue in favor of using this implementation in the reference solution.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -102,24 +102,29 @@ public void testAbilityModifierForScore18Is4() {
...
+        for (int i = 0; i < 1000; i++) {
+            assertTrue(inRange(dndCharacter.ability()));
+        }
     }
 
     @Ignore("Remove to run test")
     @Test
     public void testRandomCharacterIsValid() {
-        for (int i = 0; i < 1000; i++) {
+        for (int i = 0; i < 100; i++) {
```
jippiee, line 120: 100 should really be enough, as with 1000 it is the most expensive test. This is essentially the same as `testRandomAbilityIsWithinRange`.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -102,24 +102,29 @@ public void testAbilityModifierForScore18Is4() {
         assertEquals(4, dndCharacter.modifier(18));
     }
 
+    private boolean inRange(int ability) {
+        return ability >= 3 && ability <= 18;
+    }
```
jippiee, line 107: Reuse code

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -128,7 +133,224 @@ public void testRandomCharacterIsValid() {
...
+        /* Amount of possible combinations for throwing 4 dice
+         */
+        int referenceCount = 6 * 6 * 6 * 6;
+        /* This distribution was generated by going through all possible
+         * combinations for throwing four dice, computing the ability for that
+         * combination, and incrementing the array at that index.
+         */
+        int[] referenceDist = new int[] {
+            0, 0, 0, 1, 4, 10, 21, 38, 62, 91, 122, 148, 167, 172, 160, 131, 94, 54, 21
+        };
```
jippiee, line 288: Instead use actual generation code? Or maybe put the code in a comment?

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -128,7 +133,224 @@ public void testRandomCharacterIsValid() {
...
+    @Ignore("Remove to run test")
+    @Test
+    public void testAbilityIsProbablyNotConstant() {
+        int ability = dndCharacter.ability();
+        for (int i = 0; i < 100; i++) {
+            if (ability != dndCharacter.ability()) {
+                return;
+            }
+        }
+        fail();
```
jmrunkle, line 161: This should probably include an error message saying what happened so that the student could debug the test failure: "ability appears to always return the same value"

exercises/dnd-character/src/test/java/DnDCharacterTest.java
jmrunkle: Personally, I am leaning towards having a hints.md file that includes this extra information about this test. If it makes you feel better, put a Javadoc comment on the test method telling them to check the hints.

exercises/dnd-character/.meta/src/reference/java/DnDCharacter.java
jmrunkle: Works for me.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
```diff
@@ -128,7 +133,224 @@ public void testRandomCharacterIsValid() {
...
+         * The parameters (10000, 0.1, 0.01) were chosen based on the results
+         * of the table above, and were tested further in the test setup with a
+         * loop of 1000000, which passed.
+         *
+         * Further following should hold true:
+         *     relativeError <= 0.2
+         *     constantError <= 0.02
+         * Otherwise it becomes increasingly likely that an 'ability' method
+         * which gives "Throw three dice and sum them" would wrongly pass the
+         * test.
```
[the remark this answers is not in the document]
jmrunkle: I think this is missing the graph showing the error bounds for the green graph. This only shows an idealized distribution for the green case (roll 3, sum them). I'm curious that there is even more overlap if you consider the error bounds for the green distribution. With that said, I think this is probably sufficient to get us close enough as-is.

exercises/dnd-character/src/test/java/DnDCharacterTest.java
jmrunkle: I'm debating the merits. I assume the generation is time-prohibitive to just generate it in the test. We could leave it in an unused method just for documentation purposes?

sshine: Is this comment bad?

If you ask me the ingenious part of this PR is that @jippiee precomputes the heavy parts and remains confident to any chosen probability that there isn't false negatives. But this array is certainly magical and deserves an explanation.

sshine: Also, while some comments can be moved to the hint file, this one is overly specific about some magical constants. So I might be inclined to keep it local, since you really don't care about what the numbers are until you read them, but they also really need an explanation.

sshine: That's not a bad idea, but not an entirely uncontentious one either.

Dead code vs. looking up commits for explanations.

I would be happy with either.

sshine: I'm not saying we should go with the following, but it made me think of [golden tests](https://ro-che.info/articles/2017-12-04-golden-tests), i.e. unit tests stored in files.

jmrunkle:
@jippiee - any more updates on this PR?

sshine:
To get this PR merged, I see some points that were raised:

1. https://github.com/exercism/java/pull/1623#discussion_r410752310: Move "Extra Notes" into .meta/hints.md, or not.
2. https://github.com/exercism/java/pull/1623#discussion_r410760063: Put distribution generator into dead function, create golden test, or keep as-is.
3. https://github.com/exercism/java/pull/1623#discussion_r411857163: Add meaningful error message when ability is probably constant.

The comment links refer to the discussion on each of these.

It would be a pity to not have this included because of minor, unresolved ideas about where to place comments. :-)

Also, this PR probably needs rebasing after v3 has moved practice exercises to a sub-directory.

jmrunkle:
If there are no further updates or comments on this PR, I will probably close it in one week. Otherwise, please comment and then synchronize the files and we can continue the review.

jmrunkle:
> If there are no further updates or comments on this PR, I will probably close it in one week. Otherwise, please comment and then synchronize the files and we can continue the review.

Closing as there has been no followup on this. Feel free to reopen if / when you are willing to address the merge conflicts, etc and continue the review.

PR #1629, opened 2019-02-12
hamming: update to v2.3.0, change reference implementation

lemoncurry:
---

Reviewer Resources:


exercises/hamming/.meta/src/reference/java/Hamming.java
```diff
@@ -1,5 +1,5 @@
 class Hamming {
-    private final int hammingDistance;
+    private int hammingDistance;
```
Smarticles101: Why'd you remove `final`?

lemoncurry: In the `policies.md` I read something about avoiding the `final` keyword, but maybe it is appropriate here? Should I change back?

FridaTveit: Yes, it's true that we mention that in our policies 🙂 However, that's mainly for the starter solutions and tests as those are visible to the people solving the problems. We want to avoid using unnecessary syntax in those files so that they're as easy to understand as possible for people with limited experience of Java. However, that's not as essential in the reference solution as that's not visible to people solving the exercises.

I'd also argue that this is one case where using the `final` keyword is more common and important. It can be very useful to know that a class variable is final and won't be changed. 

Does that all make sense? Sorry if it's a bit confusing! So it's not very important, but I'd probably argue for keeping `final` in this case 🙂

lemoncurry: That makes sense. I will change it back!

PR #1633, opened 2019-02-20
Fixed tests to call correct method.

moaxcp:
---

Reviewer Resources:


Smarticles101:
Im curious, what precedent precedes this change?

moaxcp:
When I started this exercise the tests fail to compile (rather than failing). The problem is the method names do not match between the provided classes. Fixing this issue will provide a better experience.

Smarticles101:
Ah I see, the implementation stub uses `get`.

PR #1634, opened 2019-02-24
switching isEmpty tests to use hamcrest

moaxcp:
The purpose of this change is to provide better test diagnostics when an empty list is expected. Rather than a general AssertionError the test will display the content of the list if it is not empty.

---

Reviewer Resources:

### DIFF
--- a/exercises/anagram/build.gradle
+++ b/exercises/anagram/build.gradle
@@ -8,6 +8,7 @@ repositories {
 
 dependencies {
   testCompile "junit:junit:4.12"
+  testCompile "org.hamcrest:hamcrest-all:1.3"
 }
 
 test {
--- a/exercises/anagram/src/test/java/AnagramTest.java
+++ b/exercises/anagram/src/test/java/AnagramTest.java
@@ -5,8 +5,8 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
-import static org.hamcrest.CoreMatchers.*;
-import static org.junit.Assert.*;
+import static org.hamcrest.Matchers.*;
+import static org.hamcrest.MatcherAssert.assertThat;
 
 public class AnagramTest {
 
@@ -14,7 +14,8 @@ public class AnagramTest {
     @Test
     public void testNoMatches() {
         Anagram detector = new Anagram("diaper");
-        assertTrue(detector.match(Arrays.asList("hello", "world", "zombies", "pants")).isEmpty());
+        List<String> anagrams = detector.match(Arrays.asList("hello", "world", "zombies", "pants"));
+        assertThat(anagrams, empty());
     }
 
     @Ignore("Remove to run test")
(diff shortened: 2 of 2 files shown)


moaxcp:
I could understand wanting a consistent testing experience but I'm not sure the tests are consistent. Some use junit asserts while others use hamcrest. Switching between each in the same exercise can be confusing. Some have nicely formatted error messages while others do not. In my opinion hamcrest-core is incomplete and at some point you always run into the need to use hamcrest-all. This is a good example.

```
Anagram detector = new Anagram("good"); 
 List<String> anagrams = detector.match(Arrays.asList("dog", "goody")); 
assertThat(anagrams, empty()); 
```

In my exercise the list contained "dog". The in assertTrue did not help and I needed to debug the test to figure this out. With this information I was able to add a length check. If the test used hamcrest-all there would be no need to debug. That is the sign of a good test diagnostic.

Also using hamcrest-all is not any more confusing than using hamcrest-core. It has the same syntax as hamcrest-core while providing more support for the core of the java sdk (such as checking that a collection is empty).

jmrunkle:
This feels like the type of thing where we need to decide if we are going to allow the use of hamcrest or not. IMO: we should open an issue for the discussion so that we can decide on the merits.

sshine:
This feels like the right thing to do so that the track can progress in the absence of its former maintainers.

jmrunkle:
To add another dog in this fight, I would also suggest that [Guava's Truth](https://truth.dev/) package could be a better option. They read like plain English assertions (so it should be pretty obvious what is being asserted), and the assertion errors are much clearer.

An example assertion from that library:

```java
assertThat(someString).contains("testuser@google.com");
assertThat(listOfNumbers).containsExactly(1, 2, 3).inOrder();
assertThat(mapOfFirstNamesToID).containsEntry("John", 123);
```

Here is an example of what the errors look like:

```
value of    : projectsByTeam().valuesForKey(corelibs)
missing (1) : truth
───
expected    : [guava, dagger, truth, auto, caliper]
but was     : [guava, auto, dagger, caliper]
multimap was: {corelibs=[guava, auto, dagger, caliper]}
```

Full disclosure: Guava Truth is what we use at my work so I'm more than a little biased. Just wanted to include it here if additional assertion libraries are up for debate.

moaxcp:
I really like truth and have used it for an entire project. I mostly use assertj though. Those are the main assertion libraries in java and it would be good to be exposed to all of them.

jmrunkle:
It seems that we are now talking about more than just this PR specifically and it should probably be a broader discussion about allowing the use of more fully-featured assertion libraries. IMO: we should move this discussion to a dedicated issue and then create another PR to modify our policies to either (1) explicitly allow one or more libraries or (2) explicitly disallow them.

@moaxcp - would you mind creating an issue for that purpose and mentioning the three contenders (Hamcrest, Truth, and AssertJ)? Give a short proposal for your preference and then we can resume discussion over there.

jmrunkle:
> @moaxcp - would you mind creating an issue for that purpose and mentioning the three contenders (Hamcrest, Truth, and AssertJ)? Give a short proposal for your preference and then we can resume discussion over there.

Never mind - I opened #1803 myself.

jmrunkle:
It looks like the answer is going to be that we do not allow Hamcrest, but that we will allow AssertJ. Would you like to update this PR to use the AssertJ equivalent?

NOTE: We will have some other cleanup to do getting rid of Hamcrest and switching those use-cases to AssertJ.

jmrunkle:
I think I actually ended up fixing this in broad-stroke while getting rid of all hamcrest usage in #1823. As such, closing this PR. Reopen with changes if there is more to be fixed.

PR #1635, opened 2019-02-26
[#1540]Binary search: Update error handling

### LINKED ISSUE #1540: binary-search: update version file (v1.1.0 -> v1.2.0) and update error handling
Please update the version specified in the [binary-search](https://github.com/exercism/java/blob/master/exercises/binary-search/.meta/version) version file from 1.1.0 to 1.2.0.

Please also make sure that the expected error handling in the [robot-simulator tests](https://github.com/exercism/java/blob/master/exercises/binary-search/src/test/java/BinarySearchTest.java) is up to date with the [canonical data for binary-search](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search/canonical-data.json). Currently we expect a solution to binary-search to return -1 if the value searched for is not present. However the canonical data now indicates that this should cause an error in a way that is idiomatic to the language. I suggest throwing an exception in that case. Thoughts @exercism/java? 🙂 

After any changes have been made to the tests make sure that the [reference solution](https://github.com/exercism/java/blob/master/exercises/binary-search/.meta/src/reference/java/BinarySearch.java) still passes the tests. If it doesn't then please update the reference solution so that it passes all the tests.

RonakLakhotia:
Fixes #1540


FridaTveit:
@RonakLakhotia Thanks for raising this PR! 😄 

You appear to have deleted the build.gradle file for binary-search which is why the build is failing 🙂

RonakLakhotia:
Sorry, my bad :P fixing the issue soon

RonakLakhotia:
@FridaTveit ready for review

PR #1637, opened 2019-02-27
upgrade phone-number to v1.7.0

mirkoperillo:
Hi,

I'm trying to upgrade the version of phone-number exercise but I have some issues about.

The actual version of exercise in the java track is v1.6.0. 
If I compare it with the relative version in canonical-data.json here

https://github.com/exercism/problem-specifications/commit/7290b84b809ccdacfa7ab674b1cb1237ca55105f?diff=unified

it seems that the exception messages are not aligned to the canonical ones.
Do I have to align them ?

---

Reviewer Resources:


Smarticles101:
Looks like those exception messages have existed for at least 2 years. The canonical data added the error messages after that. I am in favor of changing the exception messages to be more in line with the canonical data.

Smarticles101:
As an interesting side note, github recently released [draft pull requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/) so you don't have to specify wip anymore. :)

mirkoperillo:
@Smarticles101 ok I'll fix them.
Thank you for suggest me draft PR

mirkoperillo:
@Smarticles101 done

PR #1639, opened 2019-03-07
[#1475] Update Cipher Tests

### LINKED ISSUE #1475: simple-cipher: update tests and add version file
The [simple-cipher tests](https://github.com/exercism/java/tree/master/exercises/simple-cipher/src/test/java) should be updated to exactly match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json). Also a version file should be added to match the canonical data version.

RonakLakhotia:
Fixes #1475


RonakLakhotia:
@FridaTveit how to detect the checkstyle errors?

FridaTveit:
See our [contributing guide](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#contributing-with-minimal-setup). You can run the plugin locally by running `gradle check` from the root folder of the exercise. Then the checkstyle plugin will show you every style violation of your code. Let me know if you have any issues with that 🙂

RonakLakhotia:
I seem to be getting a `command not found` error.

RonakLakhotia:
@FridaTveit gentle reminder. 🙂

FridaTveit:
@RonakLakhotia We are all maintaining this as volunteers in our spare time and therefore it will sometimes take a while for us to get back to you. Please be patient 🙂

RonakLakhotia:
> We are all maintaining this as volunteers in our spare time and therefore it will sometimes take a while for us to get back to you. Please be patient 🙂

My apologies.

RonakLakhotia:
rectified the checkstyle errors. 🙂

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
```diff
@@ -32,10 +32,14 @@ public String getKey() {
...
+        String copy = plainText;
         StringBuilder ciphertext = new StringBuilder(plainText.length());
 
-        for (int index = 0; index < Math.min(plainText.length(), key.length()); index++) {
-            ciphertext.append(encodeCharacter(plainText, index));
+        while (copy.length() > 0) {
+            for (int i = 0; i < Math.min(copy.length(), key.length()); i++) {
+                ciphertext.append(encodeCharacter(copy, i));
+            }
+            copy = plainText.substring(ciphertext.length(), plainText.length());
```
FridaTveit: Why use this instead of the for loop?

exercises/simple-cipher/src/test/java/SimpleCipherStepOneTest.java
```diff
@@ -21,8 +24,10 @@ public void setup() {
      */
     @Test
     public void cipherCanEncode() {
+        String plainText = "aaaaaaaaaa";
         String cipherText = cipherWithDefaultKey.getKey().substring(0, 10);
         assertEquals(cipherText, cipherWithDefaultKey.encode("aaaaaaaaaa"));
+        assertEquals(cipherText, cipherWithDefaultKey.encode(plainText));
```
FridaTveit: You don't need both assert equals statements 🙂

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeIncorrectKeyTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.rules.ExpectedException;
+
+public class SimpleCipherStepThreeIncorrectKeyTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Before
+    public void setup() {
+    }
```
FridaTveit: This isn't needed 🙂

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeIncorrectKeyTest.java
```diff
@@ -0,0 +1,42 @@
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class SimpleCipherStepThreeIncorrectKeyTest {
```
FridaTveit: Looks like these tests aren't part of the canonical data anymore so we can remove them 🙂

exercises/simple-cipher/.meta/version
```diff
@@ -0,0 +1 @@
+1.1.0
```
FridaTveit: We might as well bump this to version 2.0.0 are remove the step three tests so that the tests are fully up to date? 🙂

exercises/simple-cipher/src/test/java/SimpleCipherStepOneTest.java
RonakLakhotia: My bad

exercises/simple-cipher/src/test/java/SimpleCipherStepThreeIncorrectKeyTest.java
RonakLakhotia: Reverted

exercises/simple-cipher/.meta/version
RonakLakhotia: Done. 👍

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
RonakLakhotia: You mean for loop instead of substring?

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
```diff
@@ -32,10 +32,17 @@ public String getKey() {
...
 
-        for (int index = 0; index < Math.min(plainText.length(), key.length()); index++) {
-            ciphertext.append(encodeCharacter(plainText, index));
+        while (copy.length() > 0) {
+            for (int i = 0; i < Math.min(copy.length(), key.length()); i++) {
+                ciphertext.append(encodeCharacter(copy, i));
+            }
+            copy = "";
+            for (int i = ciphertext.length(); i < plainText.length(); i++) {
+                copy = copy + plainText.charAt(i);
```
RonakLakhotia: Like this you mean?

FridaTveit: No, sorry if I wasn't clear. I meant why did this have to change at all? Why not keep the for loop as it used to be? Does that not work with the changes to the tests? 🙂

RonakLakhotia: With the original code, the `cipherMessageLongerThanKey` test case seems to be failing

FridaTveit: Oh I see 🙂 Well, I think adding the extra while loop makes it a bit confusing? Maybe it would be easier to add a small helper method which will pad or cut the key so that it's the same length as the plaintext? That is something like:
```
String sameLengthKey = getKeySameLengthAsPlaintext(plaintext, key);
for (int index = 0; index < plaintext.length(); index++) {
  ciphertext.append(encodeCharacter(plaintext, index, key));
}
```

What do you think? 🙂 It's not very important though, so feel free to keep it as it is if you like. But if so please change it back to use substring 🙂

RonakLakhotia: @FridaTveit made the relevant changes as you suggested. 👍

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
```diff
@@ -32,23 +32,29 @@ public String getKey() {
...
-            }
+        for (int index = 0; index < plainText.length(); index++) {
+            ciphertext.append(encodeCharacter(plainText, index, sameLengthKey));
         }
 
         return ciphertext.toString();
     }
 
-    private char encodeCharacter(String plainText, int index) {
+    private String getKeySameLengthAsPlaintext(String plainText, String key) {
```
FridaTveit: I think this method actually returns a key that is at least as long as the plaintext? Which, when I think about it, should be fine and is probably better than trying to get it to be exactly the same length 🙂 But it's probably good to reflect that in the method name? So something like `getSufficientlyLongKeyForPlaintext` or something? 🙂

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
```diff
@@ -32,23 +32,29 @@ public String getKey() {
...
+
```
FridaTveit: Very minor but this empty line is probably not needed 🙂

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
```diff
@@ -32,23 +32,29 @@ public String getKey() {
...
+        String copy = key;
```
FridaTveit: It might be better to use a [StringBuilder](https://docs.oracle.com/javase/9/docs/api/java/lang/StringBuilder.html) here as we're doing possibly quite a few appends? And in that case naming it something like `longKeyBuilder` might be more descriptive? 🙂

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
RonakLakhotia: Good point

exercises/simple-cipher/.meta/src/reference/java/Cipher.java
RonakLakhotia: Reverted. 👍

RonakLakhotia:
> Looks great, thanks @RonakLakhotia 🙂

Thanks for the guidance @FridaTveit 👍

PR #1660, opened 2019-04-25
Update hamming reference

michael-berger-FR:
Java track/Hamming exercise/reference implementation: offer a more modern solution (Java 8), including a more concise management of exceptions

---

Reviewer Resources:

### DIFF
--- a/exercises/hamming/.meta/src/reference/java/Hamming.java
+++ b/exercises/hamming/.meta/src/reference/java/Hamming.java
@@ -1,36 +1,26 @@
+import java.util.function.IntPredicate;
+import java.util.stream.IntStream;
+
 class Hamming {
-    private final int hammingDistance;
+    private final String leftStrand, rightStrand;
 
     Hamming(String leftStrand, String rightStrand) {
-        hammingDistance = computeHammingDistance(leftStrand, rightStrand);
-    }
-
-    int getHammingDistance() {
-        return hammingDistance;
-    }
-
-    private int computeHammingDistance(String leftStrand, String rightStrand) {
-        if (leftStrand.length() == 0 && rightStrand.length() == 0) {
-            return 0;
-        } else if (leftStrand.length() == 0) {
-            throw new IllegalArgumentException("left strand must not be empty.");
-        } else if (rightStrand.length() == 0) {
-            throw new IllegalArgumentException("right strand must not be empty.");
-        } else if (leftStrand.length() != rightStrand.length()) {
-            throw new IllegalArgumentException("leftStrand and rightStrand must be of equal length.");
+        String exceptionMessage = "leftStrand and rightStrand must be of equal length.";
+        if (leftStrand.length() != rightStrand.length()) {
+            if (leftStrand.isEmpty()) {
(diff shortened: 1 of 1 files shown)


exercises/hamming/.meta/src/reference/java/Hamming.java
```diff
@@ -1,36 +1,26 @@
+import java.util.function.IntPredicate;
+import java.util.stream.IntStream;
+
 class Hamming {
-    private final int hammingDistance;
+    private final String leftStrand, rightStrand;
```
FridaTveit: Why store the left strand and right strand instead of just storing the hamming distance? As the distance won't change it seems more efficient to just compute it once? 🙂

exercises/hamming/.meta/src/reference/java/Hamming.java
```diff
@@ -1,36 +1,26 @@
...
-        }
-
-        return distance;
+        this.leftStrand = leftStrand;
+        this.rightStrand = rightStrand;
     }
 
-    private int hammingContributionAt(int index, String leftStrand, String rightStrand) {
-        return leftStrand.charAt(index) != rightStrand.charAt(index) ? 1 : 0;
+    long getHammingDistance() {
```
FridaTveit: Any reason why you've changed this to `long`? Now it won't match the starter implementation 🙂

exercises/hamming/.meta/src/reference/java/Hamming.java
michael-berger-FR: Hello @FridaTveit , thanks for the review. The reason is that, if we imagine  a Real Life situation, the Hamming Class could offer other methods than `getHammingDistance()`. I'll add that almost every solution submitted by students include the storage of both strands. But I agree that if we focus only on completing the exercise, it is not needed. I'll let you decide!

FridaTveit: I think it's best to optimise the current solution instead of considering possible other scenarios. If I were writing this code in real life I would only store the hamming distance. If at some point I needed to add other methods to the class that needed to use the individual strands then I would change my implementation. But there's no point changing it before it's needed 🙂

exercises/hamming/.meta/src/reference/java/Hamming.java
michael-berger-FR: @FridaTveit, you're right, I should have modified the starter implementation. When I think of it, maybe it's not worth to update the exercise for this matter. I'll add the `int` cast in the return statement!

FridaTveit: Great, thanks! 🙂

exercises/hamming/.meta/src/reference/java/Hamming.java
michael-berger-FR: Good point, I'll update the file accordingly!

michael-berger-FR: @FridaTveit, thanks for the REALLY quick answers ;-)
Changes commited!

PR #1661, opened 2019-04-30
Fix integer overflow in binary-search solution

### LINKED ISSUE #1658: binary-search: integer overflow is possible in solution
https://github.com/exercism/java/blob/580118a3a555fad422b9a7fdf5042e1533a2f389/exercises/binary-search/.meta/src/reference/java/BinarySearch.java#L24

It's possible for the sum to cause integer overflow.
The expression should be changed to safer `middle = (int) Math.floor(left + (right - left) / 2);`

wolf99:
Fixes #1658

---

Reviewer Resources:


lemoncurry:
Thank you @wolf99

PR #1706, opened 2019-06-20
robot-simulator: update version and tests to v3.2.0 (Fixes #1682)

### LINKED ISSUE #1682: robot-simulator: update tests and version file (v3.0.0 -> v3.2.0)
The tests for the [robot-simulator exercise](https://github.com/exercism/java/tree/master/exercises/robot-simulator/src/test/java) are not up-to-date with the [canonical data](https://github.com/exercism/problem-specifications/tree/master/exercises/robot-simulator/canonical-data.json).
To check what has changed and if new tests have been added to the canonical data, please have a look at the [commit history](https://github.com/exercism/problem-specifications/commits/master/exercises/robot-simulator/canonical-data.json) of the canonical data file for the exercise robot-simulator.
Please update version and/or test file, and, if the new test(s) fail, fix the reference implementation.
Feel free to leave a comment if you have any questions. 

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of the [canonical data file](https://github.com/exercism/problem-specifications/tree/master/exercises/robot-simulator/canonical-data.json) for that exercise.

jmrunkle:
Update the robot-simulator version and tests to v3.2.0 which Fixes #1682 

---

Reviewer Resources:


jmrunkle:
Thanks for the review.

> Since we usually don't put comments in the test file, I would suggest to add what you put into the comments right into the method names eg "testCreateRobotAtOriginFacingNorth".

How about this style (which is commonly used at my company): `createRobot_atOriginFacingNorth`

This makes the clustered groups of tests obvious by the prefix before the underscore.

> Also I would suggest to start each test method name with "test".

I think that the style of naming tests starting with the word "test" goes back to the days when JUnit required your tests to use that (it was how it found your tests in the past). Now that we are using a newer version of JUnit (which does not have that requirement), I see no significant reason to clutter the test method names with an extra word that is already there in an annotation (`@Test`).

lemoncurry:
> @lemoncurry Thanks for the review.
> 
> > Since we usually don't put comments in the test file, I would suggest to add what you put into the comments right into the method names eg "testCreateRobotAtOriginFacingNorth".
> 
> How about this style (which is commonly used at my company): `createRobot_atOriginFacingNorth`
> 
> This makes the clustered groups of tests obvious by the prefix before the underscore.

I would propose to stick to the standard naming conventions of method names in Java.

> 
> > Also I would suggest to start each test method name with "test".
> 
> I think that the style of naming tests starting with the word "test" goes back to the days when JUnit required your tests to use that (it was how it found your tests in the past). Now that we are using a newer version of JUnit (which does not have that requirement), I see no significant reason to clutter the test method names with an extra word that is already there in an annotation (`@Test`).

Yes, I think that is something that we should discuss.
Right now we don't have a uniform way of naming for the test methods, and my impression was that we have more test method names starting with "test" overall. That's why I suggested this change.
But you are right, of course with all your arguments about the prefix "test".

We should definitely discuss if and how we go about the test method naming in the future.

Issue #1589, opened 2018-10-04
Make it more clear which exercises can't be implemented for Java


FridaTveit:
There are a few exercises that can't be implemented for Java: `leap`, `grains` and `say`. See issue #3 for an explanation of why.

The build will fail if someone tries to open a PR to add them, but apart from that it's not very obvious that they can't be added. Therefore we risk someone implementing them before realising they can't be added. There's already been several PRs openened and closed for `leap`.

To avoid that it would be good if we could make it more obvious that these exercises shouldn't be added for this track. Maybe adding a sentence or two to the contributing guide or something would help? What do you think @exercism/java? 🙂

lemoncurry:
I would propose to reference maybe an open issue that includes a list of all exercises that are not implemented yet in the [contributing guide](https://github.com/exercism/java/blob/master/CONTRIBUTING.md).
This list could also include a section for those exercises that are excluded from the track. Right now there is already a link that leads to a general page for contributing -> [https://exercism.io/contribute](https://exercism.io/contribute).

junming403:
Hi, I want to contribute to this repo. I am considering adding some algorithm questions. Should I propose the topic for approvement first? Thank you

FridaTveit:
That's great that you want to contribute @junming403! 🙂 Do you mean you want to add new exercises? In which case that needs to be done in the [problem specifications repo](https://github.com/exercism/problem-specifications). We only implement the exercises that are specified in that repo 🙂 Does that make sense?

If there are any exercises in the problem specifications repo that aren't implemented on this track then you can also open a PR in this repo to implement one of them. See [our contributing guide](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#adding-a-new-exercise) for more details on how to add a new exercise 🙂

junming403:
Hi, I see, is there a list of missing problems I can try to contribute to?

FridaTveit:
There used to be but I don't think there is anymore unfortunately 🙁 So I think the easiest thing to do at the moment is just to see if there are any exercises in the problem specifications repo that aren't in our list of exercises in this repo. You can then try to port over any of those exercises (apart from `leap`, `grains` and `say` as explained in this issue).

You can also update the tests for any of the exercises. To do that you can [run a script](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#checking-tests-are-up-to-date) to check which tests need updating 🙂 

And you can suggest new exercises or improvements to the current exercises in the problem specifications repo 🙂

[...]

PR #1711, opened 2019-07-07
Create a DO_NOT_IMPLEMENT.md file for java/leap

jmrunkle:
Add a DO_NOT_IMPLEMENT.md file to make it clear why we cannot implement this exercise. Part of #1589

---

Reviewer Resources:

### DIFF
--- a/exercises/leap/DO_NOT_IMPLEMENT.md
+++ b/exercises/leap/DO_NOT_IMPLEMENT.md
@@ -1,0 +1,2 @@
+The leap exercise was insipred by [Java Ranch](https://javaranch.com/) and
+cannot be implemented on exercism due to licensing issues.


jmrunkle:
If we decide to go this route, I would suggest that we need to change the CI code to allow this (and only this) for the exercises that should not be implemented.

jmrunkle:
Looks like the error comes from the configlet that is a bit harder to change / ignore since it is used by many projects. Might have to have that discussion over in the configlet itself rather than directly in here.

SleeplessByte:
I think this makes sense.

jmrunkle:
Created https://github.com/exercism/configlet/pull/163 which I believe would allow this to work

yawpitch:
The motivation for this ended with #1712 being merged in late July. See my comment [here](https://github.com/exercism/exercism/issues/4932#issuecomment-552202038).

yawpitch:
Since exercism/exercism#4932 has been closed, this probably should be as well.

Issue #1736, opened 2019-09-18
matrix: update tests and version file (v1.1.0 -> v1.3.0)


lemoncurry:
The tests for the [matrix exercise](https://github.com/exercism/java/tree/master/exercises/matrix/src/test/java) are not up-to-date with the [canonical data](https://github.com/exercism/problem-specifications/tree/master/exercises/matrix/canonical-data.json).
To check what has changed and if new tests have been added to the canonical data, please have a look at the [commit history](https://github.com/exercism/problem-specifications/commits/master/exercises/matrix/canonical-data.json) of the canonical data file for the exercise matrix.
Please update version and/or test file, and, if the new test(s) fail, fix the reference implementation.
Feel free to leave a comment if you have any questions. 

Background info: each exercise which has canonical data should have a version file. This file states which version of the canonical data the exercise implements. The version can be found at the top of the [canonical data file](https://github.com/exercism/problem-specifications/tree/master/exercises/matrix/canonical-data.json) for that exercise.

sonapraneeth-a:
@lemoncurry , Shall I pick this up?

lemoncurry:
> @lemoncurry , Shall I pick this up?

@sonapraneeth-a Yes, you can work on it :smile:

sonapraneeth-a:
@lemoncurry , I seem to get some check style issues in the PR. Not sure what are they? Could you please help me understand them, so that I can ensure a clean build?

PR #1742, opened 2019-10-01
updated matrix exercises to version 1.3.0

sonapraneeth-a:
Signed-off-by: Sona Praneeth Akula <sonapraneeth.akula@gmail.com>

Fixes issue: [#1736](https://github.com/exercism/java/issues/1736)

---

Reviewer Resources:


exercises/matrix/src/test/java/MatrixTest.java
```diff
@@ -16,7 +16,6 @@ public void extractRowFromOneNumberMatrixTest() {
         assertArrayEquals(expectedRow, matrix.getRow(rowIndex));
     }
 
-    @Ignore("Remove to run test")
     @Test
```
lemoncurry: Except for the first test case, please re-include the @Ignore annotation

sonapraneeth-a: But would that not run the tests as per canonical-data.json as you're ignoring some tests?

lemoncurry: The annotation has no effect on testing with gradle (you can test with `gradle test`). See for example other PRs or just try for yourself. It is a track policy to include the annotation as explained.

PR #1768, opened 2019-10-06
Updated travis to openjdk11

uzilan:
Travis seems not to support jdk8 anymore. Instead of using dist: trusty as a workaround, I suggest we use openjdk11.

---

Reviewer Resources:


uzilan:
Could this PR be labeled with hacktoberfest please?

lemoncurry:
Can you explain why this change is necessary?
Also, have you tested whether the build will pass? The full journey test was not triggered by your PR.

uzilan:
Travis does not support jdk8 anymore. A workaround was used to mitigate this, by forcing travis to use the old Trusty distribution which still allowed jdk8. The main problem I see with this is that it prohibits the common solutions of the exercises from using later versions of java, specifically the current LTS one, jdk11. Oracle does no longer support jdk8 and I as a mentor have seen many jdk11 solutions used by mentees. I think it's time to move on.

I have ran the full journey on travis when I pushed the branch on my fork. It looks good and can be found here: https://travis-ci.org/uzilan/exercism-java/jobs/594125154

Btw why didn't it run fully when I created the PR?

FridaTveit:
As long as all the tests still work then I think upgrading the minimum Java version sounds sensible 🙂

lemoncurry:
> As long as all the tests still work then I think upgrading the minimum Java version sounds sensible 🙂

Yes, I think so too :smile:
But I think a change of the version should also include any file we have referencing the min. java version, such as the guides [here](https://github.com/exercism/java/tree/master/docs).
@uzilan Could you include that in your PR? 🙂

uzilan:
Will do :)

lemoncurry:
@uzilan Thank you!

uzilan:
I've started verifying openjdk11 installations on linux, in order to update [Installing Java](https://github.com/exercism/java/blob/master/docs/INSTALLATION.md) and realized that on Fedora, running `sudo dnf install gradle` installs gradle version 4.4.1 and jdk 8, and that there are no newer package than that. Googling that, I see that people suggest using [sdkman](https://sdkman.io) instead (I also use it on my mac, great for switching versions). Should we recommend using that instead?

lemoncurry:
No newer Fedora package? :disappointed:
Recommending `sdkman` seems reasonable. Did you have a chance to test it? Which Fedora version do you use?

uzilan:
I tried Fedora 30, and no, there is no newer gradle package. I tried sdkman and it's working great. Easy to install both openjdk11 and the latest gradle using it, so I think I'll use that in the instructions.

I'm also having issues with mint on docker:
```
$ sudo add-apt-repository ppa:openjdk-r/ppa
Traceback (most recent call last):
  File "/usr/lib/linuxmint/mintSources/mintSources.py", line 25, in <module>
    import gi
ImportError: No module named gi
```
So I think maybe I'll update the instructions I've been using successfully on Ubuntu and recommend using sdkman as an alternative.

PR #1769, opened 2019-10-06
implement say exercise

### LINKED ISSUE #1752: say: implement exercise
The exercise **say** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/master/exercises/say). 

mirkoperillo:
Add folder structure and settings for say exercise.
Work in  progress. Fixes #1752


mirkoperillo:
Hi @lemoncurry this is my proposal for say exercise.
I miss some settings, I think: how to produce README and how to add exercise entry into `config.json`.
Can you give me some hints about this to take the right way ? Thank you

lemoncurry:
@mirkoperillo How to create the README.md is described [here](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#updating-the-readmes).
You can just add a new entry for this exercise into config.json. You may decide on the difficulty level, topics, etc.
Let me know if you have further questions.

mirkoperillo:
done @lemoncurry.
waiting for the review

lemoncurry:
Thank you @mirkoperillo for contributing again to the Java track :smile:
There seems to be a violation of the style rules, could you please check that?
Then I will review your PR.

mirkoperillo:
@lemoncurry now `gradle check` ends without errors

exercises/say/.meta/src/reference/java/Say.java
```diff
@@ -0,0 +1,204 @@
+import java.util.Arrays;
+
+public class Say {
+
+    public String say(long number) {
+        if (number < 0) {
+            throw new IllegalArgumentException("input out of range");
+        }
+        char[] digits = String.valueOf(number).toCharArray();
```
jmrunkle, line 9: Rather than processing a list of characters, why not turn the input number into a long and then we can extract the numbers using numeric operations (eg. integer division and modulo)? This also gets rid of the leading zeros problem that you had to address multiple times in this solution.

exercises/say/.meta/src/reference/java/Say.java
```diff
@@ -0,0 +1,204 @@
+import java.util.Arrays;
+
+public class Say {
+
```
jmrunkle, line 4: General comment about this solution: I think we need a better breakdown here. This one is very chaotic and hard to follow. The instructions recommend a certain strategy that chunks things in groups of three digits. I suggest we try to have our solution mirror that suggestion.

exercises/say/.meta/src/reference/java/Say.java
```diff
@@ -0,0 +1,204 @@
...
+            case 2:
+                return "thousand";
+            case 3:
+                return "million";
+            case 4:
+                return "billion";
+            default:
+                return "";
+        }
+    }
```
jmrunkle, line 123: These feel like an enum...

exercises/say/.meta/src/reference/java/Say.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private String convertUnit(char digit) {
```
jmrunkle, line 125: If working with longs instead of individual characters, we would be able to use a three digit set (like 123) and then extract the hundreds, tens, and ones digits to correctly say what we want. Something like this:

```java
private String convertNumber(long number) {
  // optionally - throw if number is greater than 999
  int hundreds = number / 100;
  int tens = (number % 100) / 10;
  int ones = number % 10;
  StringBuilder sb = new StringBuilder();
  if (hundreds > 0) {
    sb.append(convertOnes(hundreds)).append(" hundred ")
  }
  if (tens > 0) {
    sb.append(convertTwoDigit(number / 10)).append(" ");
  } else {
    sb.append(convertOnes(ones)).append(" ");
  }
  return sb.toString();
}

String convertTwoDigit(long number) {
  // Optionally throw if number > 99
  // Determine if we are in the 10-19 (which would use a switch-case)
  if (number < 20) return ...
  String multipleOfTen = ... // eg. "forty"
  int ones = number % 10;
  if (ones == 0) {
    return multipleOfTen;
  }
  return multipleOfTen + "-" + convertOnes(ones);
}

String convertOnes(long number) {
  // This one can be a simple switch-case IMO
  ...
}

File: exercises/say/.meta/src/reference/java/Say.java [hunk 5]
```diff
@@ -0,0 +1,204 @@
...
+                    sayDigits(blockDigits, processedDigits.length - index));
+        }
+        return message;
+    }
+
+    private String sayDigits(char[] digits, int orderNumber) {
+        String sentence = "";
+
+        if (digits.length == 3) {
+            digits = filterInitialZeros(digits);
```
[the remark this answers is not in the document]
jmrunkle: Annoyingly, there is a conflict here with the track policies:

> Avoid using final in user-facing code. Liberal use of the `final` keyword remains controversial. It adds little value to test suites, and if included in starter implementations, places implicit restrictions on user solutions.

My work has more-or-less the same policy (only using the final keyword for class constants and constructor assigned member variables). Even at that, we still look at it as a bad practice to re-assign a parameter because you can lose track of the reassignment. Someone comes along later to edit the same code and does not expect the parameter to have changed so they try to check something about the input and then they debug for a while until they discover that somewhere else the input parameter has been reassigned. Principle of least surprise says don't do it.

exercises/say/.meta/src/reference/java/Say.java
jmrunkle: I find it unusual that we officially require less rigorous code for our canonical solution than we do for our students. In particular, because the canonical solution has to be maintained afterwards by the track maintainers. For example, when the problem specifications add new canonical test cases, someone else is going to have to come along and may have to modify this code to add the new test cases.

That said, I understand that there are conflicts of interest here since we want to encourage contributions. I'm just surprised that we allow so much less rigor on our canonical implementations.

jmrunkle:
> Nice of you to get back to this, @jmrunkle. Much needed.
> 
> I won't comment much on the code itself, but more on the process.
> 
> It would seem that a good strategy forward is for @mirkoperillo to decide what amount of feedback he wishes to incorporate in the reference solution. It could definitely be improved, but it shouldn't have to be perfected.

Fair enough. I was not aware that we officially require so much less of our canonical / reference implementations.

mirkoperillo:
Hi guys, thank you for the review , I will try to integrate your suggestions ASAP

config.json
```diff
@@ -235,6 +235,16 @@
         "strings"
       ]
     },
+    {
+      "slug": "say",
+      "uuid": "3c76a983-e689-4d82-8f1b-6d52f3c5434c",
+      "core": false,
+      "unlocked_by": null,
```
jmrunkle, line 242: I did not notice this earlier, but this should be unlocked by one of the core exercises. For difficulty 3, I would suggest either one of the difficulty 2 or 3 exercises should unlock this one.

jmrunkle:
> Hi guys, thank you for the review , I will try to integrate your suggestions ASAP

Any chance you are going to get around to this soonish?

msomji (approved):
LGTM

Issue #1773, opened 2019-10-13
Gradle test should show system streams (out, err)


bmuskalla:
When running unit tests for an exercise with Gradle, system out and system err are suppressed (Gradle default). Especially for beginners this might be very confusing behavior.

Happy to take this on if others agree

lemoncurry:
I think this would be a very sensible change and I would be happy about it.
Other opinions @exercism/java ?

FridaTveit:
Sounds good 🙂

PR #1774, opened 2019-10-16
Show standard streams in build (#1773)

bmuskalla:
Enable System.out and System.err to be shown during `gradle build` without additional logging parameters. Also used the short stacktrace given that gives enough information where to look for the exception without overwhelming newcomers. If necessary, people can still use `gradle build -i` to get all information as suggested by gradle on failures.

### DIFF
--- a/_template/build.gradle
+++ b/_template/build.gradle
@@ -12,7 +12,8 @@ dependencies {
 
 test {
   testLogging {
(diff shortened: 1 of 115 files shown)


bmuskalla:
@lemoncurry I think you forgot to merge the PR :)

Issue #1778, opened 2019-11-01
Naming scheme in Grains exercise. 


ViralTaco:
Hi, 

I was trying to solve an exercise today in the java course. Namingly: "grains".
In the tests source file I was greeted by this line: `BigInteger result = grains.computeNumberOfGrainsOnSquare(1);` (cf: [source](https://github.com/exercism/java/blob/8b7f8def3e0761457bb3d0c4a26d49e2db230207/exercises/grains/src/test/java/GrainsTest.java#L28)) and frankly I was left demotivated and appalled. 

What this line should say is: `BigInteger result = Grains.onSquare(1);` 
And don't even begin to pretend that it's not _just as clear_. "Grains on square" I mean you'd have to be… I'll just stop myself there. 

Sloppy overly explicit names like that tell a message to students. It screams "I don't care, you know what I mean." it says "don't try". 

Honestly whoever pushed that should be ashamed not to have taken 10 minutes of their time to think "how can I make that simpler and/or cleaner?". 

If you need my help to refactor the code I'd love to help. 

Thank you for your time.

FridaTveit:
Thanks for wanting to improve this exercise @ViralTaco 🙂I'm sorry that this made you demotivated. We always want to make sure our code is as clear as possible, so suggestions for improvements are always welcome 🙂However, in this case I think `computeNumberOfGrainsOnSquare` is more descriptive than just `onSquare`. I'm happy to discuss this further, however please try and keep your tone friendly and bear in mind that opinions other than your own are also valid 🙂

ViralTaco:
Ok. Your opinion is invalid. But I understand the misunderstanding. 
I agree that `computeNumberOfGrainsOnSquare` is more descriptive than `onSquare`.
That is true. `Grains grains = new Grains;` followed by `grains.computeNumberOfGrainsOnSquare`
is not more descriptive than `Grains.onSquare` where `onSquare` is a static method of the class `Grains`. 

- There is no reason to instantiate a class with no state. 
- "Compute" what else? What else can you do in Java on a computer than to compute something? 
If you call a method that doesn't compute then you have a serious problem and I'm afraid I'm not qualified to give medical advice. 
- "Number" because `BigInteger` might be a potato. 
- "ofGrains"… "Grains.ofGrains…"

[...]

PR #1779, opened 2019-11-01
"Grains" Simpler method names

ViralTaco:
Hi, 

Following the issue #1778 I thought it'd be better to discuss the issue over some concrete code. 
This is just changing the name of the methods to make them simpler.

Thank you for your time. 

---

Reviewer Resources:

### DIFF
--- a/exercises/grains/.meta/src/reference/java/Grains.java
+++ b/exercises/grains/.meta/src/reference/java/Grains.java
@@ -1,14 +1,22 @@
 import java.math.BigInteger;
+import java.util.stream.IntStream;
 
 class Grains {
-    BigInteger onSquare(final int square) {
-        if (square < 1 || square > 64) {
+
+    BigInteger computeNumberOfGrainsOnSquare(final int square) {
+        if (1 <= square  && square <= 64) {
+            return BigInteger.valueOf(2).pow(square - 1);
+        } else {
             throw new IllegalArgumentException("square must be between 1 and 64");
         }
-        return BigInteger.ONE.shiftLeft(square - 1);
     }
-    
-    BigInteger onBoard() {
-        return new BigInteger("18446744073709551615");
+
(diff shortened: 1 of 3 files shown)


ViralTaco:
I understand that others opinion are valid. You didn't give me an opinion to work with. 
You said the names currently in the code are more descriptive. It's true, they are more verbose but if that's the standard then why stop there? You can go in a lot more detail. 

My argument is that there is no additional information in the longer method name. "Number" is vaguer than `BigInteger` (which the method returns) "Grains" is the name of the class and "of" can be omitted since "grains" can. 
It's a bit like calling `Integer.parseInt` "Integer.getNumberFromParsedString" it's more descriptive but in the context in which it is used it doesn't provide the reader with any more information than they already had, don't you agree?

FridaTveit:
I gave you the opinion that the existing names are more descriptive 🙂Or to paraphrase (and to answer your question of "why stop there"), I don't think the names you were suggesting were descriptive enough. And also, descriptive is not the same as verbose. You can have a very long name that's not at all descriptive, and you can have a short name that's very descriptive 🙂 

One thing is the context where you're calling the method. In that case, as you've already mentioned, naming the variable anything other than `grains` could make the intention of the method unclear. 

There's also the context where you're looking at the method in the class where it's defined. I think you should be able to tell at a glance what that method does, and just `onSquare` doesn't really tell me that.

I do agree that `compute` can probably be omitted. I'd probably prefer for the method to be named `numberOfGrainsOnSquare`. But I can see the argument for omitting `numberOf` so I'd be happy to approve renaming it to `grainsOnSquare` (and `grainsOnBoard`) 🙂

ViralTaco:
> Looks great, thanks 🙂 And thanks for listening to my comments and changing your tone 🙂

No problem, sorry for the tone

PR #1782, opened 2019-11-08
add the empty unit test for the two-fer of the java track

xinri:
---

Reviewer Resources:

### DIFF
--- a/exercises/two-fer/.meta/src/reference/java/Twofer.java
+++ b/exercises/two-fer/.meta/src/reference/java/Twofer.java
@@ -1,5 +1,6 @@
 class Twofer {
     String twofer(String name) {
-        return "One for " + (name != null ? name : "you") + ", one for me.";
+        final String nameToDisplay = name == null || name.isEmpty() ? "you" : name;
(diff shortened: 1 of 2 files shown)


Smarticles101:
Hey @exercism/java this has been discussed before. Hence my comment from exercism/exercism#5113

> Java maintainer here 👋 
>
> In open source, it's worth it to check if an issue has come up in the past before. This has been brought up in multiple repos in including java.
> 
> Here's [java/#733 (pr)](https://github.com/exercism/java/pull/733) and [java/#728 (issue)](https://github.com/exercism/java/issues/728). The decision here was to treat the empty string as a name and output `"One for , one for me."` Personally, I agree with this decision under the basis that if a language can differentiate between an empty string and null, it should treat the empty string as a name as backed up in (problem-specifications/#290)[https://github.com/exercism/problem-specifications/issues/290#issuecomment-230727326].

My question is: should we revisit this? I'd like to get a feel for everyone's opinion on the matter.

xinri:
I think it is better to drop this and keep the current behavior

PR #1785, opened 2019-11-20
Remove "Use `gradlew.bat` if you're on Windows"

sshine:
It seems that the advice to use `gradlew.bat` is no longer in accordance with the advice given in https://exercism.io/tracks/java/installation on how to install Gradle via `choco install gradle` and then execute it via `gradle test`.

Perhaps a previous version of this advice asked students to download `gradlew` or `gradlew.bat` from the track repo's root, where these are still present. Perhaps for this reason, `/gradlew` and `/gradlew.bat` should also be removed.

**Note:** While this updates READMEs in the track repo, I *think* this PR will not propagate the updated README.md's to students before some change is made in the test file.

---

Reviewer Resources:

### DIFF
--- a/config/exercise-readme-insert.md
+++ b/config/exercise-readme-insert.md
@@ -14,8 +14,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/accumulate/README.md
+++ b/exercises/accumulate/README.md
@@ -41,8 +41,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/acronym/README.md
+++ b/exercises/acronym/README.md
@@ -23,8 +23,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/affine-cipher/README.md
+++ b/exercises/affine-cipher/README.md
@@ -91,8 +91,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/all-your-base/README.md
+++ b/exercises/all-your-base/README.md
@@ -47,8 +47,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/allergies/README.md
+++ b/exercises/allergies/README.md
@@ -107,8 +107,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
 Once you get a test passing, you can enable the next one by removing the
--- a/exercises/alphametics/README.md
+++ b/exercises/alphametics/README.md
@@ -47,8 +47,6 @@ terminal:
 $ gradle test
 ```
 
-> Use `gradlew.bat` if you're on Windows
-
 In the test suites all tests but the first have been skipped.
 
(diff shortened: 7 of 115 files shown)


sshine:
Summary of changes:
 - Removed comment in config/exercise-readme-insert.md
 - Ran `configlet generate .` to remove comment in all README.md
 - Discard other changes caused by `configlet generate .` (updates, inconsistencies)

So while this PR touches on 115 files, this is a cascading, homogenic effect.

mirkoperillo:
Hi,
I'm not sure that moving away from the gradle wrapper is a good idea
Using the gradle wrapper we are sure that everybody will install and use the same gradle version (`gradle-5.6.2-bin.zip` in this moment) and so we avoid possible incompatibility between the local  installation version and  `build.gradle` file of the track.

What do you think @exercism/java ?

sshine:
@mirkoperillo: are you suggesting that the advice should instead be reverted? And can you be sure that that particular version of gradle will work with any version of Java they might have? Are newer versions of gradle not backwards compatible?

I thought that the latest reports of gradle breaking reported on exercism/exercism were caused by Java being too new for the latest version of gradle, rather than gradle being too new for the build.gradle file.

mirkoperillo:
@sshine I thought again about this issue, I don't have a historical background as Java track maintainer so I miss a lot of things for sure.
I want to start with a question: "Why documentation gives instructions to install Gradle and Java without any requirements, meanwhile in the repo there is the Gradle wrapper ?"

In my experience too much freedom about versions of the tools bring to a lot of problems.
I don't know if the probability to break Gradle using a new Java version is bigger than a Gradle version incompatibility. 
Maybe a solution can be decide a Java version and a Gradle version as a requirement for the track.

I think this should be a choral discussion

jmrunkle:
> @sshine I thought again about this issue, I don't have a historical background as Java track maintainer so I miss a lot of things for sure.
> I want to start with a question: "Why documentation gives instructions to install Gradle and Java without any requirements, meanwhile in the repo there is the Gradle wrapper ?"
> 
> In my experience too much freedom about versions of the tools bring to a lot of problems.
> I don't know if the probability to break Gradle using a new Java version is bigger than a Gradle version incompatibility.
> Maybe a solution can be decide a Java version and a Gradle version as a requirement for the track.
> 
> I think this should be a choral discussion

+1 to this advice. Can you open an issue for discussion on what we should document as our recommended practice (re: gradle)?

sshine:
@jmrunkle: How about we have the discussion here?

This appears to be the subject of the PR.

So: I don't code Java daily. I have no idea about what's new and what's solid. Someone who writes Java daily should voice what they think is a good mix between "state of the art" and "doable by anybody new to Java, but not necessarily new to programming".

As someone who writes Java occasionally, I just want to open IntelliJ and hit some button or key combo and see my tests run. If I can't do that, I want at least a command that will do it. And either has to work on Windows, Mac and Linux in mostly the same way.

I imagine this is why there was a `gradlew.bat` once, because some people like that you can run tests without being bound to a particular IDE, and that you need both bash and cmd support. But years later, `gradle` (on Windows short for `gradle.exe`) is in `$PATH` / `%PATH%` and you don't need it.

This is the reasoning I applied.

It's okay that the track does not cater to all possible workflows. With Gradle, hopefully this leaves some room for the problem solver to choose if they want to use the command-line or their IDE regardless of what scripts we provide or not.

jmrunkle:
Although I do code professionally in Java every day, we do not use gradle in my work environment (nor do we use Windows) so I do not really have much of an opinion about the use of gradle.bat. I do not know many programmers who use Windows for their professional software development, but maybe one of the other maintainers has this setup?

Reading over the Java installation instructions for Windows, it looks like we have the students download/install Chocolatey, the JDK and then gradle. Assuming the current version of that gradle installation behaves as @sshine is saying (ie. adds gradle to the PATH), I think we can just go forward with this PR. That's my $0.02.

jmrunkle:
> As mentioned - I think this change is safe. Giving other maintainers a chance to comment as well though.

We will give the rest of the maintainers a week to add thoughts / responses. If nobody raises any objections by April 20, we can merge this PR.

sshine:
After five months, another week will probably not hurt.

But on this note, we should probably propose a better way to deal with stale tracks.

jmrunkle:
> But on this note, we should probably propose a better way to deal with stale tracks.

Do you mean dealing with tracks that do not have a lot of active maintainers or... what exactly?

mirkoperillo:
I work every day in Java for a long time, but I'm not a gradle user and I'm not a Windows user.
I thought maintaining the wrapper was the better solution for the "I want only to solve the exercise" user.

I think we have the problem that the track doesn't have anymore senior maintainers knowing the history of the past choices.

So to not stay in stale, we can try to apply this PR and look what it will happen

jmrunkle:
Merging this, thanks for everyone that contributed to the conversation.

Issue #1783, opened 2019-11-08
Two-fer exercise in java track, no unit test for when the input is empty 


xinri:
Hello,

I am a mentor for the java track. When I do the review, I always see that people forgot the empty string. When I check the unit tests, there is no test for this case and in the instructions, it is mentioned.

jmrunkle:
I know that this discussion is old, but the topic has been left open. Since we (the maintainers) have apparently decided to have the empty string case explicitly return "One for , one for me.", I'm actually debating opening up a PR to have a custom test case that specifically tests for *that*. If there are no complaints, I'll at least make the simple PR.

jmrunkle:
As a note, here is the official contributing policy on adding non-canonical tests:

> Check if there is canonical data for the exercise you're adding. This can be found at https://github.com/exercism/problem-specifications/tree/master/exercises/EXERCISE-SLUG/canonical-data.json. If there is canonical data for your exercise then you should follow this when making the tests. We aim to follow the canonical data as closely as possible in our tests to ensure thorough test coverage. If there is canonical data available you also need to create a file at exercises/exercise-slug/.meta/version specifying the canonical data version you have implemented (e.g. 1.0.0). The canonical data version can be found at the top of the canonical data file for that exercise. See other exercises, e.g. acronym, for an example version file.

This says nothing about adding extra track-specific tests and this feels like a good case where we should enshrine a decision that we specifically made.

jmrunkle:
Closing this issue since we now have an official test for the empty input case - clarifying the exact behavior we expect.

PR #1802, opened 2020-04-15
Explicitly test that the empty string is not treated specially in twofer

jmrunkle:
We decided in issue #1783 (and previously) that the empty string should not be treated as a special case like null. As such, I think we should have a test for it.

---

Reviewer Resources:

### DIFF
--- a/exercises/two-fer/src/test/java/TwoferTest.java
+++ b/exercises/two-fer/src/test/java/TwoferTest.java
@@ -38,4 +38,16 @@ public class TwoferTest {
 
         assertEquals(expected, twofer.twofer(input));
     }
+
+    /* Track specific */
+
+    @Ignore("Remove to run test")
+    @Test
+    public void emptyStringIsNotTheSameAsNull() {
+        String input = "";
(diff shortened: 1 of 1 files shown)


exercises/two-fer/src/test/java/TwoferTest.java
```diff
@@ -38,4 +38,16 @@ public void anotherNameGiven() {
 
         assertEquals(expected, twofer.twofer(input));
     }
+
+    /* Track specific */
```
sshine, line 42: I don't know if there is a convention for denoting track-specific tests, but this seems fine.

#733 didn't have it, though.

jmrunkle: I like being explicit about it so that it does not get removed without (at least) looking at the context of why it was added.

jmrunkle:
> The advantage of the test is clarity.

This is precisely the point. We are trying to teach and giving students (especially in these early exercises) an ambiguous problem is not helpful.

PR #1841, opened 2020-08-25
set default encoding for all problems to utf-8

msomji:
This Pr addresses issue raised in  https://github.com/exercism/exercism/issues/5353

---

Reviewer Resources:


msomji:
Yes I am not 100 convinced that this will work either, if this doesn't work, we might have to add those two lines to every build.gradle file for each exercise

Issue #1855, opened 2021-01-13
Hamming exercise ; Exceptions & Readme + Instructions


Joker018:
Hi,

In the **Hamming exercise**, we have to throw exceptions for different situations :
- If the two strands are not the same length (IllegalArgumentException("leftStrand and rightStrand must be of equal length."))
- If the left strand is empty but not the right strand (IllegalArgumentException("left strand must not be empty."))
- If the right strand is empty but not the left strand (IllegalArgumentException("right strand must not be empty.")).

But I kind of don't really see why we should throw some particular exceptions for the last two cases mentioned as those are included in the first case. So I feel like it might be best to delete the tests concerning those last two cases.

Moreover, in the _instructions_, the _readme_ and the _tips_ section, it is never mentioned what type of exception should be thrown and with what messages. So wouldn't it be best to also detail what kind of exception and what message should be thrown ? As it is now, we have to read the tests to see that information which kind of look like just hardcoding to pass the test in the end ^^'

My proposition would be to : 
1. Delete the tests concerning the two last exceptions. (considering these are included in the first exception).
2. Add in the _readme_, and in the _instructions_ what kind of exception should be thrown for the first case as well as the message the exception should contain.

If what I propose above isn't ok, I think it would still be best to at least update the _readme_ and _instructions_ accordingly. 

Of course I would be more than happy to do the changes suggested.

Thanks in advance for all of your opinions on the subject.

sanderploegsma:
Right now the exercise tests are in line with the canonical data, which explicitly describe the cases where one strand is empty while the other one isn't, so I'm not in favor of dropping that requirement in the Java track.

> Moreover, in the instructions, the readme and the tips section, it is never mentioned what type of exception should be thrown and with what messages. So wouldn't it be best to also detail what kind of exception and what message should be thrown ? As it is now, we have to read the tests to see that information which kind of look like just hardcoding to pass the test in the end ^^'

This is not a bad point actually. Generally in Java the `IllegalArgumentException` is used when a provided method argument doesn't meet the requirements, but this is not explicitly mentioned in the exercise instructions or in the `exceptions` concept, so how is a student to know this?

[...]

PR #1856, opened 2021-01-14
Fix issue #1855 (add info about exceptions + removal of tests)

Joker018:
This commit fixes the issue #1855. More precisely, it add some info about
the kind of exception that should be thrown for the hamming problem as well
as the message that should be send.
The tests checking if one strand is empty and not the other were also removed
as they did not really seem relevant.

---

Reviewer Resources:

### DIFF
--- a/exercises/hamming/.meta/hints.md
+++ b/exercises/hamming/.meta/hints.md
@@ -8,5 +8,5 @@ We use JUnit's [`ExpectedException`](http://junit.org/junit4/javadoc/4.12/org/ju
 [rule](https://github.com/junit-team/junit4/wiki/rules) throughout the track to verify that the exceptions you throw
 are:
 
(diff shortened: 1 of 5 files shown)


jmrunkle:
Mind synchronizing the files so that they can be merged cleanly? I think this PR is still valid, but I'll probably close it in a week if there is no response (just trying to get things into a clean state). Just reply if you still plan to update this PR.

jmrunkle:
> Mind synchronizing the files so that they can be merged cleanly? I think this PR is still valid, but I'll probably close it in a week if there is no response (just trying to get things into a clean state). Just reply if you still plan to update this PR.

Closing this since there has been no followup. Feel free to re-open if you are willing deal with the updates necessary to merge this cleanly.

PR #1964, opened 2021-07-28
Add icons for key features

ee7:
The icons for key features are now available, and `configlet lint` now requires that each key feature has a valid `icon` value.

This PR sets an initial icon for each key feature - feel free to change them however you want. You can see the list of available icons [here](https://github.com/exercism/docs/issues/189). Note that an icon can be used regardless of its name.

With the initial state of this PR, the key features look something like the below.

---

evolving Modern
Java is a modern, fast-evolving language with releases every 6 months.

statically-typed Statically-typed
Every expression has a type known at compile time.

multi-paradigm Multi-paradigm
Java is primarily an object-oriented language, but has many functional features introduced in v1.8.

general-purpose General purpose
Java is used for a variety of workloads like web, cloud, mobile and game applications.

portable Portable
Java was designed to be cross-platform with the slogan "Write once, run anywhere".

garbage-collected Garbage Collection
Java programs perform automatic memory management for their lifecycles.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1934,32 +1934,32 @@
   ],
   "key_features": [
     {
-      "icon": "",
+      "icon": "evolving",
       "title": "Modern",
       "content": "Java is a modern, fast-evolving language with releases every 6 months."
     },
     {
-      "icon": "features-typed",
+      "icon": "statically-typed",
       "title": "Statically-typed",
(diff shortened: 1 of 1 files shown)


mirkoperillo (approved):
@ee7 LGTM

PR #1968, opened 2021-08-17
Added more exhaustive test case for isbn verifier

ericbalawejder:
# pull request

The isbn verifier tests did not check for a valid isbn number padded with addition characters. Solutions were able to grab the valid isbn number from inside the string. This was allowing invalid string input with a valid isbn inside to pass. This test case resolves that issue.

---

Reviewer Resources:


mirkoperillo (approved):
LGTM

Issue #1965, opened 2021-08-15
New test cases for secret-handshake


Minification:
While mentoring, I have noticed that a lot of solutions do not handle numbers whose binary representation has more than five digits, because the test cases where the binary representation has a fifth digit _all_ test for reversal. Additionally, there are no test cases that have a sixth digit, but do not require reversal. 

That is why I propose addition of the following test cases:

```
@Ignore("Remove to run test")
@Test
public void testThatHandlesMoreThanFiveBinaryPlacesWithReversal() {
    assertEquals(
            asList(Signal.DOUBLE_BLINK, Signal.WINK),
            handshakeCalculator.calculateHandshake(51));
}
```

```
@Ignore("Remove to run test")
@Test
public void testThatHandlesMoreThanFiveBinaryPlacesWithoutReversal() {
    assertEquals(
            asList(Signal.WINK, Signal.DOUBLE_BLINK),
            handshakeCalculator.calculateHandshake(35));
}
```

I would add these myself, but I'm not quite set up for that yet and there was no issue for this yet, hence the creation of the new issue.

ericbalawejder:
Thank you. These are good test cases because they may change how someone handles the logic. Testing for input that is greater than 16 is different than bitwise `&` checking for 16. 111 is also a good input to check for a value with 7 binary digits but no 16 bit for reversal. 

Feel free to submit a PR with these tests if you like. The changes won't be visible until Exercism v3 is released however. If you don't feel set up to make a PR, one of the maintainers can handle it. Thank you for your contributions.

Minification:
Thanks, I have added the test case you suggested plus the original ones suggested in this issue. Now it looks like the workflow is still awaiting approval and we need a review.

PR #1970, opened 2021-08-20
Add more exhaustive tests for secret handshake

Minification:
# Pull request fixes exercism/java#1965

Some solutions simply checked if the input is greater than 16 to reverse, thus wrongly reversing the sequence for numbers greater than 16 whose bit at position 4 is not set. These test cases enforce that solutions check the bit at position 4 explicitly.

---

Reviewer Resources:


ericbalawejder (approved):
Looks good! Thank you for your contribution.

PR #1977, opened 2021-09-01
replaced TODO with blurbs from exercises these were forked from.

pdmoore:
# pull request

I copied the blurb text from the exercises these were forked from to replace the "TODO" text that was there. Most from csharp track, one from go. Don't know where the salary-calculator exercise came from so left it as a TODO.

---

Reviewer Resources:


jmrunkle (approved):
Thanks for doing this!

PR #1981, opened 2021-09-03
avoid deprecation warnings

LarsEckart:
a student was confused by the compiler warning and I think it is even displayed as error in the UI. So let's use `assertArrayEquals` instead of `assertEquals(Object[], Object[])` which is deprecated..

# pull request

---

Reviewer Resources:


mirkoperillo (approved):
LGTM thank you @LarsEckart

PR #1988, opened 2021-09-04
Fix broken Java CI with Gradle workflow

### LINKED ISSUE #1986: "Java CI with Gradle" workflow is not running properly
The [gradle workflow](https://github.com/exercism/java/blob/main/.github/workflows/gradle.yml) has a few issues:

1. It is specifically specified to run only on the master branch (which no longer exists).
2. It still runs configlet linting in a broken way which now has its own workflow.

If I find anything else I will try to fix that, but there might be some other things broken that we need to fix when we update this.

jmrunkle:
# Fix broken Java CI with Gradle workflow

* Adds canonical triggers (including manual launch).
* Removes lint since we have a separate workflow / action for Configlet.
* Fixes any pre-existing issues in the repo that break the build.

Fixes #1986 

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Compile and checkstyle with Gradle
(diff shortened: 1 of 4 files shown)


iHiD:
This is great. Thanks!!

PR #2000, opened 2021-09-06
Cleanup concepts `arrays`, `for-loops` and `foreach-loops`

### LINKED ISSUE #1997: Cleanup: concepts `arrays`, `for-loops` and `foreach-loops`
There are a few problems that need to be addressed across these concepts:

1. TODOs like the ones in introduction.md 
2. Isolate the dumped information from the main exercise into each section (arrays, for loops, and for-each loops).
3. These all seem to be missing `blurb`s in the .meta/config.json files

There may be other issues, but I will content myself with solving these three things for now.

jmrunkle:
Fixes #1997

# pull request

Fixes the main issue and adds some consistency between introduction.md and about.md per https://exercism.org/docs/building/tracks/concepts.
We might later limit the information provided in the introduction, but this is a good starting point.
Also adds blurbs that were missing.

---

Reviewer Resources:


concepts/arrays/about.md
```diff
@@ -1,37 +1,65 @@
...
+Accessing an index that is outside of the valid indexes for the array results in an `IndexOutOfBoundsException`.
+
+Arrays can be manipulated by either calling an array instance's methods or properties, or by using the static methods defined in the `Array` class (typically only used in generic code).
+The most commonly used property for arrays is its length which can be accessed like this:
+
+```java
+int arrayLength = someArray.length;
+```
+
+Java also provides a helpful utility class [`java.util.Arrays`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html) that has lots of useful array-related methods (eg. `Arrays.equals`).
```
kotp, line 53:
```suggestion
Java also provides a helpful utility class [`java.util.Arrays`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html) that has a lot of useful array-related methods (eg. `Arrays.equals`).
```

concepts/arrays/introduction.md
```diff
@@ -1,34 +1,65 @@
...
+Elements can be assigned to an array or retrieved from it using an index.
+Java arrays use zero-based indexing: the first element's index is 0, the second element's index is 1, etc.
+
+Here is the standard syntax for initializing an array:
+
+```java
+type[] variableName = new type[size];
+```
+
+The `type` is the type of elements in the array which can be a primitive type (eg. `int`) or a class (eg. `String`).
```
kotp, line 14:
```suggestion
The `type` is the type of elements in the array which may be a primitive type (e.g. `int`) or a class (e.g. `String`).
```

concepts/arrays/introduction.md
```diff
@@ -1,34 +1,65 @@
...
+
+```java
+// Two equivalent ways to declare and initialize an array (size is 3)
+int[] threeIntsV1 = new int[] { 4, 9, 7 };
+int[] threeIntsV2 = { 4, 9, 7 };
+```
 
+As the compiler can now tell how many elements the array will have, the length can be omitted.
+
+Array elements can be assigned and accessed using a bracketed index notation:
```
kotp, line 34:
```suggestion
Array elements may be assigned and accessed using a bracketed index notation:
```

concepts/for-loops/.meta/config.json
```diff
@@ -1,9 +1,8 @@
...
+  "blurb": "For loops allow for iteration over a specified range.",
   "authors": [
     "samuelteixeiras",
     "ystromm"
   ],
-  "contributors": [
-  ]
-}
+  "contributors": []
+}
```
kotp, line 8: Missing end of line for this file.  Generally, text files are expected to not have data after the last line.

concepts/for-loops/about.md
```diff
@@ -1,61 +1,67 @@
...
+/* =>
+1
+4
+9
+16
+*/
 ```
 
-A `for` loop does have some advantages over a `foreach` loop:
+If iterating through every element in a collection, a for-each loop is preferred, but it can be done with a for loop like this:
```
kotp, line 53:
```suggestion
If iterating through every element in a collection, a `for-each` loop is preferred, but it can be done with a `for` loop like this:
```

concepts/for-loops/introduction.md
```diff
@@ -1,54 +1,69 @@
...
 ```java
-// Declare array with explicit size (size is 2)
-int[] twoInts = new int[2];
+int i = 1
+```
 
-// Assign second element by index
-twoInts[1] = 8;
+The `test` expression tests if the loop should end.
+If it evaluates to true, the body and then the update expression will be executed.
```
kotp, line 21:
```suggestion
If it evaluates to `true`, the body and then the update expression will be executed.
```

concepts/for-loops/introduction.md
```diff
@@ -1,54 +1,69 @@
...
+If it evaluates to false, neither the body nor the update statement will be executed and execution resumes after the loop's closing bracket.
```
kotp, line 22:
```suggestion
If it evaluates to `false`, neither the body nor the update statement will be executed and execution resumes after the loop's closing bracket.
```

concepts/for-loops/introduction.md
```diff
@@ -1,54 +1,69 @@
...
-// Two equivalent ways to declare and initialize an array (size is 3)
-int[] threeIntsV1 = new int[] { 4, 9, 7 };
-int[] threeIntsV2 = { 4, 9, 7 };
+i++
 ```
 
-Arrays can be manipulated by either calling an array instance's methods or properties, or by using the static methods defined in the `Array` class.
-
-The fact that an array is also a _collection_ means that, besides accessing values by index, you can iterate over _all_ its values using a `foreach` loop:
+A for loop printing out the first four squares would look like this:
```
kotp, line 38:
```suggestion
A `for` loop printing out the first four squares would look like this:
```

concepts/for-loops/introduction.md
```diff
@@ -1,54 +1,69 @@
...
-// => aeiou
+```
+square of 1 is 1
+square of 2 is 4
+square of 3 is 9
+square of 4 is 16
 ```
 
-If you want more control over which values to iterate over, a `for` loop can be used:
+If iterating through every element in a collection, a for-each loop is preferred, but it can be done with a for loop like this:
```
kotp, line 55:
```suggestion
If iterating through every element in a collection, a `for-each` loop is preferred, but it can be done with a `for` loop like this:
```

concepts/for-loops/introduction.md
```diff
@@ -1,54 +1,69 @@
...
+    System.out.print(array[i]);
 }
-
-// => aei
 ```
+
+A `for` loop does have some advantages over a `for-each` loop:
+
+- You can start or stop at the index you want.
+- You can use any (boolean) termination condition you want.
```
kotp, line 66:
```suggestion
- You can use any (Boolean) termination condition you want.
```

concepts/foreach-loops/.meta/config.json
```diff
@@ -1,9 +1,8 @@
...
+  "blurb": "For-each loops allow for iteration over every element in a collection.",
   "authors": [
     "samuelteixeiras",
     "ystromm"
   ],
-  "contributors": [
-  ]
-}
+  "contributors": []
+}
```
kotp, line 8:
```suggestion
}

```

concepts/foreach-loops/about.md
```diff
@@ -1,49 +1,56 @@
...
-    // body
+    body;
 }
 ```
 
-The declaration part declares the variable used to hold the values assigned from the collection:
+The `declaration` declares the variable used to hold the values assigned from the collection.
 
-The collection is an array or a collection holding the values that will be assigned to the loop variable.
+The `collection` is an array or collection holding the values that will be assigned to the loop variable.
```
kotp, line 16:
```suggestion
The `collection` is an array or a collection holding the values that will be assigned to the loop variable.
```

The "a" is optional, but in formal writing it is generally used.

concepts/foreach-loops/introduction.md
```diff
@@ -1,54 +1,56 @@
-# Introduction
+# Introduction to For-Each Loops
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
+The [for-each loop](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/for.html) provides a mechanism for executing a block of code for each element in a collection.
+Some documentation (eg. Oracle's documentation) refers to these as enhanced for loops.
```
kotp, line 4: Link here for Oracle's documentation, maybe?

kotp (commented):
Looks pretty nice.

There seems to be a lot of duplication in Introduction and About documents.

There seems to be some missing EOL (end of line) characters for some text files.

It matters, but I would not necessarily not approve only for that.  It can make my life harder though.  https://stackoverflow.com/a/729795/485864 is a good reference, with some more official links nearby and additional information.

concepts/for-loops/.meta/config.json
jmrunkle: This was done automatically by the VSCode JSON formatter...

concepts/for-loops/introduction.md
jmrunkle: Not sure I agree with this one. Using the uppercase Boolean makes it seem like you are referring specifically to the class, but any primitive boolean would do just as well.

concepts/foreach-loops/introduction.md
jmrunkle: The link in the previous sentence _is_ to Oracle's documentation so it seems redundant to double the link. The main reason I mention it is because Oracle's documentation does not call it a for-each loop.

jmrunkle:
> There seems to be a lot of duplication in Introduction and About documents.

Yeah, that is actually intentional at the present. We are somewhat punting on deciding what information is actually necessary for the introduction. Figured duplicating is a good short term solution while we build out and we can revise later. I did mention it in the PR description: "We might later limit the information provided in the introduction, but this is a good starting point."

concepts/for-loops/introduction.md
kotp: It is a persons name, and in English, at least, Boolean is what I believe is used.  In Ruby a generic "boolean" would be anything, with only `nil` and `false` not evaluating as truthy.

In Ruby, it is the only place I do not use a capital "B" for Boolean when I am talking about Boolean in the less strict sense as we can because of the truthy/falsey nature that is allowed, otherwise I have used Boolean capitalized since gradeschool, and though electronics schooling.

I may be the exception though.  But I do not think so.

[This English SE entry](https://english.stackexchange.com/a/4491/83340) has more authoritative references than I would contain as an English as a Not First Language person.

concepts/foreach-loops/introduction.md
kotp: There you go.  I missed the link reference.  Since "some" is mentioned, is there a link where it does not do so?  Or should we say "The documentation" or "Oracle's documentation" instead?  Then it is an example, but it is explicitly riding on the essence of the link given?

kotp:
>  I did mention it in the PR description: "We might later limit the information provided in the introduction, but this is a good starting point."

I did read "limit the information" but did not translate it to "we might deduplicate the information".  Good enough!

concepts/for-loops/.meta/config.json
kotp: Yeah, the VSCode editor makes this mistake I noticed last week while pairing with a student.  It is unfortunate.

kotp (approved):

concepts/for-loops/introduction.md
jmrunkle: Actually that is technically incorrect. The name of the individual for which booleans are named is George Boole.

In Java specifically, Boolean refers to the boxed class for the primitive boolean (like Integer refers to the boxed class for the primitive int). Therefore, I think boolean should be used because we do not specifically mean the boxed Boolean class.

concepts/foreach-loops/introduction.md
jmrunkle: The markdown file we are commenting on (and exercism in general) would be the principal place where we use the terminology of for-each loops. I am sure there are more places, but have no need nor desire to track them down.

I think you may be right though, we could just say "Oracle's documentation" in order to achieve the same purpose I had in mind. I'll update the PR tomorrow to address that.

concepts/for-loops/introduction.md
jmrunkle: I can code-font `boolean` if that would help clarify things.

PR #2004, opened 2021-09-06
Use info rather than stacktrace for gradle

jmrunkle:
--info returns more helpful information than --stacktrace

# pull request

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Compile and checkstyle with Gradle
-      run: cd exercises && ../gradlew check compileStarterSourceJava --parallel --continue --stacktrace
+      run: cd exercises && ../gradlew check compileStarterSourceJava --parallel --continue --info
 
   test:
 


SleeplessByte (approved):
Might be nice for future readers to see a comparison output :)

jmrunkle:
> Might be nice for future readers to see a comparison output :)

Not sure how long these live, but here is what the stacktrace version looks like: https://github.com/exercism/java/runs/3521814015?check_suite_focus=true

It does give you the failure:

> Error: eckstyle] [ERROR] /home/runner/work/java/java/exercises/practice/transpose/src/test/java/TransposeTest.java:50: 'method def rcurly' has incorrect indentation level 8, expected level should be 4. [Indentation]

But it also includes a huge not-so-helpful stack trace:

```
Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':practice:transpose:checkstyleTest'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$3(ExecuteActionsTaskExecuter.java:186)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:268)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:184)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:173)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        ... <about another 100 lines of output for the stack trace>
```

I get the feeling that `--stacktrace` is more meant for Gradle devs rather than for people using Gradle.

`--info` gives us a more limited amount of information which includes the parts that I care about (eg. the failure for TransposeTest.java).

SleeplessByte:
Yeah, I agree with you :)

PR #2012, opened 2021-09-11
Changed return type of race to void in TestTrack.java

cjwomack:
There's no indication in the examples that TestTrack.race returns anything.... and doesn't make sense given all the other possible return values given the set of other potential return variables all have type int.


jmrunkle (changes requested):
Mind also updating the same method in the reference solution? https://github.com/exercism/java/blob/main/exercises/concept/remote-control-competition/.meta/src/reference/java/TestTrack.java

jmrunkle:
For the record, I am planning to do a little re-visiting of this exercise to improve the instructions and a few other things to improve this exercise.

jmrunkle (approved):

PR #2022, opened 2021-09-14
1861 exception handling

rv02:
# pull request

---

Reviewer Resources:


exercises/concept/calculator-conundrum/.meta/design.md
```diff
@@ -0,0 +1,23 @@
...
+* exception-handling
+* try-catch
+* throw and throws keyword
+* user-defined exceptions
+
+## Prerequisites
+
+* `basics`: know how to define methods.
+* `conditionals`: know how to do conditional logic.
+
```
jmrunkle, line 23: switch-case is another prerequisite here

exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -0,0 +1,15 @@
+# Hints
+
+### 1. Implementing operations
```
jmrunkle, line 3: This needs to be a heading 2 (so `##`) and the text needs to exactly match the text in the instructions. Same throughout this file.

exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -0,0 +1,15 @@
...
+
+* Use a `switch-case` block to match different operations and implement them using `SimpleOperation` class.
```
jmrunkle, line 5: Grammar nit:

... using the `SimpleOperation` class.

exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -0,0 +1,15 @@
...
+
+### 2. Handling invalid operation
+
+* Check for `null` operation beforehand and the rest along-with the `switch-case` block.
```
jmrunkle, line 9: I think this first bullet point should probably be split:

* Check for `null` operations before the switch.
* Add a case to the switch for the empty String.
* Use a default for anything else at the end of the switch.

exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -0,0 +1,15 @@
...
+* Throw an `IllegalOpertionException` with appropriate messages in different cases.
+
+### 3. Handling Division by zero
+
+* Use a `try-catch` block to catch `ArithmeticException`
+* Pass both message and cause as argument when throwing exception.
```
jmrunkle, line 15: Since all of our constructors take a message, just be explicit about passing the `ArithmeticException` as the `cause` parameter.

Also, grammar nit:

... when throwing the exception.

exercises/concept/calculator-conundrum/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
+# Instructions
+
+In this exercise you will be building error handling for a simple integer calculator. 
+To make matters simple, methods for calculating addition, multiplication and division are provided.
+
+The goal is to have a working calculator that returns a String with the following pattern: `16 + 51 = 67`, when provided with arguments `16`, `51` and `+`.
+
```
jmrunkle, line 7: nit: extra newline here can be deleted

exercises/concept/calculator-conundrum/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+obj.calculate(32, 6, "*"); // => returns "32 * 6 = 192"
+
+obj.calculate(512, 4, "/"); // => returns "512 / 4 = 128"
+```
+
+## 1. Implement the method calculate to support a few basic operations 
+
+The main method for implementation in this task will be the `CalculatorConundrum.calculate()` method. 
+It takes three arguments. 
+The first two arguments are integer numbers on which an operation is going to be conducted.
```
jmrunkle, line 23: Slightly better wording choice: is going to operate

It seems slightly better to say that an operation is going to operate rather than the more passive voice "to be conducted"

exercises/concept/calculator-conundrum/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+- multiplication using the `*` String
+- division using the `/` String
+
+## 2. Handle illegal operations
+
+All the following cases need to throw an `IllegalOperationException`:
+
+* when the `operation` argument is `null`
+* when the `operation` is the empty String
+* when the `operation` argument is anything other than `+`, `*`, or `/`
```
jmrunkle, line 36: Since students may not have access to the tests (eg. when using the website editor), we should explicitly say what error message to use in the instructions here.

exercises/concept/calculator-conundrum/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+## 3. Handle the thrown DivideByZero exceptions
+
+Dividing by zero throws an `ArithmeticException` which the calculator needs to catch and then throw an `IllegalOperationException` with the message `Divide by zero operation illegal` and the `ArithmeticException` as its cause.
+This should be handled using `try-catch` block.
```
jmrunkle, line 41: Grammar nit:

... using a `try-catch` block.

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
+# Exception Handling in Java
+
+The Exception Handling in Java is one of the powerful mechanism to handle the runtime errors so that the normal flow of the application can be maintained.
```
jmrunkle, line 3: Grammar nits:

Exception handling in Java is a powerful mechanism to handle errors so that the normal flow of the application can be maintained.

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+
+Good exception handling can handle errors and gracefully re-route the program to give the user still a positive experience.
```
jmrunkle, line 5: To avoid repetition:

Good exception handling can re-route the program to maintain a positive user experience.

---

Also, I would put this in the same paragraph with the first sentence (so no intervening empty line).

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Why use Exception Handling with an example
```
jmrunkle, line 7: Why jump to an H4 (`####`) rather than H2 (`##`)?

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+```java
+public static List<Player> getPlayers() throws IOException {
+    Path path = Paths.get("players.dat");
+    List<String> players = Files.readAllLines(path);
+
+    return players.stream()
+      .map(Player::new)
+      .collect(Collectors.toList());
+}
+```
```
jmrunkle, line 18: This is probably a poor example precisely because it does not actually handle any exceptions. It would be better if we wrapped this and handled the IOException, perhaps logging to stderr and returning an empty list?

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+
+This code chooses not to handle the IOException, passing it up the call stack instead. 
+In an idealized environment, the code works fine.
+
+But what might happen in production if players.dat is missing?
+
+```java
```
jmrunkle, line 25: Not a java, this is a log message.

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+The `catch` statement allows you to define a block of code to be executed, if an error occurs in the try block.
+
+#### Example
+```java
+public int getPlayerScore(String playerFile) {
+    try {
+        Scanner contents = new Scanner(new File(playerFile));
+        return Integer.parseInt(contents.nextLine());
+    } catch (FileNotFoundException noFile) {
+        throw new IllegalArgumentException("File not found");
```
jmrunkle, line 53: In general, it is best to always include the cause for the stack trace. We should teach that here.

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Handling Exceptions
+
+### Try-Catch block
+
+The `try` statement allows you to define a block of code to be tested for errors while it is being executed.
+
+The `catch` statement allows you to define a block of code to be executed, if an error occurs in the try block.
+
+#### Example
```
jmrunkle, line 46: Not sure this heading is really warranted. We can just go straight into an example with a leading sentence like:

For example:

exercises/concept/calculator-conundrum/.docs/intoduction.md
```diff
@@ -0,0 +1,122 @@
...
+        return Integer.parseInt(contents.nextLine());
+    } catch (FileNotFoundException noFile) {
+        throw new IllegalArgumentException("File not found");
+    }
+}
+```
+
+### The Throw/Throws keyword
+
+If a method does not handle a checked exception, the method must declare it using the `throws` keyword.
```
jmrunkle, line 60: This document is missing an explanation of checked vs unchecked exceptions which is important for understanding the throws keyword.

exercises/concept/calculator-conundrum/.meta/config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "blurb": "Learn about exception-handling by making a simple calculator.",
+  "authors": [
+    "rv02",
+    "jmrunkle"
+  ],
+  "files": {
```
jmrunkle, line 7: We need to add an "editor" stanza with the IllegalOperationException and the SimpleOperation java files.

jmrunkle (changes requested):

exercises/concept/calculator-conundrum/.docs/intoduction.md
rv02: This example knowingly does not handle the exception to show an example of the conditions caused in cases like these.

config.json
```diff
@@ -85,6 +85,19 @@
         ],
         "status": "active"
       },
+      {
+        "slug": "calculator-conundrum",
+        "name": "Calculator Conundrum",
+        "uuid": "",
```
jmrunkle, line 91: You get the uuid from configlet:

```
$ bin/fetchconfiglet
$ bin/configlet uuid
```

Issue #1743, opened 2019-10-01
connect: implement exercise


lemoncurry:
The exercise **connect** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/master/exercises/connect). 

How to implement a new exercise for the Java track is described in detail in [CONTRIBUTING.md](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#adding-a-new-exercise). Please have a look there first before starting working on the exercise. 
Also please make sure it is clear that you are currently working on this issue, either by asking to be assigned to it, or by opening an empty PR. 

When opening an PR, please reference this issue using any of the [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords). 

If you have had a look at the exercise description and you concluded that the exercise might not be possible to implement in the Java language, please leave a comment and describe the problem. 

In case you have any further questions, feel free to ask here.

tr808axm:
I'd be happy to work on this one. Can you please assign me?

jmrunkle:
@tr808axm - are you still working on this? It look like you have commits in your private repo, but you have not created a PR for *this* repo.

J-Ankit2020:
I'd be happy to work on this one. Can you please assign it to me?

mirkoperillo:
Hi @J-Ankit2020  ok, I dont' think @tr808axm is working on this anymore

J-Ankit2020:
hey how can i fix this issue

mirkoperillo:
@J-Ankit2020 you have two ways:
* Take the solution started by tr808axm ( there is something here https://github.com/tr808axm/java/commit/54f942facd4516648558492bb68464fd4ad4097a ), check the completeness and if it is complete open a PR about it
* Take the exercise specification (https://github.com/exercism/problem-specifications/tree/master/exercises/connect) and provide your solution

kkyb123:
I am currently working on this one. Please can you assign it to me?

mirkoperillo:
done @kkyb123

kkyb123:
@mirkoperillo I'm done with the PR please could you have a look?

PR #2027, opened 2021-09-24
#1743 Implement Connect Exercise

kkyb123:
# pull request

Provide test, starter and reference implementation for the *connect* practice exercise

---

Reviewer Resources:


exercises/practice/connect/.docs/instructions.md
```diff
@@ -0,0 +1,31 @@
...
+Compute the result for a game of Hex / Polygon.
+
+The abstract boardgame known as
+[Hex](https://en.wikipedia.org/wiki/Hex_%28board_game%29) / Polygon /
+CON-TAC-TIX is quite simple in rules, though complex in practice. Two players
+place stones on a parallelogram with hexagonal fields. The player to connect his/her
+stones to the opposite side first wins. The four sides of the parallelogram are
+divided between the two players (i.e. one player gets assigned a side and the
+side directly opposite it and the other player gets assigned the two other
+sides).
```
jmrunkle, line 12:
```suggestion
The abstract boardgame known as [Hex](https://en.wikipedia.org/wiki/Hex_%28board_game%29) / Polygon / CON-TAC-TIX is quite simple in rules, though complex in practice.
Two players place stones on a parallelogram with hexagonal fields.
The player to connect his/her stones to the opposite side first wins.
The four sides of the parallelogram are divided between the two players (i.e. one player gets assigned a side and the side directly opposite it and the other player gets assigned the two other sides).
```

Prefer splitting on sentence breaks so that we reduce the splash radius for future changes.

exercises/practice/connect/.meta/src/reference/java/Connect.java
```diff
@@ -0,0 +1,270 @@
...
+ * For additional documentation, check
+ * <a href="https://en.wikipedia.org/wiki/Disjoint-set_data_structure">Wikipedia Disjoint-set Data Structure</a>
+ * or <a href="https://algs4.cs.princeton.edu/15uf">Section 1.5</a> of
+ * <i>Algorithms, 4th Edition, A nice book on DSA</i> by Robert Sedgewick and Kevin Wayne.
+ */
+class Connect {
+
+    //The disjoint-set of player O. Each element of this array represents a field and each field is connected to itself
+    // at initialization.
+    private int[] parentO;
```
jmrunkle, line 14: nit: This solution is fundamentally hard to read / verify because it is using a 1D representation of fundamentally 2D data. Is there any significant reason to not use a 2D data structure like an `int[][]`? Remember that mentors (and some students) will refer to this solution as an example of how to solve it. That being said, as long as it correctly solves the exercise we can use it. I just dislike having to keep so many arrays with the exact same size.

exercises/practice/connect/src/main/java/Connect.java
```diff
@@ -0,0 +1,16 @@
...
+    public Winner computeWinner(String[] board) {
+        throw new UnsupportedOperationException("Implement this function");
+    }
+
+    enum Winner {
+        PLAYER_X,
+        PLAYER_O,
+        NONE
+    }
+}
```
jmrunkle, line 16:
```suggestion
}

```

Add trailing newline

exercises/practice/connect/src/test/java/ConnectTest.java
```diff
@@ -0,0 +1,220 @@
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
```
jmrunkle, line 4: We have been transitioning tests over to AssertJ - mind switching this one over as well?

The assertion becomes: assertThat(got).isEqualTo(want)

exercises/practice/connect/src/test/java/ConnectTest.java
```diff
@@ -0,0 +1,220 @@
...
+            "   . . . . .",
+            "    . . . . ."
+        };
+        Connect cut = new Connect(board);
+
+        //WHEN
+        var winner = cut.computeWinner();
+
+        //THEN
+        assertEquals(Connect.Winner.NONE, winner);
```
jmrunkle, line 25:
```suggestion
        assertThat(winner).isEqualTo(Connect.Winner.NONE);
```

An example of using AssertJ

exercises/practice/connect/src/main/java/Connect.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Winner computeWinner(String[] board) {
+        throw new UnsupportedOperationException("Implement this function");
+    }
+
+    enum Winner {
+        PLAYER_X,
+        PLAYER_O,
+        NONE
+    }
```
jmrunkle, line 15: This enum should be in a different file that the online editor displays but does not allow edits for.

jmrunkle (changes requested):
Thanks for working on this!

kkyb123:
Thanks for the fedback @jmrunkle !

exercises/practice/connect/.meta/src/reference/java/Connect.java
kkyb123: Thanks for mentioning this. I also had doubts about the clarity of the solution.
As mentioned in the code's comments, I found it pertinent to use a **disjoint-set** data structure to solve this problem. The aim here was to introduce the ideas of the **Union Find** algorithms to solve *dynamic connectivity* problems. This might be beneficial to the students in solving this exercise and other similar exercises such as the *Percolation* problem. These algorithms and data structures are usually implemented and taught using a 1D array. That being said, it still doesn't solve our issue of clarity specific to this exercise. So my suggestion is this; What do you think of abstracting the disjoint-set data structure in a class called **Board**, that will model the HEX's board, then call operations on this board from the **Connect** class? I think this will show a clear implementation of the Union Find algorithm in a separate class (**Board**) and how it is used to provide a solution for **Connect**.

jmrunkle: Yes, that sounds like a significant improvement and a better abstraction.

kkyb123: Understood. Please one more question; taking into consideration the comment about the enum that should be in a file of its own, should I also add the **Board** class in a separate file? Also when adding new solution files, do we need to include their path in the **config.json**?

jmrunkle: Yes config.json needs to include all of the files. There's a different section for editable ones vs non-editable ones. Check the spec on the website documentation on exercises.

exercises/practice/connect/src/main/java/Connect.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Connect(String[] board) {
+        throw new UnsupportedOperationException("Implement this function");
+    }
+
+    public Winner computeWinner(String[] board) {
+        throw new UnsupportedOperationException("Implement this function");
+    }
+
+    enum Winner {
```
jmrunkle, line 11: I thought we were moving this enum to an isolated file and marking it as an editor only file (ie. the student should not be able to modify it in the UI).

jmrunkle (changes requested):

exercises/practice/connect/src/test/java/ConnectTest.java
```diff
@@ -1,7 +1,7 @@
 import org.junit.Ignore;
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat(winner).isEqualTo();
```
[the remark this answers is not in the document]
kkyb123: Yeah updating that

exercises/practice/connect/src/main/java/Connect.java
kkyb123: Oh thanks for the reminder

kkyb123:
@jmrunkle please could you have a look again?

PR #2036, opened 2021-10-02
Update introduction.md

myucesan:
Added comparison operator example, removed the reference to truthy values as it will only confuse new programmers

# pull request

---

Reviewer Resources:

### DIFF
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -7,7 +7,7 @@ There are two different types of numbers in Java:
 
 The two most common numeric types in Java are `int` and `double`. An `int` is a 32-bit integer and a `double` is a 64-bit floating-point number.
 
-Arithmetic is done using the standard arithmetic operators. Numbers can be compared using the standard numeric comparison operators (5 > 4 and 4 < 5) and the equality (`==`) and inequality (`!=`) operators.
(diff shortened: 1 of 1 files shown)


concepts/numbers/introduction.md
```diff
@@ -7,7 +7,7 @@ There are two different types of numbers in Java:
 
 The two most common numeric types in Java are `int` and `double`. An `int` is a 32-bit integer and a `double` is a 64-bit floating-point number.
 
-Arithmetic is done using the standard arithmetic operators. Numbers can be compared using the standard numeric comparison operators and the equality (`==`) and inequality (`!=`) operators.
+Arithmetic is done using the standard arithmetic operators. Numbers can be compared using the standard numeric comparison operators (5 > 4 and 4 < 5) and the equality (`==`) and inequality (`!=`) operators.
```
jmrunkle, line 10:
```suggestion
Arithmetic is done using the standard arithmetic operators. Numbers can be compared using the standard numeric comparison operators (eg. `5 > 4` and `4 <= 5`) and the equality (`==`) and inequality (`!=`) operators.
```

jmrunkle (changes requested):
Do we also need to make similar changes to the about.md file?

myucesan:
> Do we also need to make similar changes to the about.md file?

No, it already has examples.

jmrunkle (approved):

PR #2052, opened 2021-10-31
Converting tests to AssertJ (exercises starting with H, I, K)

thtroyer:
I saw someone else group their AssertJ conversions together and figured I'd start doing the same.

The tests have been converted from using JUnit's assertions to use AssertJ, cleaning up and inlining portions of the tests that make sense to do so.

I also updated the reference implementation for knapsack.  It needlessly specified ArrayList in its method signatures (instead of just List).  Switching to List let me clean up the tests a little more.

Exercises updated:
- hamming
- hangman
- hello-world
- hexadecimal
- house
- isbn-verifier
- isogram
- kindergarten-garden
- knapsack

---

Reviewer Resources:


mirkoperillo (approved):
thank you @thtroyer LGTM

PR #2053, opened 2021-11-02
Add a Reflection proxy utility class

leocck:
# Stubs with Reflection

While solving the basic concept called "Cook your Lasagna", I noticed the tests would not run until I had implemented all the methods > reason being, the tests were calling the methods from the class directly, causing a build error.

To make the code compile, we should create stubs for the methods the students will implement, but the Lasagna exercise is about how to define methods, and by defining the methods in stubs we spoil the exercise somewhat.

This PR adds a utility class called ReflectionProxy, which simplifies the use of Reflection API by encapsulating the target class. With it, it is easy to have proxied stubs **and** check if the methods were implemented correctly at the same time.

This utility class was just copied over from source code we use at the [ReDI School of Digital Integration](https://www.redi-school.org/berlin) with our students and might deserve some improvements. Here's [an example](https://github.com/RediJavaClassroomF21/09---oop-4---abstract-classes-and-interfaces-leocck) of the exercises we are using it.

---

Reviewer Resources:


exercises/concept/lasagna/src/test/java/utils/Lasagna.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public String getTargetClassName() {
+        return "Lasagna";
+    }
+
+    public int expectedMinutesInOven() {
+        try {
+            return invokeMethod("expectedMinutesInOven", new Class[]{});
+        } catch (Exception e) {
+            return 0;
```
leocck, line 14: Here you will probably want to throw a UnsupportedOperationException

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
+package utils;
+
+import java.lang.reflect.*;
+
+import static java.lang.Class.forName;
+
+public abstract class ReflectionProxy {
```
leocck, line 7: This class is long because features were added from time to time, in a collaborative but unorganised way. 
I made my best to document all the methods before bringing it to this repository, but I'm sure you will find a lot of things that could be improved.

iHiD:
This is a really interesting approach. Thanks. I'll let the Java maintainers review it etc, but just wanted to say thanks for the PR :)

ericbalawejder:
As you implemented the Lasagna class for the new tests, the tests pass that check the method stub. Should we create stubs that cause all the tests to fail? Such as omitting the access modifier and changing the return type to void? This would be more consistent with the TDD approach for building each problem.

leocck:
> I like the idea of testing for the correct method signature without the build failure. This is more of an issue when using the browser editor rather than local tools which was introduced in v3.
> 
> As you implemented the Lasagna class for the new tests, the tests pass that check the method stub. Should we create stubs that cause all the tests to fail? Such as omitting the access modifier and changing the return type to void? This would be more consistent with the TDD approach for building each problem.

I think the idea of this exercise is to provide them an empty class, so that they implement the whole method (signature and logic) by themselves. If we provide them wrong code, then I think the exercise description will have to be changed for "fix the provided code" or something like that. WDYT?

ericbalawejder:
> Just to see if I understood you correctly: you mean the Lasagna class they will have to implement (the template), right? The one from the util package is just a proxy (if the methods do not exist in the target class, or have the wrong signature, the tests will still fail).

The idea is to provide a template for the student as such?
```
public class Lasagna extends ReflectionProxy {

    // Implement methods here

    @Override
    public String getTargetClassName() {
        return null;
    }

}
```

leocck:
> The idea is to provide a template for the student as such?

No, that's not the place. The students will get an empty class, from the src folder ([this file](https://github.com/exercism/java/blob/main/exercises/concept/lasagna/src/main/java/Lasagna.java)).

The one you are referring to is inside the test folder and is used by the tests as a Proxy (it will call the one that the students must implement and check if they did it correctly) 😉

ericbalawejder:
Yes, I had the path mixed up as you pointed out. I understand it as unit test --> proxy --> collaborator where the collaborator is the empty file `src/main/java/Lasagna`?

I like the idea and implementation.

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
ericbalawejder: I've seen all the implementations of Reflection Proxies throw UnsupportedOperationException as you mentioned.

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * An instance of the target class (if found)
+     */
+    private Object target;
+
```
ericbalawejder, line 13: `target` can be final for this implementation.

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    /**
+     * The default constructor, for when you have already an instance of the target class
+     * @param target An instance of the target class
+     */
+    public ReflectionProxy(Object target) {
+        this.target = target;
+    }
+
```
ericbalawejder, line 29: It doesn't technically matter for instantiation but would it be more valuable to change the constructors to `protected` to control scope?

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
+package utils;
+
+import java.lang.reflect.*;
+
```
ericbalawejder, line 4: import java.lang.reflect.Constructor;
import java.lang.reflect.Field;
import java.lang.reflect.Method;
import java.lang.reflect.Modifier;
import java.lang.reflect.ParameterizedType;

It helps future maintainers when we avoid wildcard imports.

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
...
+     * @param parameterTypes The list of desired parameter types
+     * @return True if the constructor exists, false otherwise
+     */
+    public boolean hasConstructor(Class<?>... parameterTypes) {
+        Class<?> targetClass = getTargetClass();
+        if (targetClass == null) {
+            return false;
+        }
+        try {
+            Constructor c = targetClass.getDeclaredConstructor(parameterTypes);
```
ericbalawejder, line 233: Constructor<?> c = targetClass.getDeclaredConstructor(parameterTypes);

Wildcard to avoid raw types?

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
...
+     * Creates an instance of the target class
+     * @param args The list of constructor parameters
+     * @return An instance of the target class, if found, or null otherwise
+     */
+    private Object instantiateTarget(Object... args) {
+        Class<?> targetClass = getTargetClass();
+        if (targetClass == null) {
+            return null;
+        }
+        Constructor[] constructors = getAllConstructors();
```
ericbalawejder, line 385: Constructor<?>[] constructors = getAllConstructors();

Wild card type?

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
```diff
@@ -0,0 +1,477 @@
...
+            return false;
+        }
+        try {
+            Method m = targetClass.getDeclaredMethod(name, parameterTypes);
+            return Modifier.isProtected(m.getModifiers());
+        } catch (NoSuchMethodException e) {
+            return false;
+        }
+    }
+}
```
ericbalawejder, line 477: Not all the methods are used but it seems like a good template for future use in other exercises.

exercises/concept/lasagna/src/test/java/LasagnaTest.java
```diff
@@ -1,42 +1,90 @@
...
+                .isTrue();
+    }
+
     @Test
     public void expected_minutes_in_oven() {
         assertThat(new Lasagna().expectedMinutesInOven()).isEqualTo(40);
     }
 
-    @Ignore("Remove to run test")
+    @Test
```
ericbalawejder, line 27: I believe the test runner actually removes the `@Ignore` annotation when it runs the tests on the Exercism server. The `@Ignore` annotation is normally placed on every test except the first one for local development purposes, i.e. `exercism download --exercise=lasagna --track=java` with the command line tool.

leocck: Interesting, I didn't know that, but it makes a lot of sense now

leocck:
> I like the idea and implementation.

Yay! Thanks for your review, I fixed all the issues you pointed out.

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: Yeah, I thought we could leave them for reference.

mirkoperillo: I suggest to comment the unused methods with a explicit message like `Unused at moment` or something like that.
It can improve the time to understand on the code for a maintainer who unknows it

ErikSchierboom:
Great idea! I'm planning to do something similar for the C# track.

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: I agree. What do you say about grouping the unused methods with a _**region**_ section? (it works for almost all editors and is still "understandable" in case it is not)

image

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
ericbalawejder: Were there two constructors?

leocck: The constructor that receives the target instance is not being used. Moved to the "unused" region 👍

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
leocck: done

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: done

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: done

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: done

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: done

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
ericbalawejder: Ok, the main constructor is still marked `public`.

leocck: I think that's ok, since the class is abstract, right?

ericbalawejder: > It doesn't technically matter for instantiation but would it be more valuable to change the constructors to protected to control scope?

It's more of a style thing since this may be used as a template in further exercises.

leocck: ok, done!

ericbalawejder:
@mirkoperillo The build action: `Compile check style with gradle` fails and has 65K+ lines of output. I assume it's running all of the exercises? Is this normal or intended behavior?

ericbalawejder (approved):

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
ErikSchierboom: It might be a good idea to create a Gradle package that contains the reflection proxy, as it can then be re-used across exercises. We've done a similar thing with C#: https://github.com/exercism/dotnet-tests

leocck: @ErikSchierboom : that's exactly what I wanted to do, but I don't know how to setup an internal Gradle package 😅 . If you point me in the right direction I will do it 🙏

ErikSchierboom: We have to ask @iHiD to create a repository. Ideas on a good name?

leocck: Oh, ok, I thought we could move this class to an upper folder and be able to share it with all exercises.

Naming is the most difficult part 😅 . Are there any other utility libraries to refer to, or conventions already?

If not, I would suggest something like `exercism:test-utils`

ErikSchierboom: I don't think we have any conventions. We don't yet support shared files between exercises and it might be nice to hide those files from the student anyway.

mirkoperillo:
@ericbalawejder you have right. Currently when the CI workflow starts the `gradle check` command is executed on all the exercises. We need to improve it

https://github.com/exercism/java/blob/acfdd2b654a1e2b8bb497c3af25a9e2eff1044e0/.github/workflows/gradle.yml#L23

exercises/concept/lasagna/src/test/java/utils/ReflectionProxy.java
leocck: So, should we merge this PR and create another ticket for the new repository? (are you using Jira or other issue tracking?)

mirkoperillo (changes requested):
@leocck sorry for the late feedback, your PR is alright: I run the tests against it and work fine.
Currently the code breaks some style constraints.
Can I ask to run
```
gradle clean check
```
on your PR code and fix them ? Thank you a lot

leocck:
@mirkoperillo thanks for looking into that, i always forget about the gradle check 🤦🏼

Fixed now

mirkoperillo (approved):

mirkoperillo:
@leocck thank you, now PR is ready to be merged

PR #2061, opened 2021-12-03
Improve secret-handshake assertions with AssertJ

### LINKED ISSUE #2060: Fix assertions for secret-handshake
The signature of the method is defined as: 
```
List<Signal> calculateHandshake(int number)
```
but we're not given any instructions on how the `Signal` type should look like to pass the tests. My tests keep failing because of incompatible types, for example 
```
expected: java.util.Collections$SingletonList<[JUMP]> but was: java.util.ArrayList<[JUMP]>
```
or 
```
expected: java.util.Arrays$ArrayList<[WINK, DOUBLE_BLINK]> but was: java.util.ArrayList<[WINK, DOUBLE_BLINK]>
```
I'm not sure if this is part of the exercise I should figure out? Thank you for guidance.

jmrunkle:
# pull request

Fixes #2060 

Basically, the current assertions are horribly broken, but this fixes that.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/secret-handshake/src/test/java/HandshakeCalculatorTest.java
+++ b/exercises/practice/secret-handshake/src/test/java/HandshakeCalculatorTest.java
@@ -1,130 +1,100 @@
 import org.junit.Ignore;
-import org.junit.Before;
 import org.junit.Test;
 
-import static java.util.Arrays.asList;
-import static java.util.Collections.emptyList;
-import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class HandshakeCalculatorTest {
 
-    private HandshakeCalculator handshakeCalculator;
-
-    @Before
-    public void setUp() {
-        handshakeCalculator = new HandshakeCalculator();
-    }
+    private final HandshakeCalculator handshakeCalculator = new HandshakeCalculator();
 
     @Test
     public void testThatInput1YieldsAWink() {
-        assertEquals(
-                singletonList(Signal.WINK),
-                handshakeCalculator.calculateHandshake(1));
+        assertThat(handshakeCalculator.calculateHandshake(1)).containsExactly(Signal.WINK);
     }
 
     @Ignore("Remove to run test")
     @Test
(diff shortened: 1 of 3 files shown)


ericbalawejder:
I see. The tests were checking for a collection type as well as the contents? `Collections.singletonList()`, `.emptyList()` as well as the `Signals`?

exercises/practice/secret-handshake/src/test/java/HandshakeCalculatorTest.java
```diff
@@ -1,130 +1,100 @@
...
 
 public class HandshakeCalculatorTest {
 
-    private HandshakeCalculator handshakeCalculator;
-
-    @Before
-    public void setUp() {
-        handshakeCalculator = new HandshakeCalculator();
-    }
+    private final HandshakeCalculator handshakeCalculator = new HandshakeCalculator();
```
ericbalawejder, line 9: Since the object has no state, it doesn't matter if we remove the initialization from the `@Before` block?

jmrunkle: Correct, and even if it did have state, each `@Test` is actually run independently so you would only possibly get an issue if they were running in parallel. Frankly, if a student's solution has any mutable state, that is a bad implementation of this exercise. It may eventually be worth adding a test that does multiple calculations in a row and making sure it always has the correct results across multiple runs (eg. not re-using a mutable list for each calculation).

jmrunkle:
> The tests were checking for a collection type as well as the contents? `Collections.singletonList()`, `.emptyList()` as well as the `Signals`? I'm having difficulty reproducing this error. I'm using a stream collector `.collect(Collectors.toUnmodifiableList());` with 0 - 4 `Signals`.

I believe that it considers equality based on the `equals` method, so as long as your result matches returns true for `equals` then it should pass the test (but I am pretty sure that is too restrictive). The user who submitted the issue was using an ArrayList, so you could try something like this:

```java
.collect(Collectors.toCollection(ArrayList::new))
```

jmrunkle:
Anything else @ericbalawejder?

ericbalawejder (approved):
The assertions are more clear but I still can't reproduce the error.

PR #2064, opened 2021-12-06
Remove duplicate files.exemplar entry

jmrunkle:
This now breaks configlet lint since it matches files.example.

# pull request

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -27,9 +27,6 @@
     ],
     "example": [
       ".meta/src/reference/java/%{pascal_slug}.java"
-    ],
-    "exemplar": [
-      ".meta/src/reference/java/%{pascal_slug}.java"
     ]
   },
   "exercises": {


ericbalawejder:
I honestly forget the reason behind the two different spellings? Does @ErikSchierboom know?

kotp:
I am sure he knows.  But the example is an example, not necessarily exemplar.

```
From The Collaborative International Dictionary of English v.0.48 [gcide]:

  Exemplar \Ex*em"plar\, n. [L. exemplar, exemplum: cf. F.
     exemplaire. See {Example}, and cf. {Examper}, {Sampler}.]
     1. A model, original, or pattern, to be copied or imitated; a
        specimen; sometimes; an ideal model or type, as that which
        an artist conceives.
        [1913 Webster]
  
              Such grand exemplar as make their own abilities the
              sole measure of what is fit or unfit. --South.
        [1913 Webster]
```

ErikSchierboom (approved):
We need to fix this upstream at some point. Sorry about this

PR #2065, opened 2021-12-08
robot-name: added new test

### LINKED ISSUE #2056: Add test to robot name exercise - ensure robot names are unique
Hello everyone, 

In the [robot name exercise](https://github.com/exercism/java/blob/main/exercises/practice/robot-name), the instructions require unique names:

> Using random names means a risk of collisions. Your solution must ensure that every existing robot has a unique name.

There is currently a `differentRobotsHaveDifferentNames()` test, but that doesn't ensure collisions won't happen.

There is no [canonical data for the tests](https://github.com/exercism/problem-specifications/tree/main/exercises/robot-name) of this exercise, but other tracks implement a specific test for unique names, like the [javascript track](https://github.com/exercism/javascript/blob/main/exercises/practice/robot-name/robot-name.spec.js).

I'm completely new in here and don't know the impact of changing the tests of an existing exercise. If that is ok, I would like to add this test to the exercise.

cdpaiva:
Closes #2056 

Added a test to ensure there will be no collisions between the generated names. 

I did a small research to find the minimum sample size to ensure no collisions, extending the principle of the [Birthday Problem](https://en.wikipedia.org/wiki/Birthday_problem). The ideal size I found, i.e. when the probability of a collision tends to 100%, is 5000 samples.

I also updated the reference solution, so it would pass this test.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
+++ b/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
@@ -17,9 +17,9 @@ public class Robot {
     private void assignNewName() {
         String newName = String.format("%s%d", prefix(), suffix());
         if (nameLog.add(newName)) {
-        	this.name = newName;
+            this.name = newName;
         } else {
-        	assignNewName();
+            assignNewName();
         }
     }
 
--- a/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
+++ b/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
@@ -16,7 +16,7 @@ public class Robot {
 
     private void assignNewName() {
         String newName = String.format("%s%d", prefix(), suffix());
-        if(nameLog.add(newName)) {
+        if (nameLog.add(newName)) {
         	this.name = newName;
         } else {
         	assignNewName();
--- a/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
+++ b/exercises/practice/robot-name/.meta/src/reference/java/Robot.java
@@ -1,10 +1,13 @@
+import java.util.HashSet;
 import java.util.Random;
+import java.util.Set;
 
 public class Robot {
 
     private static final String ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
     private String name;
     private final Random random;
+    private static Set<String> nameLog = new HashSet<>();
 
     public Robot() {
         random = new Random();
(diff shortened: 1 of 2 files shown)


exercises/practice/robot-name/src/test/java/RobotTest.java
```diff
@@ -34,6 +37,17 @@ public void resetName() {
         assertThat(name).isNotEqualTo(name2);
         assertIsValidName(name2);
     }
+    
+    @Ignore("Remove to run test")
+    @Test
+    public void robotNamesAreUnique() {
+        Set<String> robotNames = new HashSet<>();
+        Integer sampleSize = 5000;
```
jmrunkle, line 45: Super small nitpicky thing: use int rather than Integer here. Why? Because Integer makes it seem like you care about using the boxed value (eg. you might need null or something), but you do not in this case.

```suggestion
        int sampleSize = 5000;
```

exercises/practice/robot-name/.meta/src/reference/java/Robot.java
```diff
@@ -1,18 +1,26 @@
...
         assignNewName();
     }
 
     private void assignNewName() {
-        name = String.format("%s%d", prefix(), suffix());
+        String newName = String.format("%s%d", prefix(), suffix());
+        if (nameLog.add(newName)) {
+            this.name = newName;
+        } else {
+            assignNewName();
```
jmrunkle, line 22: This structure actually has really bad recursion after you have generated most of the names and will cause a stack overflow eventually. I recommend making this a for loop instead.

cdpaiva: Thanks for the suggestion! I'll change the code to a loop.

exercises/practice/robot-name/src/test/java/RobotTest.java
```diff
@@ -34,6 +37,17 @@ public void resetName() {
...
+    
+    @Ignore("Remove to run test")
+    @Test
+    public void robotNamesAreUnique() {
+        Set<String> robotNames = new HashSet<>();
+        int sampleSize = 5000;
+        for (int i = 0; i < sampleSize; i++) {
+            robotNames.add(new Robot().getName());
+        }
+        assertThat(robotNames.size()).isEqualTo(sampleSize);
```
jmrunkle, line 49: Missed this the first time, but I'm pretty confident that this works and does what you want more clearly / with a better assertion.

```suggestion
        assertThat(robotNames).hasSize(sampleSize);
```

cdpaiva: That is better for sure, thanks

jmrunkle (approved):

PR #2080, opened 2021-12-27
Added prerequisites up to Rotational Cipher

cdpaiva:
Issue #1867 

I started to add the prerequisites for the exercises. I will follow the order the exercises appear in the `config.json` file.

I also updated the prerequisites of the connect exercise, to match the concepts in the `concepts` array.

For new concepts, such as `enums` and `exceptions`, I'll submit new issues to create the corresponding concept exercise first. Once it is implemented, we can update the prerequisites.

I only changed the first 10 exercises. Please be extra critic in the feedback, this will help me make a better job in the next exercises yet to be updated. 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -244,8 +244,7 @@
           "conditionals-if"
         ],
         "prerequisites": [
-          "basics",
-          "conditionals-if"
+          "basics"
         ],
         "difficulty": 1
       },
@@ -254,14 +253,11 @@
         "name": "Hamming",
         "uuid": "ce899ca6-9cc7-47ba-b76f-1bbcf2630b76",
         "practices": [
-          "strings",
           "for-loops"
         ],
         "prerequisites": [
-          "basics",
           "strings",
-          "chars",
-          "for-loops"
+          "chars"
         ],
         "difficulty": 3
       },
@@ -273,8 +269,7 @@
           "numbers"
         ],
         "prerequisites": [
-          "basics",
-          "numbers"
+          "basics"
         ],
         "difficulty": 3
       },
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -229,137 +229,152 @@
...
-        "topics": [
-          "conditionals-if",
-          "strings"
-        ]
+        "practices": [
+          "strings",
+          "conditionals-if"
+        ],
+        "prerequisites": [
+          "basics",
```
jmrunkle, line 247: Do not include basics if you include anything else (because any other concept will have basics as a prerequisite).

config.json
```diff
@@ -229,137 +229,152 @@
...
-          "loops",
-          "strings"
-        ]
+        "practices": [
+          "strings",
+          "for-loops"
+        ],
+        "prerequisites": [
+          "basics",
+          "strings",
```
jmrunkle, line 262: Concepts should not be in both practices and prerequisites

jmrunkle (changes requested):
Thanks, just a few minor changes needed

config.json
cdpaiva: I was in doubt if I should include basics or not, but ended up inserting it. Thanks for the suggestion, I've removed it from all exercises that have other prerequisites. It really makes more sense, otherwise all the exercises of the track would end up with basics as a prerequisite.

config.json
cdpaiva: In my interpretation, users could need the basics of a concept in order to solve an exercise and could end up practicing that same concept once the exercise is done.

Following your recommendation, I've changed `prerequisites` to include only the strict necessary concepts. The ones that can be practiced went to `practices`.

jmrunkle (approved):

PR #2085, opened 2022-01-19
Update LasagnaTest.java

laurahannah:
# pull request

Corrected the last two tests

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/lasagna/src/test/java/utils/Lasagna.java
+++ b/exercises/concept/lasagna/src/test/java/utils/Lasagna.java
@@ -31,11 +31,11 @@ public class Lasagna extends ReflectionProxy {
         }
     }
 
-    public int totalTimeInMinutes(int amountLayers, int actualMinutes) {
+    public int totalTimeInMinutes(int amountLayers) {
         try {
-            return invokeMethod("totalTimeInMinutes", new Class[]{int.class, int.class}, amountLayers, actualMinutes);
+            return invokeMethod("totalTimeInMinutes", new Class[]{int.class}, amountLayers);
         } catch (Exception e) {
-            throw new UnsupportedOperationException("Please implement the totalTimeInMinutes(int, int) method");
+            throw new UnsupportedOperationException("Please implement the totalTimeInMinutes(int) method");
         }
     }
 }
--- a/exercises/build.gradle
+++ b/exercises/build.gradle
@@ -1,87 +1,25 @@
-// This root Gradle project is used for:
-//
-// - Checking that every test suite compiles against its corresponding reference implementation;
-// - Checking that every reference implementation passes its corresponding test suite;
-// - Checking that every provided starter implementation compiles.
-//
-// This file is never delivered to Exercism users.
+apply plugin: "java"
+apply plugin: "eclipse"
+apply plugin: "idea"
 
-def generatedTestSourceDir = "build/gen/test/java"
+// set default encoding to UTF-8
+compileJava.options.encoding = "UTF-8"
+compileTestJava.options.encoding = "UTF-8"
 
-subprojects {
-
-  apply plugin: "checkstyle"
-
-  // Add a task that copies test source files into a build directory. All @Ignore annotations
-  // are removed during this copying, so that when we run the copied tests, none are skipped and
-  // all should execute.
-  //
-  // Note that journey-test.sh also strips @Ignore annotations using sed. The
-  // stripping implementations here and in journey-test.sh should be kept
-  // consistent.
-  task copyTestsFilteringIgnores(type: Copy) {
-    from "src/test/java"
-    into generatedTestSourceDir
-    filter { line ->
-      line.contains("@Ignore") ? null : line
-    }
-  }
-
-  afterEvaluate { project ->
-    // skip project named 'concept' or 'practice'
-    // they are only folders containing exercises
-    if(project.name == 'concept' || project.name == 'practice')
-      return;
-
-    sourceSets {
-      // Set the directory containing the reference solution as the default source set. Default
-      // compile tasks will now run against this source set.
-      main {
-        java.srcDirs = [".meta/src/reference/java"]
-      }
-
-      // Define a custom source set named "starterSource" that points to the starter implementations
-      // delivered to users. We can then use the generated "compileStarterSourceJava" task to verify
-      // that the starter source compiles as-is.
-      starterSource {
-        java.srcDirs = ["src/main/java"]
-      }
-
-      // Set the directory containing the @Ignore-stripped tests as the default test source set.
-      // Default test tasks will now run against this source set.
-      test {
-        java.srcDirs = [generatedTestSourceDir]
-      }
-
-      // Log the source paths associated with each source set to verify they are what we expect.
-      logCompileTaskSourcePath(project, "compileJava") // Corresponds to the "main" source set.
-      logCompileTaskSourcePath(project, "compileStarterSourceJava")
-      logCompileTaskSourcePath(project, "compileTestJava")
-
-    }
-
-    // configuration of the linter
-    checkstyle {
-      toolVersion '7.8.1'
-      configFile file("checkstyle.xml")
-      sourceSets = [project.sourceSets.main, project.sourceSets.test]
-    }
-
-    checkstyleTest {
-      source = "src/test/java"
-    }
-
-    // When running the standard test task, make sure we prepopulate the test source set with the
-    // @Ignore-stripped tests.
-    test.dependsOn(copyTestsFilteringIgnores)
-  }
-
-
-
+repositories {
+  mavenCentral()
 }
 
-def logCompileTaskSourcePath(Project project, String taskName) {
-  project[taskName].doFirst { compileTask ->
-    println "  (source = ${compileTask.source.asPath})"
+dependencies {
+  testImplementation 'org.junit.jupiter:junit-jupiter:5.8.2'
+  testImplementation "org.assertj:assertj-core:3.15.0"
+}
+
+test {
+  useJUnitPlatform()
+  testLogging {
+    exceptionFormat = 'short'
+    showStandardStreams = true
+    events = ["passed", "failed", "skipped"]
   }
 }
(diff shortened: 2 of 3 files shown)


ericbalawejder:
Based on the instructions, the original values seem correct. Can you elaborate on how you arrived at these new test values?

laurahannah:
Sure.

The example in the README. 
Time in oven is a fixed 40 minutes.
3 layers leads to 2*3 = 6 minutes preparation time.
Preparation is 40 minutes plus 6.
Since the item has already been in the oven 20 minutes,
The total time left is 40 + 2*3 - 20 = 26

    @Test
    public void total_time_in_minutes_for_three_layers() {
        assertThat(new Lasagna().totalTimeInMinutes(3, 20)).isEqualTo(26);
    }

First test is 1 layer and already in the oven for 30 minutes
Total time left is 40 + 1*2 - 30 = 12
    @Test
    public void total_time_in_minutes_for_one_layer() {
        assertThat(new Lasagna().totalTimeInMinutes(1, 30)).isEqualTo(12);
    }

Second test is 
4 layers having been in the oven 8 minutes already
Total time left is 40 + 2*4 - 8 = 40
    @Test
    public void total_time_in_minutes_for_multiple_layers() {
        assertThat(new Lasagna().totalTimeInMinutes(4, 8)).isEqualTo(40);
    }

laurahannah:
> Based on the instructions, the original values seem correct. Can you elaborate on how you arrived at these new test values?

Sure.

The example in the README.
Time in oven is a fixed 40 minutes.
3 layers leads to 2 times 3 = 6 minutes preparation time.
Preparation is 40 minutes plus 6.
Since the item has already been in the oven 20 minutes,
The total time left is 40 + (2 times 3) - 20 = 26

@Test
public void total_time_in_minutes_for_three_layers() {
    assertThat(new Lasagna().totalTimeInMinutes(3, 20)).isEqualTo(26);
}
First test is 1 layer and already in the oven for 30 minutes
Total time left is 40 + (1 times 2) - 30 = 12
@test
public void total_time_in_minutes_for_one_layer() {
assertThat(new Lasagna().totalTimeInMinutes(1, 30)).isEqualTo(12);
}

Second test is
4 layers having been in the oven 8 minutes already
Total time left is 40 + (2 times 4) - 8 = 40
@test
public void total_time_in_minutes_for_multiple_layers() {
assertThat(new Lasagna().totalTimeInMinutes(4, 8)).isEqualTo(40);
}

AlbusPortucalis:
I believe you got the idea of the `totalTimeInMinutes()` method wrong. 

The objective of this test is to return the total time (in minutes) spent preparing the lasagna until now (`The function should return how many minutes in total you've worked on cooking the lasagna`).

> The example in the README.
Time in oven is a fixed 40 minutes.
3 layers leads to 2 times 3 = 6 minutes preparation time.
Preparation is 40 minutes plus 6.
Since the item has already been in the oven 20 minutes,
The total time left is 40 + (2 times 3) - 20 = 26

Should be:
3 layers leads to 2 times 3 = 6 minutes preparation time.
Since the item has already been in the oven for 20 minutes,
The total time spent cooking is (2 times 3) + 20 = 26

The tests should be:
``` 
    @Test
    public void total_time_in_minutes_for_one_layer() {
        assertThat(new Lasagna().totalTimeInMinutes(1, 30)).isEqualTo(32);
    }

    @Test
    public void total_time_in_minutes_for_three_layers() {
        assertThat(new Lasagna().totalTimeInMinutes(3, 20)).isEqualTo(26);
    }

    @Test
    public void total_time_in_minutes_for_multiple_layers() {
        assertThat(new Lasagna().totalTimeInMinutes(4, 8)).isEqualTo(16);
    }
}
```

AlbusPortucalis:
Closes https://github.com/exercism/exercism/issues/6199

laurahannah:
> Closes [exercism/exercism#6199](https://github.com/exercism/exercism/issues/6199)

Have updated the repo to reflect the confusing superfluous parameter and updating the description. This is a base exercise the blocks the track. I would think we would want it to be very concise. Unless the point is to challenge the learner to understand that there is a superfluous parameter on the method.

exercises/concept/lasagna/src/main/java/Lasagna.java
```diff
@@ -1,9 +1,18 @@
...
 
-    // TODO: define the 'preparationTimeInMinutes()' method
+    public static int preparationTimeInMinutes(int layers) {
+        return layers * 2;
+    }
 
-    // TODO: define the 'totalTimeInMinutes()' method
+    public static int totalTimeInMinutes(int amountLayers) {
+        return preparationTimeInMinutes(amountLayers) + expectedMinutesInOven();
+    }
```
ErikSchierboom, line 17: The goal of this exercise is to have students learn how to define methods. Putting the actual implementation in them would defeat that purpose.

ErikSchierboom (changes requested):

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
```diff
@@ -31,11 +31,11 @@ public int preparationTimeInMinutes(int amountLayers) {
         }
     }
 
-    public int totalTimeInMinutes(int amountLayers, int actualMinutes) {
+    public int totalTimeInMinutes(int amountLayers) {
```
ErikSchierboom, line 34: This changes the spec from two parameters to one, whereas we _want_ students to use two parameters.

ErikSchierboom (changes requested):
@laurahannah I appreciate your desire to help improve our exercises, but this PR seems to go against what we try to achieve with this exercise (having students learn how to define methods and call them).

I'm sorry that the time you spent on this PR will be for naught. Might I suggest opening an issue first for potential improvements/changes? That way, we can first agree upon what changes we want to see. Thanks!

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
laurahannah: I guess I am still not understanding. The goal of this exercise is to have the student use two parameters even though one is never used. I did open a defect, but it was closed as working as expected. But I don't believe that is actually the case since we were working on this one as a group and all of us had the same thought. And we are communicating and this is a good thing. This exercise is blocking our progress in the Java track. And we don't want to submit a solution that is bad practice, because then that looks bad on us if anybody reviews our solution. Please advise

laurahannah:
I have requested mentoring since it is completely unclear to me how to past this exercise and continue my learning track.

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
iHiD: @laurahannah Take a look at some other solutions: https://exercism.org/tracks/java/exercises/lasagna/solutions

For example:
```java
public class Lasagna {
    private final int COOKING_TIME_IN_MINS = 40; // cooking time is 40 minutes
    private final int MINS_PREP_PER_LAYER = 2; // each layer takes 2 minutes to cook
    
    // Returns expected length of 40 minutes
    public int expectedMinutesInOven(){
        return COOKING_TIME_IN_MINS;
    }
    // Returns minutes remaining in oven, based on expected minutes
    public int remainingMinutesInOven(int minutesPassed){
        return this.expectedMinutesInOven() - minutesPassed;
    }
    // Returns number of minutes you have spent preparing the lasagna
    public int preparationTimeInMinutes(int numLayers){
        return MINS_PREP_PER_LAYER * numLayers;
    }
    // Returns the total time you have been making the lasagna
    public int totalTimeInMinutes(int numLayers, int minutesPassed){
        return this.preparationTimeInMinutes(numLayers) + minutesPassed;
    }
}
```

You'll see both params are used when the exercise is solved correctly.

iHiD: Can the instructions be made clearer somehow?

laurahannah: maybe explain what the second parameter for the totalTimeInMinutes method is. Also, I have a virtually identical solution to one that was accepted 5 months ago that does not pass

iHiD: The instructions say:
>>> Define the totalTimeInMinutes() method that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven. The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.

How could we improve that to make it more clear?

ErikSchierboom:
> I have requested mentoring since it is completely unclear to me how to past this exercise and continue my learning track.

I think this is the way to go forward. Good luck.

I'm closing this PR.

exercises/concept/lasagna/src/test/java/utils/Lasagna.java
ynfle: I have to agree with the OP.
In my opinion, the method/function name should be `totalElapsedTimeInMinutes`.
When doing this exercise the first couple of times (in different languages) I was confused by this.
In my mind, the method would ask "how many minutes it **should** be in the oven" as apposed to "how many minutes **it has been** in the oven."
I was able to infer the correct behaviour from the tests, but the naming is still confusing

laurahannah: I agree 100%. Plus the example in the readme, when added to the unit tests, won't pass. Like, even if you are not talking about method names and such. Add the unit test from the readme, then you have this to solve:
```totalTimeInMinutes(1, 30) = 42
totalTimeIntMintues(3, 20) = 26
totalTimeInMintues(4, 8) = 48```

When in doubt, I went with the README. So, please update the README, which was the first request that I had.

Comment (ynfle) [reply exercises/concept/lasagna/src/test/java/utils/Lasagna.java:34]: > ```
> totalTimeIntMintues(3, 20) = 26
> totalTimeInMintues(4, 8) = 48
> ```

As far as I can tell, the second example isn't in the readme or tests.
The tests have the following test
```java
totalTimeInMinutes(4, 8) == 16
```
which makes total sense given the intended purpose of the method.

Also, the remake's example is 
```java
totalTimeIntMintues(3, 20) == 26
```
which also makes sense given the intended purpose of the method.

PR #2087, opened 2022-02-03
Add useful charAt function.

kwikiel:
For a very common pattern - iterating over string and doing something on each character one has to know a way to get specific character from the string. String to Char conversion. In syllabus char section has String requirement so the obvious logical connection would be to explain how to get one from another.

---

Reviewer Resources:

### DIFF
--- a/concepts/chars/introduction.md
+++ b/concepts/chars/introduction.md
@@ -16,4 +16,4 @@ This object has methods that allow a string to be constructed
 character by character and manipulated. At the end of the process
 `toString()` can be called on it to output a complete string.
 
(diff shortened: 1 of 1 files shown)


AlbusPortucalis (changes requested):
Is a good improvement, can you just use markdown in the method to keep the same configuration as the previous text? e.g: 'method'

AlbusPortucalis (approved):
This has now to be verified by a higher rank and they are with little time. But from my perspective is good. Thanks.

ericbalawejder (approved):
It's valuable to provide language specific context for basic operations, as you have shown, especially in concept exercises. Thank you for your contribution.

PR #2090, opened 2022-02-14
Miss matches of points in 4 test cases

ji10raj:
In Test Case 7 :
Co-ordinations were given ( -9, 9), so according to game rules, it should be 1 point and not 0 point.

In Test Case 8 :
Co-ordinations were given ( 7.1, -7.1), so according to game rules, it should be 1 point and not 0 point.

In Test Case 9 :
Co-ordinations were given ( -3.6 , -3.6), so according to game rules, it should be 5 points and not 1 point.

In Test Case 7 :
Co-ordinations were given ( 0.8, -0.8), so according to game rules, it should be 10 points and not 5 point.

# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/darts/src/test/java/DartsTest.java
+++ b/exercises/practice/darts/src/test/java/DartsTest.java
@@ -8,7 +8,7 @@ public class DartsTest {
     @Test
     public void missedTarget() {
         Darts darts = new Darts(-9, 9);
-        assertEquals(0, darts.score());
+        assertEquals(1, darts.score());
     }
 
     @Ignore("Remove to run test")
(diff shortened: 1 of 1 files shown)


ericbalawejder:
The circles on the target are drawn with a radius. The coordinate system for the dart is represented as `x` and `y` on a cartesian coordinate system. The radius for the circle of the coordinates provided is the hypotenuse of the coordinates.
```java
Darts(double x, double y) {
        this.dart = Math.hypot(x, y);
}
``
Recall that the hypotenuse of an isosceles triangle (all the test cases you mentioned) that is a 45, 45, 90 degree triangle is:

![45-45-90](https://user-images.githubusercontent.com/4513001/153930065-e03bbb62-5044-49d0-a87f-15684d5ec564.png)

PR #2093, opened 2022-02-22
`need-for-speed`: Change method `canFinishTrack` to `tryFinishTrack`

ynfle:
# Pull Request

This name change clarifies that the method is supposed to try and drive
the car around the track to see if it can finish the track before it
runs out of battery as opposed to exposing the state of the car and
directly calculating the remaining distance that that car can travel.

This can sometimes be confusing to those completing the exercise as to
whether to expose the state of the car or try to drive the car around
the track and see if it can make it. This clarifies that the latter
should be implemented

Co-authored-by: Erik Schierboom <erik_schierboom@hotmail.com>

Reviewer Resources:

### DIFF
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -80,6 +80,11 @@ var car = new NeedForSpeed(speed, batteryDrain);
 int distance = 100;
 var race = new RaceTrack(distance);
 
+car.distanceDriven()
+// => 0
+
 race.tryFinishTrack(car);
 // => true
-```
+
+car.distanceDriven()
+// => 100
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -70,7 +70,7 @@ car.distanceDriven();
 
 ## 6. Check if a remote control car can finish a race
 
-To finish a race, a car has to be able to drive the race's distance. This means not draining its battery before having crossed the finish line. Implement the `RaceTrack.carCanFinish()` method that takes a `NeedForSpeed` instance as its parameter and returns `true` if the car can finish the race; otherwise, return `false`:
+To finish a race, a car has to be able to drive the race's distance. This means not draining its battery before having crossed the finish line. Implement the `RaceTrack.tryFinishTrack()` method that takes a `NeedForSpeed` instance as its parameter and returns `true` if the car can finish the race; otherwise, return `false`:
 
 ```java
 int speed = 5;
@@ -80,6 +80,6 @@ var car = new NeedForSpeed(speed, batteryDrain);
 int distance = 100;
 var race = new RaceTrack(distance);
 
-race.carCanFinish(car);
+race.tryFinishTrack(car);
 // => true
 ```
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -70,7 +70,7 @@ car.distanceDriven();
 
 ## 6. Check if a remote control car can finish a race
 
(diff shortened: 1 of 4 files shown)


ericbalawejder:
The implementation is left to the users as an exercise and that is where the creating problem solving happens. An exception would be we when we are encouraging the users not to use an existing library function to solve a problem. This problem can be solved with arithmetic in O(1) time or looping through in O(n) time and that is the goal of solving data structure and algorithm style problems. If we don't want to expose the state of the car in the method, maybe we shouldn't pass in a NeedForSpeed object `carCanFinish(NeedForSpeed car)` that allows us to access it's state?

ynfle:
Concept exercises are supposed to only have one ideal solution.
I don't think the method should be able to/have to access state.
The solution in the repo uses the O(n) version that doesn't access the state

ericbalawejder:
I understand that the exemplar solution uses a loop and the mentor notes can sometimes reference the solution but do you think by simply changing the method name that it will change the way people solve the problem?

ynfle:
For me personally, for sure

ynfle:
Bump

ynfle:
Bump

exercises/concept/need-for-speed/.docs/instructions.md
```diff
@@ -80,6 +80,6 @@ var car = new NeedForSpeed(speed, batteryDrain);
 int distance = 100;
 var race = new RaceTrack(distance);
 
-race.carCanFinish(car);
+race.tryFinishTrack(car);
 // => true
 ```
```
andrerfcsantos, line 85:
```suggestion
car.distanceDriven()
// => 0

race.tryFinishTrack(car);
// => true

car.distanceDriven()
// => 100
```
```

Make it more explicit in the example that the car drove that distance when calling `tryFinishTrack`.

File: exercises/concept/need-for-speed/.docs/instructions.md [hunk 2]
```diff
@@ -70,7 +70,7 @@ car.distanceDriven();
 
 ## 6. Check if a remote control car can finish a race
 
-To finish a race, a car has to be able to drive the race's distance. This means not draining its battery before having crossed the finish line. Implement the `RaceTrack.carCanFinish()` method that takes a `NeedForSpeed` instance as its parameter and returns `true` if the car can finish the race; otherwise, return `false`:
+To finish a race, a car has to be able to drive the race's distance. This means not draining its battery before having crossed the finish line. Implement the `RaceTrack.tryFinishTrack()` method that takes a `NeedForSpeed` instance as its parameter and returns `true` if the car can finish the race; otherwise, return `false`:
```
andrerfcsantos, line 73:
```suggestion
To finish a race, a car has to be able to drive the race's distance. This means not draining its battery before having crossed the finish line. Implement the `RaceTrack.tryFinishTrack()` method that takes a `NeedForSpeed` instance as its parameter and returns `true` if the car can finish the race; otherwise, return `false`. To see if the car can finish the race, you should try to drive the car until either you reach the end of the track or the battery drains:
```

andrerfcsantos (changes requested):
I agree that `tryFinishTrack` is less ambiguous than `canFinishTrack` and can steer people into the desired solution.

However, I also think that it might not be enough for some people and added a bit to the description and to the example to make it clearer that the car must be driven. Let me know what you think.

exercises/concept/need-for-speed/.docs/instructions.md
ynfle: Should this rather be
> To see if the car can finish the race, you should try to drive the car either until you reach the end of the track or the battery drains

I'm not an expert grammarian, though

andrerfcsantos: I'm not a native speaker, but I think both forms work equally well? I see "until either" in some articles, and Grammarly doesn't complain with either form. I think we can leave as is.

andrerfcsantos (approved):
🚀

Issue #2098, opened 2022-03-16
Exercises Perfect Number and Say have no starter implementation


cdpaiva:
Exercises [Perfect Numbers](https://github.com/exercism/java/tree/main/exercises/practice/perfect-numbers) and [Say](https://github.com/exercism/java/tree/main/exercises/practice/say) are both of difficulty 3 and should have starter implementations, as described in the [policies file](https://github.com/exercism/java/blob/f94f2926542fc18d43e07cca27a11c419d4cca80/POLICIES.md#starter-implementations).

PR #2099, opened 2022-03-16
Added starter implementation for Perfect Numbers and Say

cdpaiva:
Fixes Issue #2098 

Added stubs of the required methods for both exercises.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/perfect-numbers/src/main/java/NaturalNumber.java
+++ b/exercises/practice/perfect-numbers/src/main/java/NaturalNumber.java
@@ -1,5 +1,6 @@
-class NaturalNumber {
+public class NaturalNumber {
 
-
-
+    public Classification getClassification() {
+        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");  
+    }
 }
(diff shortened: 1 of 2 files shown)


ericbalawejder (approved):
In the [policy docs](https://github.com/exercism/java/blob/f94f2926542fc18d43e07cca27a11c419d4cca80/POLICIES.md#starter-implementations), It states to "minimize the accessibility of classes and members". This can be accomplished by omitting the access modifier `public`. Based on the other implementations using it, It doesn't seem like a concern.

cdpaiva:
> In the [policy docs](https://github.com/exercism/java/blob/f94f2926542fc18d43e07cca27a11c419d4cca80/POLICIES.md#starter-implementations), It states to "minimize the accessibility of classes and members". This can be accomplished by omitting the access modifier `public`. Based on the other implementations using it, It doesn't seem like a concern.

I missed that part completely! I was in doubt weather to omit or not the access modifier. It didn't look like a problem and a few other exercises use `public` too (although I'd say the majority doesn't use) so I went for it. Thanks a lot for the review, I'll double check the policy docs next time. Until next time!

ericbalawejder:
The majority of the exercises use the default access, no modifier. Since new exercises were created for v3 and organized into concept and practice, I've noticed more `public` access modifiers. I don't know if it is worth being that picky about or that if it would bring more learning value to the exercises? It seems like a more advance topic when building applications with abstract classes, databases and frameworks.

PR #2112, opened 2022-04-23
fixed emptyStringIsNotTheSameAsNull test to align with instructions

IAmLuisJ:
# pull request

Based on the example in the README.md file, a blank string ("") should return "One for you, one for me". The test currently is looking for a blank response. Fixing the test to align with readme expectations of the problem. 

While in Java null != empty, the directions state they should both return "one for you, one for me"

Readme excerpt:
```Here are some examples:

|Name    |String to return
|:-------|:------------------
|Alice   |One for Alice, one for me.
|Bob     |One for Bob, one for me.
|        |One for you, one for me.
|Zaphod  |One for Zaphod, one for me.
```

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks for this!

PR #2115, opened 2022-04-27
Add warriors and wizards test

rowanlea:
# pull request

An issue arose when mentoring the exercise, where the person doing it had `return String.format("Fighter is a Warrior");` in their Warrior class, and knew that it was wrong. The tests however passed this incorrect behaviour.

The test I have added ensures they call the override in the correct class - the base class. I believe they may need to Google to use something like `.getClass().getSimpleName()` to figure out the class name.

One more thing to consider is that they may implement the override in each class individually, which would still be wrong really. This is probably better left up to the mentor to guide them where they could improve, as it seems the simplest way of guiding them, and won't add any extra confusion. This test at least helps guide them in the right direction.

Apologies if I do this PR wrong, it is actually my first time using the "fork" feature, I'm used to just branching.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
+++ b/exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
@@ -5,9 +5,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class FighterTest {
 
     @Test
-    public void testToString() {
+    public void testWarriorToString() {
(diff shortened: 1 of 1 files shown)


donaldli2020:
no

donaldli2020:
get rid of the label

andrerfcsantos (approved):
Thanks for this!

Sorry it took so long to get this approved :)

PR #2122, opened 2022-05-03
Use 'full' exceptionFormat

ErikSchierboom:
This should help with debugging errors.

Closes https://github.com/exercism/java/issues/2121	

Note: students won't yet be prompted to update because the `build.gradle` file isn't taken into account. Once https://github.com/exercism/website/pull/2628 is merged, we can fix this.

---

Reviewer Resources:

CC @SleeplessByte

### DIFF
--- a/_template/build.gradle
+++ b/_template/build.gradle
@@ -17,7 +17,7 @@ dependencies {
 
 test {
   testLogging {
-    exceptionFormat = 'short'
+    exceptionFormat = 'full'
     showStandardStreams = true
     events = ["passed", "failed", "skipped"]
   }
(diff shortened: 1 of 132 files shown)


andrerfcsantos:
@ErikSchierboom Before this PR is merged, should we add the `files.invalidator` key with `build.gradle` in the `config.json` for all exercises?

ErikSchierboom:
> should we add the files.invalidator key with build.gradle in the config.json for all exercises?

We could! The question then becomes: would we want any change to a `build.gradle` file to always trigger an exercise update notification? In other words: how likely is it that the `build.gradle` file is updated with changes we don't care about? And similarly: how often would it be updated?

andrerfcsantos (approved):
I also believe this won't be changed often. I think we can merge this and then add the `build.gradle` to the `files.invalidator` key when needed.

PR #2129, opened 2022-06-11
Update TwoferTest.java

### LINKED ISSUE #2128: Two Fer: test emptyStringIsNotTheSameAsNull
It's a bit confusing that the test named `emptyStringIsNotTheSameAsNull` actually expects that an empty string _is_ the same as null. If null, we should get "One for you, one for me.". If an empty string is _not_ the same as null, then the greeting should be "One for , one for me." So, perhaps either the test should be renamed or the expectation should be fixed.

```java

/* Track specific */
@Test
public void emptyStringIsNotTheSameAsNull() {
    assertThat(twofer.twofer("")).isEqualTo("One for you, one for me.");
}
```

bobahop:
FIxes #2128

Remove `test emptyStringIsNotTheSameAsNull` to align with problem specifications. See issue #2128

# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/two-fer/src/test/java/TwoferTest.java
+++ b/exercises/practice/two-fer/src/test/java/TwoferTest.java
@@ -33,13 +33,4 @@ public class TwoferTest {
                 .isEqualTo("One for Bob, one for me.");
     }
 
-    /* Track specific */
-
-    @Ignore("Remove to run test")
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks!

PR #2130, opened 2022-06-15
Clarify Instruction to state that characters that are not letters (in…

emiliolg:
…cluding digits, etc) should be removed.

Add a test that showed that more clearly.

# pull request

Doing Mentoring I noticed that instructions are not clear about removing digits and so on. 
I checked the C# track and instructions were clear. So I update the Java instructions to reflect that.
I also add a test that an additional test to show removing no letters while keeping them.

Updae hints.md

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
+++ b/exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
@@ -38,8 +38,9 @@ public class SqueakyCleanTest {
     public void string_with_no_letters() {
         assertThat(SqueakyClean.clean("\uD83D\uDE00\uD83D\uDE00\uD83D\uDE00")).isEmpty();
     }
+
     @Test
-    public void remove_no_letters() {
+    public void keep_only_letters() {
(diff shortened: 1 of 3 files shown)


exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
```diff
@@ -38,6 +38,10 @@ public void ctrl() {
     public void string_with_no_letters() {
         assertThat(SqueakyClean.clean("\uD83D\uDE00\uD83D\uDE00\uD83D\uDE00")).isEmpty();
     }
+    @Test
+    public void remove_no_letters() {
```
andrerfcsantos, line 42:
```suggestion

    @Test
    public void keep_only_letters() {
```

Maybe this is a more appropriate name for the test?

andrerfcsantos (changes requested):
Looks cool, left some minor suggestions.

Note that when changing the tasks, [hints.md](https://github.com/exercism/java/blob/main/exercises/concept/squeaky-clean/.docs/hints.md) should also be modified, and the section names must match the names for the tasks in the instructions.

emiliolg (commented):
Fix hints.md

andrerfcsantos (approved):
Thanks for this.

PR #2131, opened 2022-06-20
Squeaqy clean test

emiliolg:
# pull request

Add a test to check  kebab to camel-case when the first character after the hyphen is not a letter.
For example: "a-1c" is converted to "aC"

Add more cases to the `ctrl` test. 
Previously the only character tested was '\0'

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks for this! (and sorry for the delay in reviewing it)

PR #2132, opened 2022-06-24
Confusing Exception name

myucesan:
It said AssemblyLine instead of CarsAssemble which caused unnecessary confusion.

# pull request

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks for this!

PR #2141, opened 2022-07-14
Fix rename the url path for "master" to "main" and check the URLs

pedrobragadev:
# FIX URLs with "Not found" status

**Problem:** Some URLs it was getting NOT_FOUND (404) because some old refactor

**Solution:** Renamed most of this URLs to some its working

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks!

PR #2149, opened 2022-07-28
Update NeedForSpeed.java

dsletten:
The hints for `batteryDrained()` literally state: Remember that if the battery charge is less than the battery drain percentage, it is considered drained.
The battery need not be 0 to be considered drained.
The semantics of `tryFinishTrack()` are misleading. The termination criterion is not an empty battery but rather completing the distance of the track.

# pull request

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks for this!

Made a commit in your branch just fixing some formatting issues that were failing the CI tests.

Issue #2151, opened 2022-07-30
hello-world: remove the reference to `UnsupportedOperationException` in instructions


andrerfcsantos:
The instructions for the "Hello World" exercise tell the student to remove an `UnsupportedOperationException`:

https://github.com/exercism/java/blob/9abbe2a744625d698691249714d992fbdc17662c/exercises/practice/hello-world/.docs/instructions.append.md?plain=1#L52-L59

However, currently neither the tests or the solution contain an exception, so these instructions are not relevant anymore and must be removed to reflect the current [stub files](https://github.com/exercism/java/tree/main/exercises/practice/hello-world/src) provided.

unfirthman:
Hi @andrerfcsantos 

If this issue is still open, I'd love to try to tackle it.

andrerfcsantos:
@unfirthman Go for it :)

suraj-mandal:
Hi, is this issue open? I was getting updates of this issue being open, but it is stated that a fix has been merged. Can you close if this issue if the problem has been resolved.

ShadowZone:
I just solved the exercise and the error is still there. Stated UnsupportedOperationException in text but is not found in initial code block

andrerfcsantos:
I'll close this issue, as the main goal of fixing the instructions to not use the exception was achieved.

However, it seems there's a reference to the `UnsupportedOperationException` in the table of contents for the exercise, but I made a separate issue for that #2167

andrerfcsantos:
The table of contents issue is now fixed by #2166 which was already merged.

@ShadowZone If there are other places where the instructions should be updated that I'm not seeing, please feel free to create a new issue or PR.

PR #2153, opened 2022-07-30
removed the reference to UnsupportedOperationException in instructions

unfirthman:
# pull request remove the reference to UnsupportedOperationException in instructions #2151 

I deleted lines 52 - 59 in [java/exercises/practice/hello-world/.docs](https://github.com/exercism/java/blob/9abbe2a744625d698691249714d992fbdc17662c/exercises/practice/hello-world/.docs/instructions.append.md?plain=1#L52-L59)

## This is to solve issue #2151 

Thanks @andrerfcsantos for the issue assignment! Always happy to contribute to building Exercism. 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/hello-world/.docs/instructions.append.md
+++ b/exercises/practice/hello-world/.docs/instructions.append.md
@@ -50,15 +50,16 @@ if you want to solve the problem and run tests locally check these links below:
 
 
 
-[`Exception`s](https://docs.oracle.com/javase/tutorial/essential/exceptions/)
-are often used in Java to draw the attention of a developer when something goes
-wrong. In our case, nothing is wrong per se; we just haven't written our
-solution yet! The use of an
-[`UnsupportedOperationException`](http://docs.oracle.com/javase/8/docs/api/?java/lang/UnsupportedOperationException.html)
-in this situation is designed to remind us of exactly this fact. It effectively
-says: "Your Code Goes Here!".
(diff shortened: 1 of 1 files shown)


andrerfcsantos (changes requested):
Thanks for this.

Here we probably still want to have a step 1, but we shouldn't mention exceptions at all in the text. This step 1 can just say that the student has to modify the solution file provided just so the function returns "Hello, World." and show the example that it is at the end of step 1 currently.

The title of the new step 1 can be something like `Fix the solution`

unfirthman:
Understood. I will make these changes tomorrow and resubmit.

Thank you!

unfirthman (commented):
# Edit 1

## Remove reference to UnsupportedOperationException in instructions #2153 

### Also fixed Step 1 instructions per previous comment @andrerfcsantos 

Thank you with your patience! I believe this should be closer to what we are looking for.

My newest commit aims to clarify expectations in Step 1:

New commit: 23218428602a4b95e3d8434d386e90be8a8c0655

unfirthman:
Hello @andrerfcsantos  - just checking in. Did you have a chance to look over this next edit? No rush!!!

andrerfcsantos (approved):
Sorry for the delay in reviewing this.

There was just one place where the exception was being referred, but since it was a simple final fix, I just did it for you.

Thank you for this!

PR #2154, opened 2022-08-02
Update PythagoreanTripletTest.java

dsletten:
None of the existing withFactorsLessThanOrEqualTo() builder calls make a bit of difference. The limit must be less than half of the desired sum.

# pull request

---

Reviewer Resources:


andrerfcsantos (approved):
Thanks for this!

PR #2161, opened 2022-08-29
Numbers: Cars, Assemble! Refactored unit tests in a way that ignores machine epsilon

izzettunc:
# pull request

### Background:
I am creating this pull request as one of my students stumble upon a test case where it failed due to machine epsilon and in this pr, I have refactored the unit tests in a way that it ignores machine epsilon as i think it's too early to challenge students with this kind of errors and it was probably unintentional for this particular exercise.

### Reproduction:
While this issue can be avoided by calculating productionRatePerHour function as follows:
```defaultProductionRate  * speed * successRate(speed) ```

The issue can be reproduced by calculating productionRatePerHour function as follows which confuses students:
```defaultProductionRate * successRate(speed) * speed```

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
+++ b/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
@@ -7,6 +7,7 @@ import static org.assertj.core.api.Assertions.*;
 public class CarsAssembleTest {
 
     private CarsAssemble carsAssemble;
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for this, it is a good idea 👍

andrerfcsantos:
Made just a commit fixing some formatting on the file which was preventing CI tests to pass.

Issue #2133, opened 2022-06-26
Add missing stubs to exercises


Dark-WizZ:
## The problem

Right now for some exercises, a student is required to read the tests to know which functions they should implement. This is annoying, time consuming and might discourage people from doing the exercise entirely. We should provide stubs for all exercises. Stubs should provide the signature of functions and classes, just so the student is immediately ready to start writing the custom logic. This is an example of a stub, from the [Raindrops exercise](https://github.com/exercism/java/blob/main/exercises/practice/raindrops/src/main/java/RaindropConverter.java):

```java
class RaindropConverter {

    String convert(int number) {
        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
    }

}
```

All stubs must follow this format.

## Contributing to this issue

This is a big issue that will require changes to a lot of exercises. The good news is that you don't need to fix this is all exercises to contribute to this issue! PRs fixing this in just one exercise at a time or a couple of exercises at a time are more than welcome.

andrerfcsantos:
I absolutely agree we should do this!

@Dark-WizZ I changed the description of the issue just to make the issue and what needs to be done clearer.

andrerfcsantos:
@jcovington16 Go ahead! Are you considering changing this is all exercises or only in a subset of them?

jcovington16:
Just a subset. After this one I'll be able to knock out some more difficult tasks. Just trying to work myself up to it.

andrerfcsantos:
@jcovington16 You absolutely can work on this issue, sorry I haven't made that clear! What I meant to say before is that since you won't be working on all the exercises and only a subset of them, I won't mark the issue as claimed just so other people can join alongside you. it's totally ok for several people to work on this issue at the same time, ideally looking at different exercises.

jcovington16:
Oh ok I understand now. I'll finish up the PR I have now and get started with this one before I dive into the analyzer. This one shouldn't take me long.

jcovington16:
@Dark-WizZ @andrerfcsantos Do we know specifically which exercises need these stubs implemented? And which tracks as well?

andrerfcsantos:
It seems here in the Java track, all "medium" difficulty exercises or above don't include stubs, so probably those are a good start, but there might be more. A quick `jq` query for exercises that are medium difficulty or above gives me this:

<details>
  <summary>jq query</summary>
  
```console
$ jq '.exercises.practice | map(select(.difficulty >= 4)) | map(.slug) | sort' config.json
[...]

PR #2188, opened 2022-10-23
Add stubs to practice exercises W to Z

PakkuDon:
# pull request

Related to #2133 

This PR adds stubs for practice exercises starting with letters W through to Z. 
[`yacht`](https://github.com/exercism/java/blob/main/exercises/practice/yacht/src/main/java/Yacht.java) has been skipped as it already has a stub provided 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/word-search/src/main/java/WordSearcher.java
+++ b/exercises/practice/word-search/src/main/java/WordSearcher.java
@@ -1,10 +1,7 @@
-/*
+import java.util.*;
 
-Since this exercise has a difficulty of > 4 it doesn't come
-with any starter implementation.
-This is so that you get to practice creating classes and methods
-which is an important part of programming in Java.
-
-Please remove this comment when submitting your solution.
-
-*/
+class WordSearcher {
+    Map<String, Optional<WordLocation>> search(final Set<String> words, final char[][] grid) {
+        throw new UnsupportedOperationException("Please implement the WordSearcher.search() method.");
+    }
+}
--- a/exercises/practice/zipper/src/main/java/Zipper.java
+++ b/exercises/practice/zipper/src/main/java/Zipper.java
@@ -1,10 +1,43 @@
-/*
+class Zipper {
+    Zipper(int val) {
+        throw new UnsupportedOperationException("Please implement the Zipper(int) method.");
+    }
 
-Since this exercise has a difficulty of > 4 it doesn't come
-with any starter implementation.
-This is so that you get to practice creating classes and methods
-which is an important part of programming in Java.
+    BinaryTree toTree() {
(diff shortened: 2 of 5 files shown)


exercises/practice/zipper/src/main/java/Zipper.java
```diff
@@ -1,10 +1,43 @@
-/*
+class Zipper {
+    Zipper(int val) {
+        throw new UnsupportedOperationException("Please implement the Zipper(int) method.");
```
andrerfcsantos, line 3:
```suggestion
        throw new UnsupportedOperationException("Please implement the Zipper(int) constructor.");
```

exercises/practice/word-search/src/main/java/WordSearcher.java
```diff
@@ -1,10 +1,7 @@
-/*
+import java.util.*;
```
andrerfcsantos, line 1:
```suggestion
import java.util.Map;
import java.util.Optional;
```

Avoiding using the * import, as it can be considered bad practice. It's better to be explicit in this case.

exercises/practice/zipper/src/main/java/Zipper.java
```diff
@@ -1,10 +1,43 @@
...
+    }
+
+    void setValue(int val) {
+        throw new UnsupportedOperationException("Please implement the Zipper.setValue() method.");
+    }
+}
+
+class BinaryTree {
+    BinaryTree(int value) {
+        throw new UnsupportedOperationException("Please implement the BinaryTree(int) method.");
```
andrerfcsantos, line 29:
```suggestion
        throw new UnsupportedOperationException("Please implement the BinaryTree(int) constructor.");
```

exercises/practice/zipper/src/main/java/Zipper.java
```diff
@@ -1,10 +1,43 @@
...
+    }
+
+    BinaryTree(Zipper root) {
+        throw new UnsupportedOperationException("Please implement the BinaryTree(Zipper) method.");
```
andrerfcsantos, line 33:
```suggestion
        throw new UnsupportedOperationException("Please implement the BinaryTree(Zipper) constructor.");
```

andrerfcsantos (changes requested):
Thanks for doing this! Left some suggestions.

PakkuDon:
Thank you for your feedback @andrerfcsantos! 
I've updated the stubs. Let me know how it looks

andrerfcsantos (approved):
Thanks!

CI was failing due to a missing `java.utils.Set` import I didn't suggest earlier, but I committed that change.

Issue #2157, opened 2022-08-24
Secret Handshake exercise should have `Signal` class/enum visible/editable


artamonovkirill:
More context: https://exercism-team.slack.com/archives/CATD6MX62/p1661262277270479

For the online editor, it's not clear that the enum is already defined, so students might attempt to redefine it inside `HandshakeCalculator`. That causes deceiving compilation errors.

andrerfcsantos:
Thanks for pointing this out.

The fix here is to add a `files.editor` key in `exercises/practice/secret-handshake/.meta/config.json` containing that extra file - this will make the editor aware of that file and show it in reading mode.

artamonovkirill:
https://github.com/exercism/java/pull/2191
@andrerfcsantos is there any way I can test this change before merging?

andrerfcsantos:
@artamonovkirill Unfortunately, no. I left a comment on the PR.

Thanks for doing this. I noticed this is a problem in a lot of other exercises too - they have extra files but are not declared in either the `files.solution` or `files.editor` key. I've meaning to write a little script to fix instances where this happens, but haven't find the time yet. In the meantime, feel free to create PRs about this for the exercises you are solving.

andrerfcsantos:
It appears for me now 🎉

<img width="526" alt="image" src="https://user-images.githubusercontent.com/6928620/198849949-0da9e429-85fe-4611-979c-b67688ffa658.png">

artamonovkirill:
> @artamonovkirill Unfortunately, no. I left a comment on the PR.
> 
> Thanks for doing this. I noticed this is a problem in a lot of other exercises too - they have extra files but are not declared in either the `files.solution` or `files.editor` key (#2170). I've meaning to write a little script to fix instances where this happens, but haven't find the time yet. In the meantime, feel free to create PRs about this for the exercises you are solving.

@andrerfcsantos I wrote the script and added missing files: https://github.com/exercism/java/pull/2195 😁

PR #2191, opened 2022-10-28
Make Signal class visible in online editor

artamonovkirill:
# pull request

Closes https://github.com/exercism/java/issues/2157

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/secret-handshake/.meta/config.json
+++ b/exercises/practice/secret-handshake/.meta/config.json
@@ -22,7 +22,9 @@
   ],
   "files": {
     "solution": [
-      "src/main/java/HandshakeCalculator.java",
+      "src/main/java/HandshakeCalculator.java"
+    ],
+    "editor": [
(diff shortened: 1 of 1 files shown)


andrerfcsantos (changes requested):
Thanks for this.

Since the `Signal.java` is not supposed to be modified, we can mark it to appear in the editor as a read-only tab using the `files.editor` instead of the `files.solution` key. You can read more about it [here](https://exercism.org/docs/building/tracks/practice-exercises#:~:text=error%20handling%20tests.-,File%3A,.meta/config.json,-Purpose%3A%20Contains) or [this example for the Go track](https://github.com/exercism/go/blob/main/exercises/practice/react/.meta/config.json).

artamonovkirill:
@andrerfcsantos thanks for the tip, updated

andrerfcsantos (approved):

Issue #2147, opened 2022-07-23
Using AssertJ assertions on entire exercises


thethoughtcode:
The test classes use mix of Junit asserts and AssertJ. Since AssertJ assertions are more robust and extensive as compared to Junit5.

Like for example, usage of `assertThatExceptionOfType` from AssertJ is much more readable as compare to `assertThrows` from JUnit 5.

```java
assertThatExceptionOfType(RuntimeException.class)
         .isThrownBy(() -> { throw new RuntimeException(new IllegalArgumentException("boom!")); })
         .havingCause()
         .withMessage("boom!");
```

We can migrate all the asserts from Junit to AsserJ.

## Working on this issue

This issue affects several exercises. If you want to contribute to this issue, you don't have to change this in all exercises, feel free to submit small PRs fixing this issue for a single exercise or a small subset of exercises!

andrerfcsantos:
This sounds like a good idea, thanks for suggesting!

I edited the description with how to contribute to this issue.

anandhakrishnanaji:
I'm working on this

suraj-mandal:
Hi, this is my first time, I will be contributing towards open source. May I work on this issue, and migrate some of the tests from junit5 to AssertJ?

danielmrcl:
Convert exception assertions to AssertJ in classes starting with 'A' and 'B'. #2156

jennylia:
I am working on the Haming

jennylia:
https://github.com/exercism/java/pull/2178 is my PR

jennylia:
https://github.com/exercism/java/pull/2181 another PR

qwet11:
Hi. First time contributing to open source. Is this issue resolved?

sanderploegsma:
I am in favour of this change, but it does mean that we have to revise the existing policy in POLICIES.md that states that we prefer JUnit assertions over AssertJ.

pavanbaloju:
Hello @sanderploegsma, I would like to contribute to a few exercises.

pavanbaloju:
Raised one more PR for few more tests. Please review !

pavanbaloju:
Raised one more PR for few more tests. Please review!

pavanbaloju:
Raised one more PR #2513 which covers all the pending tests which needs to be updated

sanderploegsma:
I found three more practice exercises to update.

- Grade School
- Tree Building
- Zipper

[...]

PR #2202, opened 2022-11-16
assertEquals converted to assertThat for MarkdownTest

smcg468:
related to #2147 
Importing AssertJ and changing all assertEquals to assertThat.

Example of Change:

assertEquals(expected, markdown.parse(input));

To

assertThat(markdown.parse(input)).isEqualTo(expected);


andrerfcsantos (approved):
Thanks for this!

Issue #2207, opened 2022-11-18
Seperate Three Domino Loops Test missing from DominoesTest.java


smcg468:
I was looking through the test cases outlined in problem-specifications, specifically the canonical-data json file and noticed that the 'Three seperate-domino loops' test case was not covered in the DominoesTest.java. I checked the C# repository and noticed the test was present there, I therefore have added in the test and will create a pull request for you to review.

Thanks.

andrerfcsantos:
Merged in https://github.com/exercism/java/pull/2209

PR #2209, opened 2022-11-18
I have added in the separate three-domino-loops test to the DominoesT…

smcg468:
#2207 

I have updated the DominoesTest.java file to include the three separate domino-loops test case that was mentioned in the canonical-data.json file but not present in the DominoesTest.java file. 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/dominoes/src/test/java/DominoesTest.java
+++ b/exercises/practice/dominoes/src/test/java/DominoesTest.java
@@ -184,6 +184,7 @@ public class DominoesTest {
     }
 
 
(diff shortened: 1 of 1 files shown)


smcg468:
Thought this was due to a mismatch in gradle versions but doesn't seem to be. Will look at the logs and try see if I can resolve the issue as to why the build is failing.

andrerfcsantos (approved):
@smcg468 Sorry for the delay in reviewing and thanks for this! 🚀

PR #2213, opened 2022-11-23
update Darts.java score method signature to match test class parameter types

thomasaziz:
# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/darts/.docs/instructions.md
+++ b/exercises/practice/darts/.docs/instructions.md
@@ -7,11 +7,11 @@ throw darts to a [target](https://en.wikipedia.org/wiki/Darts#/media/File:Darts_
 
 In our particular instance of the game, the target rewards with 4 different amounts of points, depending on where the dart lands:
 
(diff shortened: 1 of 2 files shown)


BastienVH:
I agree with the change about the argument variable type. These should be doubles and not ints.
Your change from radius to diameter is incorrect. The original wording should be kept.
As seen in the image below the diameter is radius * 2.
image
In test case 2, the dart is expected to be on the outer circle with coordinates x: 0, y:10.
If the diameter is 10, the radius is 5 and the dart would fall far outside the outer circle.

andrerfcsantos (approved):
@thomasaziz Thanks for this and sorry for the delay in reviewing it.

I accepted the change of making the parameters `double`, but rejected the changes to the instructions. Those instructions are taken from the [problem-specifications repository](https://github.com/exercism/problem-specifications/blob/main/exercises/darts/description.md), so any problems with them should be discussed there, so all tracks can benefit from it.

PR #2235, opened 2023-01-21
Fix: Change to False

copydataai:
Why: because if the prisoner is sleeping he cannot be rescued

# pull request

link from the problem: https://exercism.org/tracks/java/exercises/annalyns-infiltration
Text of the logic or validations: 
Free prisoner: Annalyn can try sneaking into the camp to free the prisoner. This is a risky thing to do and can only succeed in one of two ways:
If Annalyn has her pet dog with her she can rescue the prisoner if the archer is asleep. The knight is scared of the dog and the archer will not have time to get ready before Annalyn and the prisoner can escape.
If Annalyn does not have her dog then she and the prisoner must be very sneaky! Annalyn can free the prisoner if the prisoner is awake and the knight and archer are both sleeping, but if the prisoner is sleeping they can't be rescued: the prisoner would be startled by Annalyn's sudden appearance and wake up the knight and archer.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/annalyns-infiltration/src/test/java/AnnalynsInfiltrationTest.java
+++ b/exercises/concept/annalyns-infiltration/src/test/java/AnnalynsInfiltrationTest.java
@@ -133,7 +133,7 @@ public class AnnalynsInfiltrationTest {
         boolean prisonerIsAwake = false;
         boolean petDogIsPresent = true;
         assertThat(AnnalynsInfiltration.canFreePrisoner(knightIsAwake, archerIsAwake,
-                   prisonerIsAwake, petDogIsPresent)).isTrue();
+                   prisonerIsAwake, petDogIsPresent)).isFalse();
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


andrerfcsantos:
Hey!

> Why: because if the prisoner is sleeping he cannot be rescued

This is only valid for the second way the prisoner can be rescued. The first way (archer sleeping) doesn't depend on whether the prisoner is awake or not:

> If Annalyn has her pet dog with her she can rescue the prisoner if the archer is asleep. The knight is scared of the dog and the archer will not have time to get ready before Annalyn and the prisoner can escape.

In both the test cases you changed this happens (archer sleeping and dog present), so it falls under this first scenario.

However, I do see how the instructions can be contradicting. The first scenario claims that if the archer is sleeping:

> `the archer will not have time to get ready` but the second scenario claims that 

But the second scenario claims that the archer can be awakened and now that's a problem:

> if the prisoner is sleeping they can't be rescued: the prisoner would be startled by Annalyn's sudden appearance and wake up the knight and archer.

If you'd like, feel free to take this discussion [to the forum](https://forum.exercism.org/) and continue the discussion there. Since this exercise is present in several tracks, posting this problem there and gathering opinions from people from different tracks can be useful.

PR #2240, opened 2023-01-25
Fix typo in the Dig Deeper text for Grains

nomicos:
Fixed the wrong power of two that gives us 8.
https://exercism.org/tracks/java/exercises/grains/dig_deeper

---

Reviewer Resources:


bobahop:
Good catch! This is probably in all the other Grains approaches I've done for other tracks...

nomicos:
@bobahop Is there a centralised place where this text can be amended? If not, I would be happy to have this contribution approved at least for the Java track; and then the mistake can be fixed in other repositories 🙂

And thank you for all the work you put into these articles!

bobahop:
@nomicos : thanks, but there is no centralized place, so I've already made PRs for the other tracks, some of which have already been approved.

andrerfcsantos:
Closing and reopening to refresh CI.

andrerfcsantos (approved):
@nomicos Thanks!

PR #2244, opened 2023-01-27
Isogram approaches

bobahop:
# pull request

Approaches for Isogram

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/isogram/.approaches/filter-map-allmatch/content.md
+++ b/exercises/practice/isogram/.approaches/filter-map-allmatch/content.md
@@ -1,0 +1,49 @@
+# `filter()` and `map()` with `allMatch()`
+
+```java
+import java.util.HashSet;
+
+final class IsogramChecker {
+
+    boolean isIsogram(final String phrase) {
+        return phrase.chars()
+            .filter(Character::isLetter)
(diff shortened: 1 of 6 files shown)


bobahop:
@ErikSchierboom: Java Isogram approaches ready for review.

exercises/practice/isogram/.approaches/filter-map-allmatch/content.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+The surviving codepoints are passed to the [`map()`][map] method which converts the codepoints to lower case.
+
+The lowercased codepoints are passed into the [`allMatch()`][allmatch] method which passes each codepoint
+into the `HashSet`'s [`add()`][add] method.
+The `add()` method returns `true` if the value is not already in the `HashSet` and `false` if the value is already in the `HashSet`.
+If `add()` returns `false`, then the `allMatch()` method can know that it will never be `true` and [short-circuit][short-circuiting]
+with a result of `false`.
+If `add()` always returns `true`, then `allMatch()` will return `true`.
```
ErikSchierboom, line 35: Clever!

ErikSchierboom (approved):

PR #2251, opened 2023-02-01
Pig Latin approach

bobahop:
# pull request

Approach for Pig Latin

---

Reviewer Resources:


bobahop:
@ErikSchierboom: Java Pig Latin approach ready for review. I really don't like the other approaches, so I just did the one. In mentoring I find this approach is usually overlooked.

ErikSchierboom (approved):

ErikSchierboom:
>  I really don't like the other approaches, so I just did the one.

I know which approach you're referring to probably (my regex one :D)

Issue #2170, opened 2022-09-29
Add exercises' extra files to files.editor or files.solution


andrerfcsantos:
Recently in my mentoring, I noticed that some exercises have some files that are part of the solution (but the student shouldn't modify them) that are not appearing on the online editor and they should. The reason for this is that these extra files are not in the `files.solution` or `files.editor` keys of `.meta/config.json` on each exercise.

An example of this is the [Triangle exercise](https://github.com/exercism/java/tree/main/exercises/practice/triangle). The file [TriangleException.java](https://github.com/exercism/java/blob/main/exercises/practice/triangle/src/main/java/TriangleException.java) is something the student must be aware it exists, although the file shouldn't be modified. This file should then be included in the `files.editor` key of `.meta/config.json` of this exercise just it shows up in the editor, both for students and mentors.

There are other exercises with a similar problem and ideally, we should identify and fix all the exercises that have these extra files that are not being acknowledged in the `files.editor` or `files.solution` keys of `.meta/config.json`.

## Contributing to this issue

If you want to contribute to this issue, comment below your intention!

Even though ideally we should fix this in all exercises that suffer from this issue, PRs that fix this in a single exercise or just a few exercises are welcome.

Some useful links:

- [Exercism docs: Practice exercises](https://exercism.org/docs/building/tracks/practice-exercises)

PS: I imagine it wouldn't be difficult to make a script that tries to identify exercises with this issue and try to fix the `.meta/config.json` manually. The script could read the exercises' folders and check for files that are not in `config.json` and add them to `files.editor` by default. If you create such script, feel free to share it with us and use it to create PR(s).

AdarshPrasad02:
The coding playground could include the collapsible function for the student to view if he/she wants to. This will let them know about the present code beforehand.

suraj-mandal:
@andrerfcsantos Hi, I was able to implement the script doing exactly what you have mentioned. I am creating a pull request in the meantime. Can you check if possible?

suraj-mandal:
Hi, @andrerfcsantos I have created the pull request for the same.

andrerfcsantos:
@suraj-mandal Thanks, added my review there.

sanderploegsma:
I went through both the concept and practice exercises and as of right now it looks like all exercises have their files correctly referenced in `.meta/config.json`.

@andrerfcsantos unless you still want to add a script to automatically add unreferenced exercise files to their respective config files, perhaps this issue can be closed.

PR #2252, opened 2023-02-05
Implementation of script to automatically add java files to files.editor of config.json

Greg-GJE3:
# pull request

This pull request is implementation of the script as discussed in #2170 .  As per the discussions, it was desired to create a script that identifies all the exercises having extra java files and adds them to the `files.editor` keys of `config.json`.

This PR basically has created such a script that implements the feature as stated in the discussion. Thus for subsequent exercises there is no need to manually configure the additional java files in the `config.json`.

The script gives the ability to
1. Update all folders
2. Select a folder name and then the script will update the `config.json` file present in that folder _(This feature will help to update future exercises that involve multiple java files)_.

PS: _The script has been tried and tested with all the folders involving multiple java files and works flawlessly._

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/sublist/.meta/config.json
+++ b/exercises/practice/sublist/.meta/config.json
@@ -31,8 +31,7 @@
     ],
     "example": [
       ".meta/src/reference/java/RelationshipComputer.java"
-    ],
-    "editor": []
+    ]
   },
   "blurb": "Write a function to determine if a list is a sublist of another list."
 }
--- a/exercises/practice/sublist/.meta/config.json
+++ b/exercises/practice/sublist/.meta/config.json
@@ -23,7 +23,8 @@
   ],
   "files": {
     "solution": [
-      "src/main/java/RelationshipComputer.java"
+      "src/main/java/RelationshipComputer.java",
+      "src/main/java/Relationship.java"
     ],
     "test": [
       "src/test/java/RelationshipComputerTest.java"
@@ -31,9 +32,7 @@
     "example": [
       ".meta/src/reference/java/RelationshipComputer.java"
     ],
-    "editor": [
-      "src/main/java/Relationship.java"
-    ]
+    "editor": []
   },
   "blurb": "Write a function to determine if a list is a sublist of another list."
 }
--- a/scripts/add_extra_files_to_editor.sh
+++ b/scripts/add_extra_files_to_editor.sh
@@ -9,7 +9,7 @@ command -v jq >/dev/null 2>&1 || {
 ################### HELP ##############################
 print_usage() {
         echo -e "\nDescription:\n"
-        echo -e "This script goes through each exercise directory, picks all the java files in the src directory, except the main editable solution file, and adds them to 'fonts.editor' section of the config.json file present in the .meta folder of the exercise" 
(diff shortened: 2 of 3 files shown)


scripts/add_extra_files_to_editor.sh
```diff
@@ -0,0 +1,126 @@
...
+        java_source_dir="$exercises_location/$1/src/main/java"
+
+        data=$(find "$java_source_dir" -type f -name "*.java")
+
+        # now getting the config.json file and here getting the details of the solution file
+        json_file="$exercises_location/$1/.meta/config.json"
+
+        # this is the solution file here
+        solution_file=$(jq '.| .files.solution[0]' $json_file)
+        solution_file=${solution_file:1:-1} # removing the quotes
```
andrerfcsantos, line 50:
```suggestion
        solution_file=$(jq -r '.files.solution[0]' $json_file)
```

The `-r` flag means "raw output" and makes jq output strings without quotes. Also `${solution_file:1:-1}` was not working on my machine with an `line 50: -1: substring expression < 0` error.

Also, it'd be great if the script could take into account all solution files, and not just the first one. For instance, the exercise [error-handling](https://github.com/exercism/java/tree/main/exercises/practice/error-handling) has 3 files as solution files. The current version of the script adds 2 of them to the `editor` key, when it shouldn't.

andrerfcsantos (changes requested):
@suraj-mandal Thanks for working on this and sorry for the delay in reviewing.

Left a comment below.

suraj-mandal:
Hi @andrerfcsantos , is there any way to determine which file to be considered as a solution and which one to be considered in the editor part?

suraj-mandal:
Hi @andrerfcsantos , I have updated the script as you have requested. Also now it takes all the files in the solution list into consideration rather than the first file. Can you check once?

andrerfcsantos:
@suraj-mandal
> is there any way to determine which file to be considered as a solution and which one to be considered in the editor part?

The best way I think is to take into consideration all the files that are in the source folder and discard the ones already in the solution array. The idea here is that the solution file array contains editable files, so they will appear in the editor by default as editable files. For the editor array, we are interested in files that are part of the source but should be read-only because the student has no benefit from editing them but should see they exist on the editor.

Glancing over the script and with my hand-wavy knowledge of bash I think this is what your version of the script is doing (hadn't the time to test it yet, but will take a closer look later).

exercises/practice/sublist/.meta/config.json
```diff
@@ -23,17 +23,16 @@
...
     "test": [
       "src/test/java/RelationshipComputerTest.java"
     ],
     "example": [
       ".meta/src/reference/java/RelationshipComputer.java"
     ],
-    "editor": [
-      "src/main/java/Relationship.java"
-    ]
+    "editor": []
```
andrerfcsantos, line 35: When the array is empty, the key should be removed entirely. If you want, you can use [configlet](https://github.com/exercism/configlet) locally to check and auto-fix these issues.

suraj-mandal: Sure.. doing it..

suraj-mandal:
Yeah, i am doing difference between the list of Java files in src and the Java files in solution list

suraj-mandal:
Hi! I have updated the script, now if there are no editor files then it will not create the editor key with the empty list in the config.json file. Also I have fixed the empty editor key as well.

PR #2258, opened 2023-02-18
Two Buckets. Add test cases 

rabestro:
# pull request

Add test cases for attempts to measure when the goal is unreachable.

### Test cases:
- 'not possible to reach the goal.'
- 'goal larger than both buckets is impossible.' 
- Test platform upgraded to JUnit 5

---

Reviewer Resources:


andrerfcsantos:
Hi!

Thanks for your interest in contributing. A couple of things:

- For practice exercises, new test cases are discussed in [problem-specifications](https://github.com/exercism/problem-specifications) first, before they should be added to all tracks that implement it.
- All exercises should use the same dependencies, unless there is a strong reason not to. Changes to the dependencies also must be reflected in the [java-test-runner](https://github.com/exercism/java-test-runner), but don't see a reason to have groovy tests just for a single exercise and use java for others.

For these reasons I'll close this, but by all means feel free to continue this discussion here or [on the forum](https://forum.exercism.org/).

PR #2259, opened 2023-02-19
Fixing the instructions example.

BoonAtW0rk:
# pull request

From the text written in the introduction of the task, a example of an invalid DNA should follow. In the given example a valid DNA (contains only the letters A, C, G, or T) is given, but the output shows an error.

Therefor the example got changed, so an actual invalid DNA is given and the expected information about the invalid DNA is correct.

---

Reviewer Resources:


andrerfcsantos:
Changes to problem descriptions of practice exercises must be discussed first in the problem-specifications repo and there this change isn't specified: https://github.com/exercism/problem-specifications/blob/main/exercises/nucleotide-count/description.md

PR #2265, opened 2023-02-26
2257 birds watcher remove test

### LINKED ISSUE #2257: Bird Watcher - Empty array scenario
Hi,

This is the exercise: https://exercism.org/tracks/java/exercises/bird-watcher

One of the general tips of the exercise states that:

> The bird count per day is an array that contains exactly 7 integers.

But there is a test that contemplates an empty array scenario:

``` Java
    @Test
    public void itShouldReturnZeroIfBirdWatcherLastWeekIsEmpty() {
        int[] lastWeekEmpty = new int[0];
        birdWatcher = new BirdWatcher(lastWeekEmpty);
        assertThat(birdWatcher.getToday()).isEqualTo(0);
    }
```

My suggestion is to remove that test.

If testing the exercise against an empty array is desired I would suggest removing the hint and also implement the same case for other methods. But, in my opinion (since this is one of the first exercises), having a pre-condition of an existing array keep thing simple and seems the path to follow.

Please let me know if removing the test sounds good and I will submit a PR.

Thanks.

vorenusCoA:
# pull request

Closes #2257 

Since the pre-condition for the exercise is that we have a fixed number of elements in the array there should not be a test checking for an empty array. Having that test will prevent some valid solutions not to pass.

Two files have been modified:

1) exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java: removing the conflicting test

2) exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java: removing the check for a length of zero in the reference code since that check is not necessary

---

Reviewer Resources:


andrerfcsantos (commented):
Thanks!

PR #2266, opened 2023-02-26
Add react exercise

### LINKED ISSUE #1815: react: implement exercise
The exercise **react** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/master/exercises/react). 

kahgoh:
# pull request

Add react exercise.

Closes #1815 

---

Reviewer Resources:


exercises/practice/react/src/main/java/React.java
```diff
@@ -0,0 +1,44 @@
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public class React<T> {
```
sanderploegsma, line 5: I don't think the type parameter is required here, right?

exercises/practice/react/src/main/java/React.java
```diff
@@ -0,0 +1,44 @@
...
+
+    interface Cell<T> {
+        T getValue();
+
+        void addCallback(Consumer<T> consumer);
+
+        void removeCallback(Consumer<T> callback);
+    }
+
+    static class InputCell<T> implements Cell<T> {
```
sanderploegsma, line 15: I'm not sure it makes sense to provide a stub class for input cells here, while not providing a stub class for compute cells. This might lead a student to get confused as for what to implement themselves.

Maybe it makes more sense to only add interfaces in the stub file, so that it's up to the student to derive concrete classes for them.

``` java
public class React {
    interface Cell<T> {
        T getValue();
    }

    interface InputCell<T> extends Cell<T> {
        void setValue(T newValue);
    }

    interface ComputeCell<T> extends Cell<T> {
        void addCallback(Consumer<T> callback);
        void removeCallback(Consumer<T> callback);
    }

    public static <T> InputCell<T> inputCell(T initialValue) {
        throw new UnsupportedOperationException("Please implement the React.inputCell() method");
    }

    public static <T> ComputeCell<T> computeCell(Function<List<T>, T> function, List<Cell<T>> cells) {
        throw new UnsupportedOperationException("Please implement the React.computeCell() method");
    }
}
```

exercises/practice/react/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+
+repositories {
+  mavenCentral()
+}
+
+gradle.projectsEvaluated {
+  tasks.withType(JavaCompile) {
+    options.compilerArgs << "-Xlint:unchecked"
+  }
+}
```
sanderploegsma, line 17: Any particular reason to add this setting for this exercise? Other exercises don't seem to need it.

sanderploegsma (changes requested):
Thanks for your contribution! I left a couple of inline comments, but overall it already looks very good.

exercises/practice/react/src/main/java/React.java
kahgoh: That sounds reasonable. I'll make the update.

exercises/practice/react/build.gradle
kahgoh: No, I can't remember why I added it in the first place ... I'll remove it.

config.json
```diff
@@ -1553,6 +1553,21 @@
...
+        "name": "React",
+        "uuid": "33531718-1d8a-4abd-bd2a-090303ad0c39",
+        "practices": [
+          "classes",
+          "generic-types"
+        ],
+        "prerequisites": [
+          "classes",
+          "generic-types"
+        ],
```
sanderploegsma, line 1567: I would probably replace `classes` with `interfaces` as a prerequisite, since that concept is vital for this exercise right now. No need to have both as prerequisites either, since one requires the other.

So I suggest the following:
- Prerequisites: `interfaces`, `generic-types`
- Practices: `classes`, `interfaces`, `generic-types`

exercises/practice/react/.meta/src/reference/java/React.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    interface ComputeCell<T> extends Cell<T> {
+        void prepareTentativeUpdate();
+        void commitTentativeUpdate();
+        void addCallback(Consumer<T> callback);
+        void removeCallback(Consumer<T> callback);
+    }
+
+    static class CellImpl<T> implements Cell<T> {
```
sanderploegsma, line 28: I'm wondering if there is a real need for this 'base' class. It seems to mix the concepts of the input and compute cells, and adds functionality to both subclasses that is only intended for a single subclass:

- The `ComputeCellImpl` class has a method `setValue` even though this should not be possible for a computed cell
- The `InputCellImpl` class has methods to add/remove callbacks and dependent cells, which doesn't make a lot of sense.

The only 'real' shared functionality of both types is already in the original interface: a getter for the cell value. IMO that's not even enough to warrant a base class at all.

exercises/practice/react/.meta/src/reference/java/React.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.stream.Collectors;
+
+public class React {
+
+    interface Cell<T> {
+        T getValue();
+        default T getTentativeValue() {
+            return getValue();
+        }
+        void addDependent(ComputeCell<T> dependent);
```
sanderploegsma, line 14: To be honest I don't think we should encourage students to add methods to an interface that are meant for internal use only. I would leave the interfaces as-is compared to the starter implementation.

sanderploegsma (changes requested):
That already looks a lot better!

I have one more comment about the exercise's configuration which should be addressed. The other comments on the reference implementation are a bit more subjective, so since you're the author of this exercise I leave it up to you to decide. In case you don't agree with my comments just let me know and resolve them.

exercises/practice/react/src/test/java/ReactTest.java
sanderploegsma, line 1: Looking at the CI build failures, it seems this file is missing an import for the `@Ignore` annotation.

exercises/practice/react/.meta/src/reference/java/React.java
kahgoh: I think this exercise can't be solved without adding some kind of internal method because of `testCallbacksShouldOnlyBeCalledOnceEvenIfMultipleDependenciesChange`. The test requires the `ComputeCell` to tell when a value change has finished propagating to know when it can finally fire the callbacks.

I think it is possible to solve this by adding the internal method to just `ComputeCellImpl`, but then that would require the implementing classes to hold a `Collection<ComputeCellImpl>` instead of `Collection<ComputeCell>` which doesn't feel right to me. I wonder if we should use just classes instead of interfaces which would allow the internal methods to be marked private or package private.

I took a look at other this is solved in other tracks (namely from the [Kotlin](https://github.com/exercism/kotlin/blob/main/exercises/practice/react/.meta/src/reference/kotlin/React.kt) and [Crystal](https://github.com/exercism/crystal/blob/main/exercises/practice/react/.meta/src/example.cr) tracks). I think their approach is a lot neater and will help avoid adding `getTentativeValue` on the base class/interface. I plan to try changing to this approach in the next commit.

exercises/practice/react/.meta/src/reference/java/React.java
kahgoh: There is more than `getValue()` that is shared here - there is also adding and tracking `ComputeCells` that depend on it.

config.json
kahgoh: With the latest commit, I ended up leaving this one for now as the interfaces were removed.

exercises/practice/react/.meta/src/reference/java/React.java
sanderploegsma: > I think it is possible to solve this by adding the internal method to just ComputeCellImpl, but then that would require the implementing classes to hold a `Collection<ComputeCellImpl>` instead of `Collection<ComputeCell>` which doesn't feel right to me.

That was exactly the direction I was going for. I'm considering the interfaces the 'public' API for the exercise, and the classes implementing them could even be private. I see no reason why these classes should not contain references to each other in private fields.

sanderploegsma (approved):

sanderploegsma:
Thanks! 👍

PR #2308, opened 2023-05-15
Wizards and warriors: more generic param names for #damagePoints

andeemarks:
Within the context of the exercise, using explicit names for the Fighter params to #damagePoints is correct, but it seems overly specific given the topic being introduced is inheritance and only knowing the argument is of type Fighter is sufficient.

# pull request

---

Reviewer Resources:


sanderploegsma:
Yes this bothered me too. In the context of the exercise the naming the parameter `target` might be even better!

SleeplessByte:
I like `target` too. Would you care to update this @andeemarks ?

sanderploegsma (approved):

PR #2310, opened 2023-05-18
remove parameter in method 'computeWinner'

FazarSN:
# pull request

the method in the test file is called without any parameter, this will adjust the main file.
I don't actually know if this is intentional or not, but I don't find something like this in another java exercises so far.

---

Reviewer Resources:


SleeplessByte (approved):

SleeplessByte:
Thanks. This was a remnant from a non-class based approach.

PR #2316, opened 2023-06-06
Update LasagnaTest.java

helixp90:
The expected outputs for:

* 4 layers, 8 minutes => 16
* 1 layer, 30 minutes => 32

are incorrect. It should be 40 & 12 respectively. My solution in Exercism worked for the other 8 cases instead of these two.

# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/lasagna/src/test/java/LasagnaTest.java
+++ b/exercises/concept/lasagna/src/test/java/LasagnaTest.java
@@ -80,11 +80,11 @@ public class LasagnaTest {
 
     @Test
     public void total_time_in_minutes_for_one_layer() {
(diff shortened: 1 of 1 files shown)


sanderploegsma:
They should be correct, the `totalTimeInMinutes()` method should return the number of minutes that the lasagna has been cooking, which is equal to the time it takes to prepare each layer plus the time it's been sitting in the oven. So bearing in mind that it takes 2 minutes to prepare each layer:

- 4 layers, 8 minutes in oven => 4 * 2 + 8 = 16
- 1 layer, 30 minutes in oven => 1 * 2 + 30 = 32

SleeplessByte:
Sander is correct. The tests are correct.

PR #2315, opened 2023-06-01
#2133: Added UnsupportedOperationException PigLatinTranslator

mnyon-grandkru:
Added a constructor and a `translate()` method that raise `UnsupportedOperationException`s, as mentioned in issue #2133.

Also added code to remove gradle warning as specified with this [Stack Overflow](https://stackoverflow.com/a/73904820/20832780) post.

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#event-checklist)

PR #2319, opened 2023-06-07
Exercise stubs for missing exercises

DWalz:
Create Exercis Stubs as mentioned in #2133 

Add Exercise Stubs for all exercises mentioned in [this comment](https://github.com/exercism/java/issues/2133#issuecomment-1566970411) except for `allergies` (#2303) and `pig-latin` (#2315) which are still pending.

~~I have used `final` for every function parameter.~~ **Function parameters are [not `final`](https://github.com/exercism/java/blob/main/POLICIES.md#avoid-using-final)**
All classes are `package-private`.

### DIFF
--- a/exercises/practice/knapsack/src/main/java/Item.java
+++ b/exercises/practice/knapsack/src/main/java/Item.java
@@ -1,0 +1,13 @@
+class Item {
+
+    // Weight of the item
+    int weight;
+    // Value of the item
+    int value;
+
+    Item(int itemWeight, int itemValue) {
+        this.weight = itemWeight;
+        this.value = itemValue;
+    }
+
+}
--- a/exercises/practice/queen-attack/src/main/java/Queen.java
+++ b/exercises/practice/queen-attack/src/main/java/Queen.java
@@ -1,6 +1,6 @@
 class Queen {
 
-    Queen(final int row, final int column) {
+    Queen(int row, int column) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
--- a/exercises/practice/queen-attack/src/main/java/Queen.java
+++ b/exercises/practice/queen-attack/src/main/java/Queen.java
@@ -1,0 +1,7 @@
+class Queen {
+
+    Queen(final int row, final int column) {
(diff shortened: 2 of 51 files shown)


SleeplessByte:
@DWalz 3 exercises are failing. Knapsack, Hangman, and Rest-Api. 

Can you check if this is "you" or something else? If it's something else we can merge this and fix it separately.

sanderploegsma:
> @DWalz 3 exercises are failing. Knapsack, Hangman, and Rest-Api.
> 
> Can you check if this is "you" or something else? If it's something else we can merge this and fix it separately.

I looked into the failures and two of them were caused by a Gradle configuration issue that had nothing to do with the individual exercise configs. I pushed a fix for that, because it's just made visible by the changes in this PR even though this issue was already present in `main`.

The Knapsack exercise still doesn't compile because it references a class `Item` that is not present in the stub. @DWalz if you could address this we can get this merged.

DWalz:
The `Item` class was indeed missing. I added it with the same implementation from the solution. I assume this class doesn't have to be implemented by the solver themselves.

exercises/practice/knapsack/src/main/java/Item.java
sanderploegsma, line 1: Any files you add should also be added in the exercise's `.meta/config.json` file, either in `files.solution` for files that should be edited as part of the exercise, or `files.editor` for read-only files that can be viewed by students in the online editor.

In this case since the class seems to be fully implemented already, please add it to the `files.editor` list.

exercises/practice/queen-attack/src/main/java/Queen.java
sanderploegsma, line 1: Any files you add should also be added in the exercise's `.meta/config.json` file, either in `files.solution` for files that should be edited as part of the exercise, or `files.editor` for read-only files that can be viewed by students in the online editor.

In this case since the class needs to be implemented by the student, please add it to the `files.solution` list.

sanderploegsma (changes requested):
Nice work! I found two more small things to take care of though.

sanderploegsma (approved):
LGTM! @SleeplessByte do you want to have a final look?

sanderploegsma:
I’m guessing that’s a no 😉

PR #2323, opened 2023-06-25
Fix typo in Introduction of Need for Speed

detinsley1s:
# pull request

The original spelling was the possessive form for more than one class, but since the Introduction details only a single class, the singular possessive form should be used.

---

Reviewer Resources:


SleeplessByte (approved):
I don't agree with the reasoning (the example has little to do with that sentence) but I do agree with the outcome as constructors _in general_ only instantiate a _single_ class, and only have one class attached, and thus it should have been singular.

PR #2333, opened 2023-08-11
Update Introduction

ChrolloRifat:
# pull request

I have updated some of the information to make it easier for beginners to understand.

---

Reviewer Resources:

### DIFF
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -24,7 +24,7 @@ class Calculator {
 }
 ```
 
-A function within a class is referred to as a _method_. Each _method_ can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from _methods_ using the `return` keyword. To allow a _method_ to be called by other classes, the `public` access modifier must be added.
+A function within a class is referred to as a _method_. Each method can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from methods using the `return` keyword. To allow a method to be called by other classes, the `public` access modifier must be added.
 
 ```java
 class Calculator {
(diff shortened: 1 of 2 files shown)


concepts/basics/introduction.md
```diff
@@ -1,30 +1,30 @@
...
 int count = 1; // Assign initial value
 count = 2;     // Update to new value
 
-// Compiler error when assigning different type
+// Compiler error when assigning a different type
 // count = false;
 ```
 
-Java is an [object-oriented language][object-oriented-programming] and requires all functions to be defined in a _class_. The `class` keyword is used to define a class.
+Java is an [object-oriented language][object-oriented programming] and requires all functions to be defined in a _class_. The `class` keyword is used to define a class.
```
sanderploegsma, line 19: This change should be reverted, as `[object-oriented-programming]` is a [reference-style link](https://www.markdownguide.org/basic-syntax/#reference-style-links),  and refers to the link defined at the bottom of the file.

concepts/basics/introduction.md
```diff
@@ -1,30 +1,30 @@
...
 
 ```java
 class Calculator {
     // ...
 }
 ```
 
-A function within a class is referred to as a _method_. Each method can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from functions using the `return` keyword. To allow a method to be called by other classes, the `public` access modifier must be added.
+A function within a class is referred to as a _method_. Each _method_ can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from _methods_ using the `return` keyword. To allow a _method_ to be called by other classes, the `public` access modifier must be added.
```
sanderploegsma, line 27: IMO there is no need to keep emphasizing the word "method" after it is introduced. It's common writing style to emphasize the first occurrence of a term only.

sanderploegsma (changes requested):
Nice improvements! I found two small things to address, if you like.

concepts/basics/introduction.md
ChrolloRifat: I think that's great 👍🏼

concepts/basics/introduction.md
ChrolloRifat: great 👍🏼

concepts/basics/introduction.md
sanderploegsma: OK so if you could push a change that reverts the change to the Markdown link I'll be happy to approve it for you!

ChrolloRifat: Okay. I have edited it. Let me know if everything good to go now.

sanderploegsma (approved):
LGTM, thanks for contributing!

Issue #2330, opened 2023-08-05
"Remote Control Competition" exercice: Incoherent task and test


armand-pierre:
In this exercise, the task 4 asks for implementing a static method sorting two different `ProductionRemoteControlCar` in arguments, but the test code pass in arguments a `List<ProductionRemoteControlCar>` object...

One could think of either modifying the source code or the task.

Issue #2318, opened 2023-06-07
remote-control-competition different values passed to getRankedCars


Willyanto39:
getRankedCars method in TestTrack takes 2 ProductionRemoteControlCar but RemoteControlCarTest calls getRankedCars method by passing a list of ProductionRemoteControlCar.

bljdavidson:
Is this a bug or are we supposed to be implementing the code for this test?

https://github.com/exercism/java/blob/main/exercises/concept/remote-control-competition/src/test/java/RemoteControlCarTest.java#L59-L91

Getting this:

>List<ProductionRemoteControlCar> rankings = TestTrack.getRankedCars(unsortedCars);
  required: ProductionRemoteControlCar,ProductionRemoteControlCar
  found:    List<ProductionRemoteControlCar>
  reason: actual and formal argument lists differ in length

>Required type: ProductionRemoteControlCar
Provided: List<ProductionRemoteControlCar>

SleeplessByte:
Fixed by #2337

PR #2337, opened 2023-08-20
Change TestTrack.getRankedCars stub to take a list of cars, consistent with the unit tests

IsaacG:
---

Reviewer Resources:


sanderploegsma:
- Not sure why CI isn't picking this up, but the exemplar implementation in `.meta/src/reference/java/TestTrack.java` also contains the invalid method signature and should therefore be updated as well
- The code snippet in the instructions should probably be updated to match the new signature by passing it `unsortedCars`

iHiD (approved):
Blind approving for you :)

IsaacG:
Thanks! I updated additional sites called out by @sanderploegsma

sanderploegsma:
This also fixes #2330 and #2318 by the way

PR #2331, opened 2023-08-08
Fix Remote Control Competition TestTrack parameters

jneurock:
# pull request

This PR addresses an issue in the Remote Control Competition concept exercise where the `TestTrack.getRankedCars` method starts out with a different set of parameters than the tests pass in. It does so by making the initial method signature match what the tests expect.

Fixes #2330 

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#event-checklist)

PR #2283, opened 2023-03-14
Improve instructions and tests for Remote Control Competition exercise

eneuschild:
# Improve instructions and tests for Remote Control Competition exercise

Issue #2164 points out that Remote Control Competition's instructions don't match its tests very well. This PR is an attempt to improve this.

Completing each instruction subsection now increases the number of passing tests, which wasn't given previously.

## Open Questions
- I have not yet touched the tests themselves, as I feel this would be too big a change for a single (first) PR. Would you agree?
  - **Update**: Since @SleeplessByte disagreed with this assessment, I have now introduced more tests that gradually guide the student to a complete solution. 
- The way I understand the [Updating the README](https://github.com/exercism/java/blob/6ef04c4cd2ab62bca7f8d821ff18c5e81390eedd/CONTRIBUTING.md#updating-the-readmes) section of the CONTRIBUTING.md, I should be running `configlet` to maintain README files. Am I understanding correctly that this isn't really applicable here because this exercise is not part of the more generic problem specifications? :thinking: 

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#event-checklist)

PR #2338, opened 2023-08-22
Fix compile error in Remote Control Competition

### LINKED ISSUE #2330: "Remote Control Competition" exercice: Incoherent task and test

sanderploegsma:
The default implementation of the `TestTrack` class in Remote Control Competition doesn't match the expected signature in the unit tests, causing the compilation to fail until the student updates the dummy method to match the correct signature.

It looks like this was introduced in #2283 where the unit tests were updated but the stub implementation wasn't.

Note that this is a duplicate of #2331, #2336 and #2337, but the former updates the assertions in a way that doesn't make a lot of sense, and the latter two update the stub but not the exemplar implementation nor the instructions.

Fixes #2330, #2318

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/remote-control-competition/.docs/instructions.md
+++ b/exercises/concept/remote-control-competition/.docs/instructions.md
@@ -59,7 +59,7 @@ prc2.setNumberOfVictories(2);
 List<ProductionRemoteControlCar> unsortedCars = new ArrayList<>();
 unsortedCars.add(prc1);
 unsortedCars.add(prc2);
-List<ProductionRemoteControlCar> rankings = TestTrack.getRankedCars(prc1, prc2);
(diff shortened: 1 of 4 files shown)


sanderploegsma:
And a side note: I don't get why this wasn't caught in CI? AFAIK the CI test job runs the unit tests on the examplar implementations, right? This should also have resulted in a compilation error, so I suspect the CI job is not 100% doing what it's supposed to do...

sanderploegsma:
Looks like @IsaacG beat me to it, so closing as duplicate of #2337

PR #2339, opened 2023-08-22
Fix CI silently not running exercise tests

sanderploegsma:
I was investigating why the bug in #2330 didn't cause a failure in CI, and apparently the exercises aren't tested at all due to an error in the test script!

```
>>> solve_all_exercises(exercism_exercises_dir="/home/runner/work/java/java/build/exercism")
+ solve_all_exercises /home/runner/work/java/java/build/exercism
+ local exercism_exercises_dir=/home/runner/work/java/java/build/exercism
+ echo '>>> solve_all_exercises(exercism_exercises_dir="/home/runner/work/java/java/build/exercism")'
++ pwd
+ local track_root=/home/runner/work/java/java
++ jq '.exercises[].slug + " "' --join-output
++ cat config.json
jq: error (at <stdin>:2069): Cannot index array with string "slug"
```

It appears there are multiple things wrong with this script:
- It ignores errors from commands in a pipe, causing it to swallow the failure
- Apparently the `bin/journey-test.sh` script wasn't updated to parse `config.json` correctly after the addition of concept exercises, so none of the tests even ran

After some fiddling with the script I noticed that the `exercises/build.gradle` script is configured to do exactly what is required in CI: it pre-processes the test files to remove any `@Ignore` tags and then runs the tests against the exemplar/example implementation. So I figured the easiest way to fix the CI pipeline is to just run `gradle test`.

Note that because the CI pipeline has probably been broken for a while, there were a few exercises of which the example implementation doesn't pass their own test cases. I fixed the easy ones, but the following practice exercises seem to require a bit more attention to get their tests passing:

- Pythagorean Triplet
- SGF Parsing

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -7,33 +7,31 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
-    - name: Set up JDK 1.17
-      uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2
-      with:
-        java-version: 17
-        distribution: 'temurin'
-    - name: Grant execute permission for gradlew
-      run: chmod +x gradlew
-    - name: Compile and checkstyle with Gradle
-      run: cd exercises && ../gradlew check compileStarterSourceJava --parallel --continue --info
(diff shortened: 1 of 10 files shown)


ErikSchierboom:
@sanderploegsma Could you look into the CI failure please?

sanderploegsma:
@ErikSchierboom all done, they should pass now: https://github.com/sanderploegsma/exercism-track-java/actions/runs/5962186142

ErikSchierboom (approved):
Great!

PR #2341, opened 2023-08-24
Update phone number exercise

djspinmonkey:
# pull request

The main change here is to rename the `.getNumber()` method implied by the tests to be called `.toString()` instead.

A phone number either warrants its own type or it does not. If it does, then what this method is returning is the canonical string representation of a given phone number, and in Java we call methods that do that `.toString()`.

(If a phone number does _not_ warrant its own type, then this exercise should be significantly restructured to use a `PhoneNumberFormatter` class that works with Strings, or some such, but that doesn't really seem necessary and would be a much larger change.)

The other minor changes are to clean up the formatting of the description a bit, and remove a reference to the exercise being "difficulty 5", which to the best of my knowledge does not mean anything in the context of Exercism.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/phone-number/.docs/instructions.append.md
+++ b/exercises/practice/phone-number/.docs/instructions.append.md
@@ -1,9 +1,8 @@
 # Instructions append
 
-Since this exercise has difficulty 5 it doesn't come with any starter implementation.
-This is so that you get to practice creating classes and methods which is an important part of programming in Java.
-It does mean that when you first try to run the tests, they won't compile.
-They will give you an error similar to:
+This exercise doesn't come with any starter implementation.  This is so that you get to practice creating classes and
+methods, which is an important part of programming in Java.  That means that when you first try to run the tests, they
+won't compile.  They will give you an error similar to:
 ```
  path-to-exercism-dir\exercism\java\name-of-exercise\src\test\java\ExerciseClassNameTest.java:14: error: cannot find symbol
         ExerciseClassName exerciseClassName = new ExerciseClassName();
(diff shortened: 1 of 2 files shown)


exercises/practice/phone-number/.docs/instructions.append.md
```diff
@@ -1,22 +1,20 @@
 # Instructions append
 
-Since this exercise has difficulty 5 it doesn't come with any starter implementation.
-This is so that you get to practice creating classes and methods which is an important part of programming in Java.
-It does mean that when you first try to run the tests, they won't compile.
-They will give you an error similar to:
+This exercise doesn't come with any starter implementation.  This is so that you get to practice creating classes and
+methods, which is an important part of programming in Java.  That means that when you first try to run the tests, they
+won't compile.  They will give you an error similar to:
```
SleeplessByte, line 5: We adhere to the [Markdown style](https://exercism.org/docs/building/markdown/markdown) given by Exercism, which is one sentence per line. Can you revert these changes in this file?

exercises/practice/phone-number/src/test/java/PhoneNumberTest.java
```diff
@@ -10,7 +10,7 @@ public class PhoneNumberTest {
     @Test
     public void cleansTheNumber() {
         String expectedNumber = "2234567890";
-        String actualNumber = new PhoneNumber("(223) 456-7890").getNumber();
+        String actualNumber = new PhoneNumber("(223) 456-7890").toString();
```
SleeplessByte, line 13: If the tests are changed, the _proof_ or _example_ implementation in the `phone-number` folder also needs to be updated, otherwise the CI will fail to run.

sanderploegsma:
Closing this because of inactivity. @djspinmonkey let us know if you still want to finish this, we’ll reopen this PR.

PR #2347, opened 2023-08-29
Compile starter source with tests

sanderploegsma:
This updates the Gradle build configuration to enable us to check whether the starter sources are able to compile with the unit tests. Hopefully this can help students by making sure that the code they receive for each exercise is able to compile without errors from the start, as figuring out why the tests don't compile is out of scope wrt learning on Exercism IMO.

Note that this PR is depending on each exercise having valid stubs, which are being added as part of #2133.

---

Reviewer Resources:


SleeplessByte (approved):
I agree with this change.

PR #2351, opened 2023-08-31
Remove '(static)' in default exception messages for non-static method…

joe-mccann-dev:
…s, as represented in ElonsToyCarTest.java.

# pull request

This pull request revises the stub's exception messages to match behavior in the test file. Three methods were referred to as "static" but are instance methods in the test file.

---

Reviewer Resources:


sanderploegsma (approved):
LGTM, thanks 👍

PR #2353, opened 2023-09-13
Update Rational.java

spinscale:
Rational.hashCode() accidentally used the same field twice instead of numerator and denumerator.

---

Reviewer Resources:


sanderploegsma (approved):
Thanks for your contribution! 👍

PR #2364, opened 2023-09-20
Use templates to generate concept exercise introductions

sanderploegsma:
This updates all concept exercises to use an `introduction.md.tpl` template file (as described in the [docs](https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-introduction-md-tpl)), since the introduction for each concept exercise typically contains an introduction to the concept it's trying to teach.

This PR is a good showcase of why this change is necessary: the diff of each `introduction.md` in _both the exercises and the concepts themselves_ shows that it's very easy for the two to diverge over time. Of course there might be reasons for concept exercise introductions to not include the concept's introduction verbatim, but judging by the current state at least it's a sane default. Plus, this makes diverging from re-using the concept introduction explicit per concept exercise.

One thing I noticed is how the `conditionals-if` concept is teaching more than it's supposed to: it also contains an introduction about boolean logic (which is already covered in `booleans`) and about switch statements (which is covered in a later concept `switch-statements`). The Blackjack exercise (teaching the `conditionals-if` concept) is also mixing if/else and switch statements. We should probably open a separate issue to tackle this.

---

Reviewer Resources:

### DIFF
--- a/concepts/chars/introduction.md
+++ b/concepts/chars/introduction.md
@@ -4,9 +4,9 @@ The Java `char` type represents the smallest addressable components of text.
 Multiple `char`s can comprise a string such as `"word"` or `char`s can be
 processed independently. Their literals have single quotes e.g. `'A'`.
 
-Java `char`s support Unicode encoding, so in addition to the Latin character set
-most modern writing systems  can be represented,
-e.g.  Greek `'β'`. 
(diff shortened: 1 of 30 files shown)


ErikSchierboom (approved):

sanderploegsma:
I opened #2365 to address the issues mentioned in the PR description regarding the `conditionals-if` concept.

PR #2425, opened 2023-09-22
Sync docs for practice exercise transpose

IsaacG:
# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/transpose/.docs/instructions.md
+++ b/exercises/practice/transpose/.docs/instructions.md
@@ -17,7 +17,8 @@ BE
 CF
 ```
 
-Rows become columns and columns become rows. See <https://en.wikipedia.org/wiki/Transpose>.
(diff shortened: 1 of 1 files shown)


exercises/practice/transpose/.docs/instructions.md
```diff
@@ -17,7 +17,8 @@ BE
 CF
 ```
 
-Rows become columns and columns become rows. See <https://en.wikipedia.org/wiki/Transpose>.
+Rows become columns and columns become rows.
+See [transpose][].
```
sanderploegsma, line 21: Interesting change... I know it's possible to omit the link text and just show the reference, like `[transpose]`, but I didn't know this was possible. Or is it not and is this a bug? Anyway, let's merge and see.

sanderploegsma (approved):

Issue #2356, opened 2023-09-19
Update concept exercise tests with task identifiers


sanderploegsma:
## Introduction

Exercism's Java test runner recently got updated to support more features of the online editor. One of these features is the ability to show the information of each exercise task in the test result output ([reference](https://exercism.org/docs/building/tooling/test-runners/interface#h-task-id)). The `lasagna` concept exercise has been updated already to make use of these features, and the following screenshot shows the result:

<img width="555" alt="Screenshot 2023-09-19 at 08 23 01" src="https://github.com/exercism/java/assets/1564254/8ca5a60f-5252-4e8f-b95a-1be326375af5">

The goal of this issue is to convert the other concept exercise tests as well.

## Requirements

For each concept exercise:

- Make sure that each test in the test suite is linked to exactly one task in the exercise instructions. Do this by annotating the test with `@Tag("task:X")`, where `X` is the number of that task.
- Add the `@DisplayName` annotation to each test case and provide a human-readable and meaningful name for the test.

This requires the concept exercise to migrate from JUnit 4 to JUnit 5, since the `@Tag` and `@DisplayName` annotations are not supported by JUnit 4.

For reference, [check out the PR updating the `lasagna` concept exercise](https://github.com/exercism/java/pull/2354), which was updated to serve as an example.

## Exercises to be updated

- [x] `annalyns-infiltration`
- [x] `bird-watcher`
- [x] `blackjack`
- [x] `calculator-conundrum`
- [x] `cars-assemble`
- [x] `elons-toy-car`
- [x] `football-match-reports`
- [x] `karls-languages`
- [x] `lasagna`
- [x] `log-levels`
- [x] `need-for-speed`
- [x] `remote-control-competition`
- [x] `salary-calculator`
- [x] `squeaky-clean`
- [x] `wizards-and-warriors`

## Contributing guidelines

Before opening a Pull Request, make sure to check out the contributing guidelines in [CONTRIBUTING.md](https://github.com/exercism/java/blob/main/CONTRIBUTING.md).

When opening a Pull Request for this issue, make sure to reference this issue in its description so that we can keep track of the status.

In case you have any further questions regarding this issue, feel free to ask here.

aldotele:
Hi there, I would like to take care of it! Thanks

sanderploegsma:
Great! I’ll assign it to you.

Feel free to open multiple PRs to tackle the exercises one by one, that also makes it easier to review.

sanderploegsma:
[...]

PR #2428, opened 2023-09-23
Update cars assemble to use test runner v3 features

aldotele:
This PR addresses the cars-assemble task on issue #2356 

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
+++ b/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
@@ -18,84 +18,84 @@ public class CarsAssembleTest {
 
     @Test
     @Tag("task:1")
-    @DisplayName("The productionRatePerHour method returns the correct result given line's speed")
+    @DisplayName("The productionRatePerHour method returns the correct result when line's speed is 0")
     public void productionRatePerHourForSpeedZero() {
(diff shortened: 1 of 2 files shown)


exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
```diff
@@ -9,67 +11,91 @@ public class CarsAssembleTest {
...
 
-    @Before
+    @BeforeEach
     public void setUp() {
         carsAssemble = new CarsAssemble();
     }
 
     @Test
+    @Tag("task:1")
+    @DisplayName("The productionRatePerHour method returns the correct result given line's speed")
```
sanderploegsma, line 21: The test runner requires that each test has a unique name. Plus, for students it’s also more meaningful if each test is clearly named.

I would suggest naming these methods “The productionRatePerHour method returns the correct result when the line’s speed is 0/1/etc”. Or maybe even “The productionRatePerHour should be X for line speed Y”.

sanderploegsma (changes requested):

exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
aldotele: Very clear, I will keep it in mind for next PRs, thanks!

sanderploegsma (approved):

PR #2492, opened 2023-09-25
Update salary calculator to use test runner v3 features

aldotele:
This PR addresses the salary calculator rask on issue #2356 

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/salary-calculator/build.gradle
+++ b/exercises/concept/salary-calculator/build.gradle
@@ -1,23 +1,22 @@
-apply plugin: "java"
-apply plugin: "eclipse"
-apply plugin: "idea"
-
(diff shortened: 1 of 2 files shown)


sanderploegsma (approved):
Hmm, this makes it painfully explicit that there are no unit tests for tasks 1 and 2, which is a bit of an annoying experience for the student. Ideally there should be some tests that guide students through each task in the exercise, to let them know that they are on the right track.

However, this is not something to fix here, I'll create a separate issue for it.

aldotele:
Yes, I noticed the lack of tests for task 1 and 2 methods but, as you mentioned, I thought it would have been addressed in a separate issue. Thanks for your review!

PR #2500, opened 2023-09-27
Update need for speed to use test runner v3 features

aldotele:
This PR addresses the need for speed task on issue #2356 

---

Reviewer Resources:


aldotele:
**Note**:
I split the _nitro_has_correct_battery_drain_ test in two tests as it was testing two different things.

Now the _nitro_has_correct_battery_drain_ is just testing that the battery is not fully drained after driving 24 times, while the _nitro_battery_completely_drains_ is testing that the battery completely drains after driving 25 times.

sanderploegsma (approved):
:shipit:

PR #2501, opened 2023-09-28
Changed method names to make more idiomatic in complex-numbers (#2317)

### LINKED ISSUE #2317: complex-numbers: naming inconsistencies
I am currently creating method stubs for Exercises and when working on the `compley-numbers` exercise I've noticed an inconsistency in the naming of some methods: The methods are sometimes named after the mathematical operators and sometimes after established function names in programming.

An example of that are the Operators `+` and `-`. `+` is named `add` while `-` is named `minus`.

With the method stubs it will be clear which method is responsible for which action but I still think that the naming convention should be unified for increased clarity.

Personally I like the usage of already established names, so I would propose the following name refactorings:

- `+`: `add` -> `add` (*no change*)
- `-`: `minus` -> `sub`
- `*`: `times` -> `mult` / `mul`
- `/`: `div` -> `div` (*no change*)
- `e^x`: `exponentialOf` -> `exp`

Wasy18:
# pull request

Made small changes to make methods in exercise more idiomatic and for better readability

Fixes #2317 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/complex-numbers/.meta/config.json
+++ b/exercises/practice/complex-numbers/.meta/config.json
@@ -13,7 +13,8 @@
     "muzimuzhi",
     "SleeplessByte",
     "Smarticles101",
-    "sshine"
+    "sshine",
+    "Wasy18"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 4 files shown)


sanderploegsma (approved):
LGTM, thanks!

PR #2502, opened 2023-09-28
Switch to assertj for exceptions

keszocze:
# pull request

Switching to AssertJ for the
* Forth
* Ledger
* OCR
* Palindrom Calculator
* Phone Number
* SGF Parsing
* Simple Linked List
* Triangle
exercises.

This request brings #2147 closer to being fully solved.

---

Reviewer Resources:


sanderploegsma:
It looks like this includes updates to `error-handling`, which was already merged. Could you rebase your branch on `exercism:main`?

keszocze:
The change is very tiny in the sense that I only make the imports explicit. I *probably* can rebase it but have to look up what that is and how it works. I am wondering whether that really is worth the hassle, though.

sanderploegsma:
Ah well the reason I mentioned it is because there are conflicts with the `main` branch.

If rebasing is new for you it’s also fine if you just merge the `main` branch into yours and resolve the conflicts that way 👍

sanderploegsma:
@keszocze thanks for picking up all of these! I'd like to ask you to make additional updates in a different PR: updates to exercise tests trigger new test runs of submitted solutions on the Exercism infrastructure, so we try not to update too many at once at risk of overloading the servers.

keszocze:
Sure can do that. Then I stop adding more to this PR. As all changes belong to a single issue, I thought that this was fine. I sure do not want to make your life unnecessarily complicated ;)

exercises/practice/custom-set/src/test/java/CustomSetTest.java
```diff
@@ -187,24 +187,26 @@ public void addToEmptySet() {
 
         actual.add(element);
 
-        assertNotNull(actual);
-        assertFalse(actual.isEmpty());
-        assertTrue(expected.equals(actual));
+        assertThat(actual).isNotNull().returns(true, s -> s.equals(expected));
```
sanderploegsma, line 190: I'm not sure this is an improvement over just using separate assertions:

```java
assertThat(actual).isNotNull();
assertThat(actual.isEmpty()).isFalse();
assertThat(actual.equals(expected)).isTrue();
```

The main reason for saying this is that students will have a look at the test code when some tests are failing to help them out, and they shouldn't have to spend a lot of extra effort understanding what the tests are doing. As far as test code goes, the simpler the better.

This comment addresses all uses of `assertThat(...).returns()` in the tests, to save on the amount of review comments in this PR :)

exercises/practice/custom-set/src/test/java/CustomSetTest.java
```diff
@@ -187,24 +187,26 @@ public void addToEmptySet() {
...
     @Test
     public void addToNonEmptySet() {
         char element = '3';
         CustomSet<Character> expected = new CustomSet<>(Collections.unmodifiableList(
-            Arrays.asList('1', '2', '3', '4')));
+                Arrays.asList('1', '2', '3', '4')));
         CustomSet<Character> actual = new CustomSet<>(Arrays.asList('1', '2', '4'));
 
         actual.add(element);
+        System.out.println(actual);
```
sanderploegsma, line 204: Remove this print statement please.

exercises/practice/ledger/src/test/java/LedgerTest.java
```diff
@@ -227,13 +226,10 @@ public void givenInvalidLocale_ThenException() {
         var entries = new Ledger.LedgerEntry[0];
         var locale = "it-IT";
 
-        // when
-        IllegalArgumentException illegalArgumentException = assertThrows(
-            IllegalArgumentException.class,
-            () -> ledger.format(EUR_CURRENCY, locale, entries));
-
         // then
```
sanderploegsma, line 229: You can remove the `// given` and `// then` comments since they don't make a lot of sense anymore with this style.

exercises/practice/sgf-parsing/src/test/java/SgfParsingTest.java
```diff
@@ -12,27 +12,23 @@ public class SgfParsingTest {
     @Test
     public void emptyInput() {
         String input = "";
-        assertThrows("tree missing",
-            SgfParsingException.class,
-            () -> new SgfParsing().parse(input));
+        assertThatExceptionOfType(SgfParsingException.class).isThrownBy(() -> new SgfParsing().parse(input));
```
sanderploegsma, line 15:
```suggestion
        assertThatExceptionOfType(SgfParsingException.class).as("tree missing")
                .isThrownBy(() -> new SgfParsing().parse(input));
```

exercises/practice/sgf-parsing/src/test/java/SgfParsingTest.java
```diff
@@ -12,27 +12,23 @@ public class SgfParsingTest {
...
 
     @Test
     @Ignore("Remove to run test")
     public void treeWithNoNodes() {
         String input = "()";
-        assertThrows("tree with no nodes",
-            SgfParsingException.class,
-            () -> new SgfParsing().parse(input));
+        assertThatExceptionOfType(SgfParsingException.class).as("tree missing")
+                .isThrownBy(() -> new SgfParsing().parse(input));
```
sanderploegsma, line 23:
```suggestion
        assertThatExceptionOfType(SgfParsingException.class)
                .as("tree with no nodes")
                .isThrownBy(() -> new SgfParsing().parse(input));
```

sanderploegsma (changes requested):
Nice work so far! I have left a couple of comments for you to address before we can merge this.

exercises/practice/custom-set/src/test/java/CustomSetTest.java
keszocze: I can do so but isn't the whole point of using AssertJ that it has a better API for doing this? Shouldn't putting each new test on a new line already address the issue you just raised?

I do agree, though, that the `returns` call is not necessarily the most beautiful code.

sanderploegsma: Ah yes I didn't express myself clearly I see. I don't have anything against chaining AssertJ assertions per se. For instance, this is totally clear:

```java
assertThat(someList)
    .isNotNull()
    .containsExactly(1, 2, 3, 4);
```

My problem in this case is with the `.returns` method. The syntax is a bit cumbersome, and probably requires students to look up what it does, which makes the tests more complex than needed. The intention of my comment therefore was to replace the uses of `.returns` with separate assertions.

keszocze: Ah okay. Yes, I'll remove that then.

sanderploegsma (approved):
All good, thanks for contributing!

PR #2521, opened 2023-10-17
Update CarsAssembleTest.java

pdmoore:
Tests for Task 1 used a boolean comparison to determine whether tests pass. Replaced those with more idiomatic assertEquals(double, double, epsilon) form. 

When the new style of test fails the user is shown the expected value and the actual value: 
Expected :1591.2
Actual   :1989.0
The previous implementation showed the user the  less clear: 
Expected :true
Actual   :false

The test implementation being replaced required the comparison to be strictly less than epsilon, not inclusive of epsilon. The JUnit assertEquals replacement check is inclusive of epsilon. OLD: Math.abs(value1 - value2) < epsilon)
NEW: Math.abs(value1 - value2) <= epsilon

Given an epsilon of 0.0000001d this is unlikely to see any existing solutions fail given the expected answers only check the tenths place.

# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
+++ b/exercises/concept/cars-assemble/src/test/java/CarsAssembleTest.java
@@ -4,6 +4,7 @@ import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.*;
 
 
 public class CarsAssembleTest {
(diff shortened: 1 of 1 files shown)


sanderploegsma (changes requested):
Thanks for your contribution! This would be an improvement indeed, much clearer for students when a test fails.

However, could you update the assertions to use AssertJ instead of JUnit according to our [Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#prefer-assertj-assertions)?

sanderploegsma (approved):
LGTM, thanks for your contribution!

PR #2523, opened 2023-10-18
[simple-cipher] Fix outdated reference to SimpleCipherStepThreeTest

apetresc:
# pull request

This test file was removed years ago in #1639, but the instructions still mention it (which caused me, at least, quite a bit of confusion since I thought I had deleted something erroneously).

This fixes the documentation to remove the reference to Step Three. If the intent was to never have deleted SimpleCipherStepThreeTest.java in the first place, let me know and I'll amend the PR to that.

---

Reviewer Resources:


exercises/practice/simple-cipher/.docs/instructions.append.md
```diff
@@ -1,8 +1,8 @@
 # Instructions append
 
-This exercise has three test classes: SimpleCipherStepOneTest.java, SimpleCipherStepTwoTest.java and SimpleCipherStepThreeTest.java.
+This exercise has two test classes: SimpleCipherStepOneTest.java and SimpleCipherStepTwoTest.java.
```
sanderploegsma, line 3: I just checked and apparently the second one is called `SimpleCipherStepTwoSubstitutionTest.java` now. Could you fix that as well?

sanderploegsma (commented):
Thanks for fixing this! One comment to take care of if you would.

exercises/practice/simple-cipher/.docs/instructions.append.md
apetresc: Good catch! Although the problem goes even deeper 😆  The test file probably _should_ be called `SimpleCipherStepTwoSubstitutionTest.java` but [in actuality it's called](https://github.com/exercism/java/blob/main/exercises/practice/simple-cipher/src/test/java/SimpleCipherStepTwoSubsitutionTest.java) `SimpleCipherStepTwoSubsitutionTest.java` (note the missing 't' in 'Substitution').

Should I fix the incorrect class name as well? Or will that break people's automations and submissions?

sanderploegsma: Ugh good call, I missed that...

The consequences of renaming the file are:
- Existing submissions will be retested but that should not break anything as the functionality remains the same.
- The only students that are affected are the ones who use the Exercism CLI to redownload the updated files using `exercism download --track=java --exercise=simple-cipher --force`, as it won't remove the old file, so they end up with both `SimpleCipherStepTwoSubstitutionTest.java` and `SimpleCipherStepTwoSubsitutionTest.java`.

Looking at the [canonical data for this exercise](https://github.com/exercism/problem-specifications/blob/main/exercises/simple-cipher/canonical-data.json) I see why the tests were broken up into multiple files in the first case; there are tests that verify the cipher when a random key is used vs when a static key is used. However, the names of the files are misleading here, they seem to indicate that they are testing the functionality from the steps defined in the instructions while they are not.

The original purpose of having these separate files as explained in this file seems to be able to run the tests for each step individually, so that a student can check if they solved each step before moving on to the next one. However, since this is not the case, I actually doubt whether there is a benefit to having multiple files at all. So if we are renaming files anyway, I would suggest to merge all test cases into a single file and removing the `instructions.append.md` altogether.

apetresc: I agree; for what it's worth, as someone who just finished this exercise as a student, the three-step split was very confusing for exactly the reasons you described:

a) It makes it seem like it matches the progression described in the problem, when it doesn't
b) The logical dependency is backwards; if you just implement things normally you'll finish StepTwo before you finish StepOne. StepOne is just the same as StepTwo except you _also_ have to generate a random key. Getting encode()/decode() to work isn't any easier when working with a random key instead of a fixed key, if anything it's harder.
c) StepThree was missing

Given that, I agree things would be a lot clearer with a single test file.

Should I go ahead and make that change in this PR?

sanderploegsma: Yes please! If you want to keep it in line with the canonical data it would be fine if you use nested test classes to separate the two groups of tests, like so:

```java
@RunWith(Suite.class)
@Suite.SuiteClasses({SimpleCipherTest.RandomKeyCipher.class, SimpleCipherTest.SubstitutionCipher.class})
public class SimpleCipherTest {

    public static class RandomKeyCipher {
        @Test
        public void canEncode() {
            // ...
        }

        // ...
    }

    public static class SubstitutionCipher {
        @Test
        public void canEncode() {
            // ...
        }

        // ...
    }
}
```

This way it's clear how each test maps to the canonical data. This is not required however, so I'll leave that up to you.

apetresc: Cool, done (I think), and works on my machine:

```
$ gradle test

> Task :practice:simple-cipher:compileTestJava
  (source = /Users/apetresc/tmp/java/exercises/practice/simple-cipher/build/gen/test/java/SimpleCipherTest.java)

> Task :practice:simple-cipher:test

SimpleCipherTest$RandomKeyCipher > keyIsLowercaseLetters PASSED

SimpleCipherTest$RandomKeyCipher > cipherCanDecode PASSED

SimpleCipherTest$RandomKeyCipher > cipherCanEncode PASSED

SimpleCipherTest$RandomKeyCipher > cipherIsReversible PASSED

SimpleCipherTest$SubstitutionCipher > cipherCanDoubleShiftEncode PASSED

SimpleCipherTest$SubstitutionCipher > cipherCanDecode PASSED

SimpleCipherTest$SubstitutionCipher > cipherCanEncode PASSED

SimpleCipherTest$SubstitutionCipher > cipherMessageLongerThanKey PASSED

SimpleCipherTest$SubstitutionCipher > cipherIsReversibleGivenKey PASSED

SimpleCipherTest$SubstitutionCipher > cipherCanWrapDecode PASSED

SimpleCipherTest$SubstitutionCipher > cipherCanWrapEncode PASSED

SimpleCipherTest > SimpleCipherTest$RandomKeyCipher.keyIsLowercaseLetters PASSED

SimpleCipherTest > SimpleCipherTest$RandomKeyCipher.cipherCanDecode PASSED

SimpleCipherTest > SimpleCipherTest$RandomKeyCipher.cipherCanEncode PASSED

SimpleCipherTest > SimpleCipherTest$RandomKeyCipher.cipherIsReversible PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherCanDoubleShiftEncode PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherCanDecode PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherCanEncode PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherMessageLongerThanKey PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherIsReversibleGivenKey PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherCanWrapDecode PASSED

SimpleCipherTest > SimpleCipherTest$SubstitutionCipher.cipherCanWrapEncode PASSED

BUILD SUCCESSFUL in 1s
4 actionable tasks: 3 executed, 1 up-to-date
```

Not sure if I need to explicitly do something to re-generate all the generated files or if that happens automatically with a Github Action.

sanderploegsma: You can run all tasks locally that are run in GitHub Actions workflow, and Checkstyle is run as part of the `check` Gradle task. So running `gradle :practice:simple-cipher:check` should do it.

I also noticed that the Configlet workflow failed, this is because the `exercises/practice/simple-cipher/.meta/config.json` file needs to contain references to all source files belonging to the exercise. Since we made changes to the test files, we need to update the `.meta/config.json` accordingly.

apetresc:
@sanderploegsma I'm worried this might've screwed something up 😅 Since this change went live and I updated my problem on Exercism and resubmitted, the new test files pass locally without a problem but even after 3 submissions spread days apart, the test runner on the site gives me:

> An error occurred while running your tests. This might mean that there was an issue in our infrastructure, or it might mean that you have something in your code that's causing our systems to break.
>
> Please check your code, and if nothing seems to be wrong, try running the tests again.

I don't _think_ there's anything wrong with my submission so I assume there's some metadata file somewhere we forgot to update to use the new test classes?

In general I have no idea how to diagnose issues with the system test runners; I don't even know where those are hosted or if I'm supposed to have access. Can you give me a tip on how to start investigating this?

sanderploegsma:
Hmm that is unfortunate. If the tests run fine locally then perhaps the test runner is having trouble, I suspect with the `@Enclosed` annotation. 

The Java test runner is open source as well, you can find it here: https://github.com/exercism/java-test-runner. You should be able to run it locally and point it to your local folder containing the exercise files. Maybe that will help understand what is happening behind the scenes.

PR #2541, opened 2023-10-28
Remove duplicate Gradle workflows being triggered

sanderploegsma:
Currently the Gradle CI action is run twice for every push to every PR, one for the push event and another for the PR event. This changes the workflow to only run for push events to the `main` branch.

---

Reviewer Resources:


ErikSchierboom (approved):
Nice!

PR #2543, opened 2023-10-28
Update installation instructions to install Java 21 (#2368)

### LINKED ISSUE #2368: Update installation instructions to install latest Java LTS version
Currently the [installation instructions shown to users](https://exercism.org/docs/tracks/java/installation) tell them to install Java 11. However, active support for Java 11 is going to end soon ([source](https://endoflife.date/java)), so we should probably update the instructions to point to the latest LTS release instead. At the time of writing this is Java 21.

danilopiazza:
# pull request

I updated the installation instructions to install the latest Java LTS version from OpenJDK.

For the most part, I replaces references to AdoptOpenJDK with Adoptium and/or Eclipse Temurin.

Fixes #2368 

---

Reviewer Resources:

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -31,7 +31,7 @@ Open an administrative command prompt.  (If you need assistance opening an admin
 2.  Install the JDK:
 
      ```batchfile
-     C:\Windows\system32> choco install openjdk11
+     C:\Windows\system32> choco install openjdk
      ...
      C:\Windows\system32> refreshenv
      ...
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -31,7 +31,7 @@ Open an administrative command prompt.  (If you need assistance opening an admin
 2.  Install the JDK:
 
      ```batchfile
-     C:\Windows\system32> choco install openjdk11
+     C:\Windows\system32> choco install openjdk
```
sanderploegsma, line 34: Can we replace this with `choco install openjdk21`? We try to make sure that students are using a version of Java that is also supported by our Java test runner, so I'd prefer it if the instructions are written in such a way that students will install a specific version of Java, instead of the latest one.

docs/INSTALLATION.md
```diff
@@ -63,19 +63,13 @@ Below are instructions for install using the most common method - using Homebrew
...
-     ```
-     $ brew tap adoptopenjdk/openjdk
-     ```
-
-3.  Install the JDK:
+2.  Install the JDK:
 
      ```
-     $ brew install --cask adoptopenjdk11
+     $ brew install --cask temurin
```
sanderploegsma, line 69: Same question here, can we make sure students install Java 21 specifically, instead of the latest version? According to [the adoptium installation instructions](https://adoptium.net/installation/) this should be the way to do so:

```sh
$ brew tap homebrew/cask-version
$ brew install --cask temurin21
```

sanderploegsma (changes requested):
Thanks for picking this up! I left two comments that I think should be resolved before we can merge this.

docs/INSTALLATION.md
danilopiazza: Thanks! Sorry for missing that.

docs/INSTALLATION.md
danilopiazza: Unfortunately, there is no `openjdk21` Chocolatey package. I switched to the `temurin21` package.

sanderploegsma (approved):
LGTM, thanks! 👍

Issue #2539, opened 2023-10-28
Remove nested test classes from practice exercise `rest-api`


sanderploegsma:
Due to a [bug in the Java test runner](https://github.com/exercism/java-test-runner/issues/69) the current tests for the `rest-api` exercise fail to run on Exercism. 
Since this bug is not easily fixed, the best workaround is to refactor the exercise tests into a single class containing all of the test methods, instead of using nested classes. 

## Contribution notes

As always, make sure to follow the [Exercism contributing guidelines](https://exercism.org/docs/building) and the [Java track contributing guidelines](https://github.com/exercism/java/blob/0e1f0088c7fd4580cb7d8f93dd4771743624c802/CONTRIBUTING.md).

Don't hesitate to ask here in case you can't find answers to your questions!

Yordaniss:
@sanderploegsma hey 🙂 

Are the tests for `rest-api` not already in one single class `RestApiTest`? Or do you mean not to use static classes like `UserManagementTest` and move the tests under main class?

sanderploegsma:
> Or do you mean not to use static classes like `UserManagementTest` and move the tests under main class?

Yes, that is what I mean. Apologies if this wasn’t clear from the issue description. I updated it a bit, perhaps now it’s more clear?

Yordaniss:
> > Or do you mean not to use static classes like `UserManagementTest` and move the tests under main class?
> 
> Yes, that is what I mean. Apologies if this wasn’t clear from the issue description. I updated it a bit, perhaps now it’s more clear?
> 
> 

Thanks for an answer. I would like to work than on this issue 🙂

sanderploegsma:
Great, I assigned it to you!

PR #2544, opened 2023-10-29
Remove nested test classes from practice exercise rest-api 

Yordaniss:
closes: #2539

* removed nested classes from main RestApiTest class

---

Reviewer Resources:


exercises/practice/rest-api/src/test/java/RestApiTest.java
```diff
@@ -10,286 +10,282 @@
 @RunWith(Enclosed.class)
```
sanderploegsma, line 10: This annotation is no longer needed and can be removed.

sanderploegsma (commented):
One more thing to take care of, looks good so far!

sanderploegsma (approved):
Thanks for your contribution!

Issue #2538, opened 2023-10-28
Remove nested test classes from practice exercise `simple-cipher`


sanderploegsma:
Due to a [bug in the Java test runner](https://github.com/exercism/java-test-runner/issues/69) the current tests for the `simple-cipher` exercise fail to run on Exercism. 
Since this bug is not easily fixed, the best workaround is to refactor the exercise tests into a single class containing all of the test methods.

## Implementation notes

Since some tests have the same method name (e.g. `RandomKeyCipher.cipherCanDecode()` and `SubstitutionCipher.cipherCanDecode()`) the test methods need to be renamed. 

When doing so, make sure to have a look at the exercise's `.meta/tests.toml` file. 
This contains descriptions for each test; try to have the test method names reflect the descriptions when possible. 
This makes it easy for future maintenance to determine which tests have been implemented already.

## Contribution notes

As always, make sure to follow the [Exercism contributing guidelines](https://exercism.org/docs/building) and the [Java track contributing guidelines](https://github.com/exercism/java/blob/0e1f0088c7fd4580cb7d8f93dd4771743624c802/CONTRIBUTING.md).

Don't hesitate to ask here in case you can't find answers to your questions!

PR #2545, opened 2023-10-30
Remove nested test classes from practice exercise simple-cipher (#2538)

danilopiazza:
# pull request

As suggested in #2538, I refactored the exercise tests into a single class.

All the mehods with the same name have been renamed by prepending the name of the class they come from: for example, `RandomKeyCipher.cipherCanDecode` became `randomKeyCipherCanDecode`, and `SubstitutionCipher.cipherCanDecode` is now `subsitituonCipherCanDecode`.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/simple-cipher/src/test/java/SimpleCipherTest.java
+++ b/exercises/practice/simple-cipher/src/test/java/SimpleCipherTest.java
@@ -1,113 +1,104 @@
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
-import org.junit.experimental.runners.Enclosed;
-import org.junit.runner.RunWith;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-
-@RunWith(Enclosed.class)
 public class SimpleCipherTest {
-    public static class RandomKeyCipher {
-        private Cipher cipherWithDefaultKey;
+    private Cipher randomKeyCipher;
+    private Cipher substitutionCipher = new Cipher("abcdefghij");
 
-        @Before
-        public void setup() {
-            cipherWithDefaultKey = new Cipher();
-        }
-
-        /**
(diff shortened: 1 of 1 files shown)


exercises/practice/simple-cipher/src/test/java/SimpleCipherTest.java
```diff
@@ -1,113 +1,104 @@
...
+    public void randomKeyCipherIsReversible() {
+        String plainText = "abcdefghij";
+        assertThat(randomKeyCipher.decode(randomKeyCipher.encode(plainText))).isEqualTo(plainText);
     }
 
-    public static class SubstitutionCipher {
-        private Cipher cipherWithDefaultKey = new Cipher("abcdefghij");
+    @Ignore("Remove to run test")
+    @Test
+    public void keyIsLowercaseLetters() {
```
sanderploegsma, line 44: I think it makes sense to rename this to `randomCipherKeyIsLowercaseLetters()` as well, since the test only makes sense in the context of a randomly generated key.

sanderploegsma (changes requested):
Thanks for picking this up! I have one small remark to take care of, after that it should be good to merge 👍

exercises/practice/simple-cipher/src/test/java/SimpleCipherTest.java
danilopiazza: You're right, I should have renamed that too. I renamed this method to `randomKeyCipherKeyIsLowercaseLetters` for consistency.

danilopiazza:
Should I squash the two commits into a single one?

sanderploegsma:
> Should I squash the two commits into a single one?

No need, we squash all commits by default when merging the PR 👍

sanderploegsma (approved):

sanderploegsma:
Thanks for contributing! 🙏

PR #2551, opened 2023-11-05
Add practice exercise: killer-sudoku-helper

### LINKED ISSUE #2361: Implement new practice exercise: `killer-sudoku-helper`
The exercise `killer-sudoku-helper` has not been implemented yet for the Java track.
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/main/exercises/killer-sudoku-helper).

sougat818:
# pull request
Fixes #2361
I have followed the [Julia implementation](https://github.com/exercism/julia/tree/main/exercises/practice/killer-sudoku-helper). 
The tests are also similar to the Julia tests over the Python ones. 

One change I have made is to use method overloading for the implementations as `excludes` is optional. I have tagged the exercise `constructors` as that's the closest to method overloading I could find. Please let me know if this is not the focus of the exercise. 

I have assumed recursion to be the focus but the question can always be solved without recursion. 

---

Reviewer Resources:


sanderploegsma:
@sougat818 thanks! I'll try to have a look some time in the next few days. In the mean time I have triggered the Github Actions workflows, so you can already have a look whether all checks pass and fix stuff where needed.

config.json
```diff
@@ -1135,6 +1135,24 @@
...
+          "lists",
+          "recursion"
+        ],
+        "prerequisites": [
+          "basics",
+          "numbers",
+          "loops",
+          "arrays",
+          "if-else-statements"
+        ],
```
sanderploegsma, line 1153: - The prerequisites should list existing concepts (you can find them in the [concepts](https://github.com/exercism/java/tree/main/concepts) directory). I think `loops` should therefore be `for-loops`.
- There is no need to add "transitive" dependencies: the `basics` concept is a prerequisite of other concepts like `numbers`, so you only need to list the concept the exercise directly depends on.
- For practice exercises we normally add the concepts from the `practices` array in the `prerequisites` array too. That way students first have to complete the concept exercise for a concept. Practice exercises can then be used to advance their knowledge in the concept. The exception here is when a concept isn't required to complete the exercise. For example, you mentioned that this exercise is _probably_ going to be solved using recursion, so it sounds like that wouldn't be a prerequisite since there are other ways of solving it.

config.json
```diff
@@ -1135,6 +1135,24 @@
...
         ]
       },
+      {
+        "slug": "killer-sudoku-helper",
+        "name": "Killer Sudoku Helper",
+        "uuid": "8c45a47d-b3e3-484c-a124-90136ec838fd",
+        "practices": [
+          "constructors",
+          "lists",
+          "recursion"
```
sanderploegsma, line 1145: There is no concept `recursion` (yet), so this shouldn't be here.

exercises/practice/killer-sudoku-helper/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+}
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  testImplementation platform("org.junit:junit-bom:5.10.0")
+  testImplementation "org.junit.jupiter:junit-jupiter"
+  testImplementation "org.assertj:assertj-core:3.24.2"
```
sanderploegsma, line 12: The [java-test-runner](https://github.com/exercism/java-test-runner/blob/main/lib/build.gradle) currently has AssertJ Core 3.15.0. Since the test runner is run without any network connections, that means that all submissions from students also run with that version of AssertJ Core.

To stay safe I think it's best to stick to that version here as well. This way we are sure that the test suite can be run on the test runner.

exercises/practice/killer-sudoku-helper/src/test/java/KillerSudokuHelperTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class KillerSudokuHelperTest {
+
+    private KillerSudokuHelper helper = new KillerSudokuHelper();
+
+    @Test
+    public void testTrivialOneDigitCages() {
+        for (int n = 1; n <= 9; n++) {
+            List<List<Integer>> expected = List.of(List.of(n));
+            assertEquals(expected, helper.combinationsInCage(n, 1));
```
sanderploegsma, line 17: [We prefer using AssertJ assertions over JUnit assertions](https://github.com/exercism/java/blob/main/POLICIES.md#prefer-assertj-assertions). Could you please update the tests to use AssertJ instead?

exercises/practice/killer-sudoku-helper/src/test/java/KillerSudokuHelperTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class KillerSudokuHelperTest {
+
+    private KillerSudokuHelper helper = new KillerSudokuHelper();
```
sanderploegsma, line 11: To help catch issues in student implementations where they store state in the class instance, it might be better to create a new instance for every test using a setup method. What do you think?

exercises/practice/killer-sudoku-helper/src/test/java/KillerSudokuHelperTest.java
```diff
@@ -0,0 +1,57 @@
...
+        for (int n = 1; n <= 9; n++) {
+            List<List<Integer>> expected = List.of(List.of(n));
+            assertEquals(expected, helper.combinationsInCage(n, 1));
+        }
+    }
+
+    @Test
+    @Disabled("Remove to run test")
+    public void testCageWithSum45ContainsAllDigits() {
+        List<List<Integer>> expected = List.of(asList(1, 2, 3, 4, 5, 6, 7, 8, 9));
```
sanderploegsma, line 24: Just me asking, but what is the difference between this and the following?

```java
List.of(List.of(1, 2, 3, 4, 5, 6, 7, 8, 9));
```

sanderploegsma (changes requested):
Thanks for picking this up! It already looks really good, there is just a couple of things I'd like you to have a look at.

config.json
sanderploegsma: Taking all of this in mind and looking at [the current syllabus](https://exercism.org/tracks/java/concepts) and how the concepts depend on each other, I think we can make do with the following prerequisites:

- numbers
- lists

config.json
```diff
@@ -1135,6 +1135,24 @@
           "strings"
         ]
       },
+      {
+        "slug": "killer-sudoku-helper",
+        "name": "Killer Sudoku Helper",
+        "uuid": "8c45a47d-b3e3-484c-a124-90136ec838fd",
+        "practices": [
+          "constructors",
```
sanderploegsma, line 1143: > One change I have made is to use method overloading for the implementations as `excludes` is optional. I have tagged the exercise `constructors` as that's the closest to method overloading I could find. Please let me know if this is not the focus of the exercise.

I think this isn't the way to go, instead I think the track could benefit from a concept for method overloading. Tagging it with `constructors` is a bit misleading and will cause confusion as the `KillerSudokuHelper` class doesn't even have any constructors.

config.json
sougat818: Thanks a lot for the detailed explanation. So at the moment this would look like 


```
        "practices": [
          "for-loops"
        ],
        "prerequisites": [
          "basics",
          "numbers"
        ],
```

sanderploegsma: Hmm no, that is not what I meant. I was thinking more like

```json
{
  "practices": [
    "for-loops",
    "if-else-statements",
    "numbers",
    "lists"
  ],
  "prerequisites": [
    "numbers",
    "lists"
  ]
}
```

config.json
sougat818: Should I add the concept pages as part of this PR?

exercises/practice/killer-sudoku-helper/src/test/java/KillerSudokuHelperTest.java
sougat818: `asList `creates mutable lists. I was trying something else out when I added that but finally didn't need it. Its better to use `List.of()` for consistency. Thanks.

sanderploegsma (approved):
Everything here looks great, thank you so much for contributing! 👍

PR #2559, opened 2023-11-08
Run tests using java-test-runner

sanderploegsma:
Inspired by this forum topic: https://forum.exercism.org/t/re-using-the-test-runners-in-corresponding-track-ci-jobs/8070.

Note that this required some updates to some of the exercises, particularly ones where not all files that are part of the example solution were correctly listed in the exercise's `config.json`. 

The Mazy Mice exercise required a small change because of a compilation error due to a class being defined in both the example solution as well as the tests. Not sure why this wasn't caught when using the tests with Gradle. It shouldn't impact students though.

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,3 +14,9 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
       - name: Test all exercises using java-test-runner
         run: bin/test-with-test-runner
+      - name: Archive test results
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-results
+          path: exercises/**/build/results.json
+        if: always()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,16 @@
+name: Test
+
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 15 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
ErikSchierboom, line 12:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Test all exercises using java-test-runner
+        run: bin/test-with-test-runner
+      - name: Archive test results
+        uses: actions/upload-artifact@v3
```
ErikSchierboom, line 18: Could you pin the workflow to a SHA?

ErikSchierboom (approved):

ErikSchierboom (approved):

PR #2565, opened 2023-11-12
concepts: method-overloading

### LINKED ISSUE #2554: Add concept: `method-overloading`
The goal of this issue is to add a new concept to the Java track that introduces `method-overloading`. The corresponding concept exercise can probably be based on the ["Wizards and Warriors 2" exercise from the C# track](https://github.com/exercism/csharp/tree/main/exercises/concept/wizards-and-warriors-2).

Instructions on how to contribute to the Java track and how to implement concept exercises can be found in the [contributing guide](https://github.com/exercism/java/blob/main/CONTRIBUTING.md).

As always, leave a comment here if you want to work on this exercise, and we'll assign it to you. In case you have any further questions, feel free to ask here.

sougat818:
# pull request

closes #2554

---

Reviewer Resources:


exercises/concept/wizards-and-warriors-2/src/test/java/GameMasterTest.java
sanderploegsma, line 1: Most of the other concept exercise tests use `@DisplayName()` to add a human-readable name to each test. This is the name that is shown to students in the online test editor. Could you do the same for these tests?

exercises/concept/wizards-and-warriors-2/src/main/java/GameMaster.java
sanderploegsma, line 1: I'm wondering whether providing stubs for all of the overloads might be giving too much away. The goal of this exercise is to teach students _how to overload methods_. This is also why the implementation for each method is relatively simple. With that in mind, wouldn't it make more sense if they had to define the overloads themselves?

For reference, the same principle applies to the Lasagna exercise, which is the only other exercise not providing a stub for the methods for the exact same reason: the goal is to learn how to _define_ a method.

sanderploegsma, line 1: Since the student should not edit the `TravelMethod` enum nor the `Character` and `Destination` classes, it would make sense to move them to their own respective files, and list these files under `files.editor[]` in the exercise metadata. That way it's clear for students what parts of the exercise should be edited.

Since the files have to be present for both the solution and example source sets, what I suggest is to place these files in `src/main/java` and then create symlinks in `.meta/src/reference/java`. That way the Gradle projects are happy, and there is no way to accidentally update the classes in one source set but not in the other.

sanderploegsma (changes requested):
Thanks so much for working on this, excellent work already! I have left a few comments that should be addressed though.

concepts/method-overloading/about.md
```diff
@@ -0,0 +1,47 @@
+# About 
+
+Method overloading is a key concept in Java, allowing a class to have more than one method with the same name, as long as each method has a different parameter list. This feature is a fundamental aspect of Java's object-oriented programming and provides a way to perform similar operations with different types or numbers of inputs.
```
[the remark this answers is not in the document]
sanderploegsma: Looks like `concepts/method-overloading/introduction.md` was missed, could you check that one again and then re-run `bin/configlet generate` to sync the exercise introduction?

sanderploegsma:
@sougat818 Are you still working on this? Please comment if you are, otherwise I'm going to close this due to inactivity.

sougat818:
> @sougat818 Are you still working on this? Please comment if you are, otherwise I'm going to close this due to inactivity.

Yes I am. Been a bit busy, will pick up soon . 🙏

exercises/concept/wizards-and-warriors-2/src/main/java/GameMaster.java
sougat818: I have made this change but the configlet seems to not like the files.editor bit. I'm not sure what is wrong in there.

sougat818: @sanderploegsma I have added the symlinks. While updating the tests in the same lines as the `Lasagna` exercise. I am having trouble accessing the `TravelMethod` enum nor the `Character` and `Destination` classes in the utils package for the Reflection Proxy . I could either move the Reflection Classes into the default package or Move all the other files into a non default class. Just wanted to check before doing any of it as there will surely be some side effects. Since the `Lasagna` exercise only uses primitive types this wasn't a problem in that exercise.

sanderploegsma: > I have made this change but the configlet seems to not like the files.editor bit. I'm not sure what is wrong in there.

It looks like the configlet is complaining about the formatting of the exercise's `.meta/config.json`. You can see if this fixes it:

```sh
bin/configlet fmt --update --exercise wizards-and-warriors-2
```

exercises/concept/wizards-and-warriors-2/.meta/src/reference/java/GameMaster.java
```diff
@@ -0,0 +1,28 @@
...
+        return "You're traveling to your destination on horseback.";
+
+    }
+
+    public static String describe(Character character, Destination destination, TravelMethod travelMethod) {
+        return describe(character) + " " + describe(travelMethod) + " " + describe(destination);
+    }
+
+    public static String describe(Character character, Destination destination) {
+        return describe(character) + " " + describe(TravelMethod.WALKING) + " " + describe(destination);
```
sanderploegsma, line 26: Actually maybe it would be good practice for students to learn calling one overload from another as part of this exercise, since that is very common in Java, especially when using multiple constructors.

```suggestion
        return describe(character, destination, TravelMethod.WALKING);
```

exercises/concept/wizards-and-warriors-2/src/main/java/GameMaster.java
```diff
@@ -0,0 +1,13 @@
...
+
+    // TODO: define the 'describe(Character character)' static method
+
+    // TODO: define the 'describe(Destination destination)' static method
+
+    // TODO: define the 'describe(TravelMethod travelMethod)' static method
+
+    // TODO: define the 'describe(Character character, Destination destination, TravelMethod travelMethod)' static method
+
+    // TODO: define the 'describe(Character character, Destination destination)' static method
```
sanderploegsma, line 11: These TODO comments still give away too much IMO, as they contain the full method signature so it's too easy to figure out I think. How about this?

```suggestion
    // TODO: define a 'describe' method that returns a description of a Character

    // TODO: define a 'describe' method that returns a description of a Destination

    // TODO: define a 'describe' method that returns a description of a TravelMethod

    // TODO: define a 'describe' method that returns a description of a Character, Destination and TravelMethod

    // TODO: define a 'describe' method that returns a description of a Character and Destination
```

sanderploegsma (commented):
This is getting there, great work! I left some additional comments for you to look at, and there are some outstanding comments left to take care of. I resolved the ones that have been addressed.

exercises/concept/wizards-and-warriors-2/src/main/java/GameMaster.java
sanderploegsma: > @sanderploegsma I have added the symlinks. While updating the tests in the same lines as the `Lasagna` exercise. I am having trouble accessing the `TravelMethod` enum nor the `Character` and `Destination` classes in the utils package for the Reflection Proxy . I am moving the Reflection Classes into the default package. Just wanted to check as there will might be some side effects. Since the `Lasagna` exercise only uses primitive types this wasn't a problem in that exercise.

OK, let me have a look to see whether I could get this to work. I assumed it would work similar to the `Lasagna` exercise but if not, maybe there is another way we can do it.

sanderploegsma: OK I see what you mean, and I don't think there is an issue moving the `ReflectionProxy` to the default package. The only side-effect I see is that you have to use a different name for the `GameMaster` proxy class, as it can't also be called `GameMaster`. Using `GameMasterProxy` should be fine though.

exercises/concept/wizards-and-warriors-2/.meta/src/reference/java/GameMaster.java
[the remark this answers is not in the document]
sougat818: I prefer string concats for readability, however I don't really mind changing it. The exemplar is for students and not me 😄 . Whatever would be best for them. Its also super easy to change with IDEs. It could also be because I personally feel the String.format is too C like 😆 .

exercises/concept/wizards-and-warriors-2/src/main/java/GameMaster.java
sougat818: Yeah its working fine in terms of tests and code. I was wondering more about how this would be displayed on the UI etc if there are any side effects to that with these files being in the default package. They are not in the files list in the config so my guess is it would be fine.

sougat818: Another change needed was to always throw exceptions from `invokeMethod` . For Lasagna returning `null` for an `int` return type would fail but in this scenario it would continue with the Assertion.

sanderploegsma: > Yeah its working fine in terms of tests and code. I was wondering more about how this would be displayed on the UI etc if there are any side effects to that with these files being in the default package. They are not in the files list in the config so my guess is it would be fine.

Correct, any files not listed in the `.meta/config.json` are not shown to students in the online editor. For now that is fine, I'm not aware of tooling that expects _all_ exercise files to be listed.

Students downloading the exercise via the CLI of course can see the files, which makes me think whether we should add a class-level comment pointing them to the `GameMasterTest` class, something like this:

```java
/**
 * This is a helper class to be able to run the tests for this exercise.
 * The tests are located in the {@link GameMasterTest} class.
 * @see GameMasterTest
 */
public class GameMasterProxy {
    // ...
}
```

exercises/concept/wizards-and-warriors-2/.docs/instructions.md
sanderploegsma, line 1: Can you make sure that this file contains one sentence per line like the other Markdown files?

concepts/method-overloading/introduction.md
sanderploegsma, line 1: Can you make sure that this file contains one sentence per line like the other Markdown files?

After that, regenerate the exercise introduction using

```sh
bin/configlet generate
```

sanderploegsma (approved):
Really great work, thank you for bearing with me during the review! 😀

PR #2588, opened 2023-11-28
Fix broken test in Captains Log

### LINKED ISSUE #2589: Captain's Log, typo in the instructions or error in the test for task 2
Hi,

That challenge was funny to solve. Really great exercise to work with random number :)

For the task 2, you ask to generate a registry number that starts with the prefix "NCC-" and then uses a number from 1000 to 9999 (inclusive).

<details>
<summary>So that code should validates the challenge.</summary>  
return String.format("NCC-%d", 1000 + this.random.nextInt(9000));
</details>

But it doesn't. I had to subtract 1 from 9000 to pass the tests. That means the result is a random number in this range [1000 - 9998] and not [1000 - 9999].

So:
- or there's a typo in the instructions and it's not 9999 (inclusive) but 9999 (exclusive)
- or the unit test is wrong. The upper bound of this assertion [at line 63](https://github.com/exercism/java/blob/b0db7b6c078a950fff0282afe69d7354773740c1/exercises/concept/captains-log/src/test/java/CaptainsLogTest.java#L63) is exclusive.

WDYT?

santialb:
Changed the end index of the randomShipRegistryNumber() so that the range goes from 1000 to 9999

Fixes #2589 

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/captains-log/.meta/src/reference/java/CaptainsLog.java
+++ b/exercises/concept/captains-log/.meta/src/reference/java/CaptainsLog.java
@@ -17,7 +17,7 @@ class CaptainsLog {
 
     String randomShipRegistryNumber() {
         var start = 1000;
-        var end = 9999;
+        var end = 11000;
         return String.format("NCC-%d", start + random.nextInt(end - start));
     }
 
--- a/exercises/concept/captains-log/.meta/src/reference/java/CaptainsLog.java
+++ b/exercises/concept/captains-log/.meta/src/reference/java/CaptainsLog.java
@@ -17,7 +17,7 @@ class CaptainsLog {
 
     String randomShipRegistryNumber() {
         var start = 1000;
-        var end = 11000;
+        var end = 10000;
         return String.format("NCC-%d", start + random.nextInt(end - start));
     }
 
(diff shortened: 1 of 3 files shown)


exercises/concept/captains-log/.meta/src/reference/java/CaptainsLog.java
```diff
@@ -17,7 +17,7 @@ char randomPlanetClass() {
 
     String randomShipRegistryNumber() {
         var start = 1000;
-        var end = 9999;
+        var end = 11000;
```
sanderploegsma, line 20: I think this should be `10000`, not `11000`: the numbers should be in the range `1000`-`9999`, so the upper limit should be `9999`+`1`.

sanderploegsma (commented):
Great start, and I appreciate your willingness to try and contribute!

Unfortunately it looks like you made a small typo when updating `CaptainsLog.java`, see the inline comment.

Next to that the test class also needs updating. If you make sure to commit all changes to the same branch (called `patch-1` in your copy of this repository) they should all end up in this same Pull Request. 👍

santialb (commented):
I made the changes to the previous PR, changing the end range to 10000 and changing the test file to the new expected values (which are NCC-6258, NCC-1683 and NCC-4922 respectively).

sanderploegsma (approved):
Great work, thanks for fixing this!

sanderploegsma:
Congratulations with your first contribution @santialb! 🎉

PR #2594, opened 2023-12-04
Update config.json

mic-en:
removed Exercise darts from syllabus "constructors"

# pull request

As discussed here: https://forum.exercism.org/t/syllabus-darts-exercise-and-constructors/8389/2
I suggest to remove the darts exercise from constructor syllabus

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -586,8 +586,7 @@
           "if-else-statements"
         ],
         "prerequisites": [
-          "if-else-statements",
-          "constructors"
+          "if-else-statements"
         ],
         "difficulty": 2
       },
--- a/config.json
+++ b/config.json
@@ -583,8 +583,7 @@
         "name": "Darts",
         "uuid": "4d400a44-b190-4a0c-affb-99fad8ea18da",
         "practices": [
-          "if-else-statements",
-          "constructors"
+          "if-else-statements"
         ],
         "prerequisites": [
           "if-else-statements",


config.json
```diff
@@ -583,8 +583,7 @@
         "name": "Darts",
         "uuid": "4d400a44-b190-4a0c-affb-99fad8ea18da",
         "practices": [
-          "if-else-statements",
-          "constructors"
+          "if-else-statements"
         ],
         "prerequisites": [
```
sanderploegsma, line 588: I see that the "constructors" concept is also in the list of prerequisites for this exercise. This means that the exercise is only unlocked after completing the learning exercise covering the "constructors" concept.

Since we agreed that the concept is not applicable for this exercise, we should also remove it as a prerequisite by removing it from this list.

sanderploegsma (changes requested):
Thanks for opening a PR to fix this! I have one inline comment to resolve and then we can get it merged.

config.json
sanderploegsma: @mic-en I see you resolved this, did you make the change but forget to push perhaps? I don't see any new commits on this PR.

mic-en:
thanks for the hint, now it should be fine :)

sanderploegsma:
Great, thanks for fixing! 👍

PR #2595, opened 2023-12-04
Removing (static) from exception message

smcg468:
Removing '(static)' from the exception message thrown for the tim-from-marketing concept exercise.
---

Reviewer Resources:


sanderploegsma (approved):
Good catch, thanks!

Issue #2388, opened 2023-09-22
[Recurring] Keep practice exercises up-to-date with problem specifications


sanderploegsma:
This is a recurring task that has to be done from time to time. 
**Do not reference this issue by writing `resolves/fixes #XXX` in your PR because that will close it.**

We want to keep the track up to date with [problem specifications](https://github.com/exercism/problem-specifications), which means regularly checking if anything changes and needs updating.

How to do this task:

## Step 1: fetch configlet

Run:

```sh
./bin/fetch-configlet
```

## Step 2: check which exercises need updating and in which way

Run:

```sh
./bin/configlet sync 
```

The output will list all Java exercises that are out of sync in some way. Example output:

```text
[warn] docs: instructions unsynced: affine-cipher
[warn] docs: instructions unsynced: all-your-base

[warn] metadata: unsynced: acronym
[warn] metadata: unsynced: armstrong-numbers

[warn] collatz-conjecture: missing 2 test cases
       - zero is an error (2187673d-77d6-4543-975e-66df6c50e2da)
       - negative value is an error (ec11f479-56bc-47fd-a434-bcd7a31a7a2e)
```

Choose one or more problems and attempt to fix it. 
Metadata and doc updates can be all done together for many exercises in a single PR, but please create separate PRs per exercise when changing tests.

## Step 3.1: Sync all docs

Run this command to update all exercise docs (introductions):

```sh
 ./bin/configlet sync --docs --update
```

Press `y` when asked `sync the above docs ([y]es/[n]o)?`.

## Step 3.2: Sync all metadata

Run this command to update all exercise metadata:

```sh
 ./bin/configlet sync --metadata --update
```

Press `y` when asked `sync the above metadata ([y]es/[n]o)?`.

## Step 3.3: Update tests for an exercise

The general goal is that all Java practice exercises follow problem specifications, except for when it doesn't make any sense in Java.

Important:

Get familiar with the [problem specifications](https://github.com/exercism/problem-specifications) repository and the canonical-data.json format it uses for describing tests cases.
Read [CONTRIBUTING.md](https://github.com/exercism/java/blob/main/CONTRIBUTING.md) for the general rules of this repository.
Please create separate a PR per exercise when changing tests.

### Step 3.3.1 Update tests.toml

Update `tests.toml` by running this command for a chosen exercise:

```sh
./bin/configlet sync --exercise <exercise-slug> --tests --update
```

[...]

PR #2615, opened 2023-12-30
Sync tests for practice exercise ``markdown``

manumafe98:
# pull request

The goal is to sync the tests of the markdown practice exercise

Reviewer Resources:

### DIFF
--- a/exercises/practice/markdown/.meta/src/reference/java/Markdown.java
+++ b/exercises/practice/markdown/.meta/src/reference/java/Markdown.java
@@ -57,6 +57,10 @@ class Markdown {
             return null;
         }
 
+        if (count > 6) {
+            return parseParagraph(markdown);
+        }
+
         return wrap(markdown.substring(count + 1), "h" + Integer.toString(count));
     }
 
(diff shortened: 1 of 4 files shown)


manumafe98:
I guess to solve the tests failing I should update this function

https://github.com/exercism/java/blob/56b0d792aea56904e0053a63e45b4819fa33162e/exercises/practice/markdown/src/main/java/Markdown.java#L43

So if the count is greater than 7 to output the input but in a paragraph, right?

sanderploegsma (approved):

sanderploegsma:
> I guess to solve the tests failing I should update this function
> 
> https://github.com/exercism/java/blob/56b0d792aea56904e0053a63e45b4819fa33162e/exercises/practice/markdown/src/main/java/Markdown.java#L43
> 
> So if the count is greater than 7 to output the input but in a paragraph, right?

Yes, but don't update the one in `src/main/java`, update the one in `.meta/src/reference/java` instead. The former is the stub file received by students, the latter is the reference implementation.

manumafe98:
Yeah I was going to update that one as well, but the goal of the exercise is that passes all test and students reformat the code, that's why I thought on updating ``src/main/java``.

sanderploegsma:
> Yeah I was going to update that one as well, but the goal of the exercise is that passes all test and students reformat the code, that's why I thought on updating ``src/main/java``.

Ah yes that's right, I forgot that this is a refactoring exercise. In that case you have to update both indeed. Sorry for the mixup!

sanderploegsma (changes requested):
As discussed the `src/main/java/Markdown.java` file needs to be updated to reflect the new behavior as well.

sanderploegsma (approved):
Thanks! 🙏

PR #2617, opened 2024-01-01
Sync tests for practice exercise ``protein translation``

manumafe98:
# pull request

The goal is to sync the tests of the phone number protein translation

Reviewer Resources:

### DIFF
--- a/exercises/practice/protein-translation/src/test/java/ProteinTranslatorTest.java
+++ b/exercises/practice/protein-translation/src/test/java/ProteinTranslatorTest.java
@@ -13,6 +13,11 @@ public class ProteinTranslatorTest {
     public void setUp() {
         proteinTranslator = new ProteinTranslator();
     }
+    
+    @Test
+    public void testEmptyRnaSequenceResultInNoproteins() {
+        assertThat(proteinTranslator.translate("")).isEmpty();
+    }
 
     @Test
     public void testMethionineRnaSequence() {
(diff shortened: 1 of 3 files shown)


manumafe98:
I'm  going to update the reference resolution on this one as well

exercises/practice/protein-translation/src/test/java/ProteinTranslatorTest.java
```diff
@@ -149,7 +150,54 @@ public void testTranslationStopsIfStopCodonInMiddle1() {
...
         assertThat(proteinTranslator.translate("UGGUGUUAUUAAUGGUUU"))
-            .containsExactly("Tryptophan", "Cysteine", "Tyrosine");
+                .containsExactly("Tryptophan", "Cysteine", "Tyrosine");
     }
 
+    @Ignore("Remove to run test")
+    @Test
+    public void testEmptyRnaSequenceResultInNoproteins() {
+        assertThat(proteinTranslator.translate("")).isEmpty();
+    }
```
sanderploegsma, line 160: Note: The order of the tests in the canonical data may be ordered like that on purpose, for instance to cover basic test cases before more elaborate ones. This should aid students in their implementation.

To that end, we try to keep the order of the tests in the Java track in the same way as they are ordered in the canonical data. That would mean that this test for an empty sequence becomes the first test. Note that this also means that it shouldn't have an `@Ignore` tag and that the existing `testTranslationOfRnaToProteinList` method should be `@Ignore`d.

sanderploegsma (approved):
Changes look good! I left one comment regarding the order of the tests compared to the canonical data, could you have a look at that?

exercises/practice/protein-translation/src/test/java/ProteinTranslatorTest.java
manumafe98: Oh okey good to know! I updated the order to match the canonical data!

sanderploegsma (approved):

PR #2619, opened 2024-01-02
Sync tests for practice exercise ``parallel letter frequency``

manumafe98:
# pull request

The goal is to sync the tests of the parallel letter frequency practice exercise

Reviewer Resources:


exercises/practice/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
sanderploegsma, line 1: I think that this refactor lost track of this requirement from the instructions (emphasis mine)

> Count the frequency of letters in texts **using parallel computation**.
>
> Parallelism is about doing things in parallel that can also be done sequentially. A common example is counting the frequency of letters. Create a function that returns the total frequency of each letter in a list of texts and **that employs parallelism**.

This is a key aspect of this exercise, so I don't think we can get away with not using parallelism in the example solution.

sanderploegsma (changes requested):

exercises/practice/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
manumafe98: Oh okey, I will try to solve it using parallelism then, do you have any suggestions on how I can take on it? Thanks!

exercises/practice/parallel-letter-frequency/src/main/java/ParallelLetterFrequency.java
```diff
@@ -2,12 +2,12 @@
 
 class ParallelLetterFrequency {
 
-    ParallelLetterFrequency(String letters) {
+    ParallelLetterFrequency(String[] letters) {
         throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
     }
 
-    Map<Integer, Integer> letterCounts() {
+    Map<Character, Integer> letterCounts() {
```
sanderploegsma, line 9: Maybe rename this method to `countLetters()` or `getLetterCounts()` since we are changing it anyway.

exercises/practice/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
manumafe98: Okey so as this passes to be an array instead of a String, I could leave that function literally the same and call it every iteration, but my problem was how do I achieve the final value, like can I sum the Maps that are being generated?

sanderploegsma: I'm not sure I get what you mean. In the snippet above, `this.letters` is a single `String` containing all of the letters of the array of strings passed to the constructor, right? So there is no need to iterate over anything in `letterCounts()`.

manumafe98: Check this:

````
The following test case is missing:
{
  "uuid": "708ff1e0-f14a-43fd-adb5-e76750dcf108",
  "description": "two texts with one letter",
  "property": "calculateFrequencies",
  "input": {
    "texts": [
      "e",
      "f"
    ]
  },
  "expected": {
    "e": 1,
    "f": 1
  }
}

Do you want to include the test case ([y]es/[n]o/[s]kip)?

````

The new Synced tests instead of passing a ``String`` are passing an ``Array``
If you check with ``bin/configlet sync --tests --update --exercise parallel-letter-frequency`` you probably will understand better what im talking about hahaha

sanderploegsma: Yes I noticed that from the updated tests indeed. Still I fail to see why the implementation I posted above wouldn't work.

The constructor takes an array of strings as input (so it should be compatible with the tests) and concatenates them all into one big string, removing any characters that are not letters. After that, the original implementation can be reused that counts the occurrence of each letter in that string in parallel. 

What am I missing?

manumafe98: I reverted the change of the map from ``<Character, Integer>`` to ``<Integer, Integer>`` and change the test casting the chars as integers, and that was it. Sorry for the misunderstanding your example was above was great!

exercises/practice/parallel-letter-frequency/src/main/java/ParallelLetterFrequency.java
manumafe98: Tell me if this is still required after reverting again to ``<Integer, Integer>``

exercises/practice/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
sanderploegsma: I couldn't help myself and changed it back to using `Map<Character, Integer>`, since I feel it would be more natural for students to work with chars instead of ints. Hope you don't mind 😉

sanderploegsma (approved):

exercises/practice/parallel-letter-frequency/.meta/src/reference/java/ParallelLetterFrequency.java
manumafe98: Oh yeah! I agree with you, that's why originally I changed it to ``Map<Character, Integer>`` and when rollbacking to the solution using parallelism I thought it needed to be ```<Integer, Integer>``. Im still learning, so thanks for the review!

Issue #1750, opened 2019-10-01
resistor-color-trio: implement exercise


lemoncurry:
The exercise **resistor-color-trio** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/main/exercises/resistor-color-trio). 

How to implement a new exercise for the Java track is described in detail in [CONTRIBUTING.md](https://github.com/exercism/java/blob/main/CONTRIBUTING.md). Please have a look there first before starting working on the exercise. 
Also please make sure it is clear that you are currently working on this issue, either by asking to be assigned to it, or by opening an empty PR. 

When opening an PR, please reference this issue using any of the [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords). 

If you have had a look at the exercise description and you concluded that the exercise might not be possible to implement in the Java language, please leave a comment and describe the problem. 

In case you have any further questions, feel free to ask here.

HuskyNator:
I'd like to be assigned to this issue.

andrerfcsantos:
@HuskyNator Still interested in working on this?

GhostCoderz:
I would like to be assigned to this issue

andrerfcsantos:
@GhostCoderz Sorry for taking some time to reply. Are you still interested in working on this?

manumafe98:
@sanderploegsma there is someone working on this one? I would like to try on if not

sanderploegsma:
> @sanderploegsma there is someone working on this one? I would like to try on if not

Fair question, I think the latest PR referencing this issue was closed for inactivity. IMO it's fine if you want to work on this! 👍

PR #2637, opened 2024-01-11
Adding ``resistor color trio`` practice exercise

manumafe98:
# pull request

Related issue: [#1750](https://github.com/exercism/java/issues/1750)

The goal is to add the resistor color trio exercise to the java track

Reviewer Resources:


manumafe98:
I have a couple of doubts on this one mainly from the ``config.json`` file.
I used the same difficulty and requirements that the ``resistor-color-duo`` exercise, is it correct? 
And also the ``uuid`` I generated one, but it says it has to be the same across all tracks, but I checked the pythond and go tracks and they're different there.

sanderploegsma:
> I used the same difficulty and requirements that the `resistor-color-duo` exercise, is it correct?

I think that's fine. Other tracks use the same difficulty for all three exercises as well, IIRC.

> And also the `uuid` I generated one, but it says it has to be the same across all tracks, but I checked the pythond and go tracks and they're different there.

Hmm, where does it say that? UUIDs across Exercism should be unique as far as I can remember, so if something somewhere says otherwise then it should probably be updated.

config.json
```diff
@@ -456,6 +456,17 @@
...
         "difficulty": 2
       },
+      {
+        "slug": "resistor-color-trio",
+        "name": "Resistor Color Trio",
+        "uuid": "e8e6c84e-4982-4bb0-af49-fd67e6f32920",
+        "practices": [],
+        "prerequisites": [
+          "arrays",
+          "enums"
```
sanderploegsma, line 466: This exercise doesn't seem to use enums, so I'm not sure it should be a prerequisite.

Actually, now that I double-check it, the same is true for the `resistor-color-duo` exercise. The example implementation happens to use an enum, but that is by no means the only way to solve it.

Could you remove the `enums` concept from the prerequisites for both exercises?

sanderploegsma (changes requested):
One thing to fix, the rest looks really great! 😄

config.json
manumafe98: Sure!

manumafe98:
> > I used the same difficulty and requirements that the `resistor-color-duo` exercise, is it correct?
> 
> I think that's fine. Other tracks use the same difficulty for all three exercises as well, IIRC.
> 
> > And also the `uuid` I generated one, but it says it has to be the same across all tracks, but I checked the pythond and go tracks and they're different there.
> 
> Hmm, where does it say that? UUIDs across Exercism should be unique as far as I can remember, so if something somewhere says otherwise then it should probably be updated.

So, what I read is that it should be the same across all tracks but check this:

https://github.com/exercism/go/blob/cb6c3b60290671552e1a16423deb3ebe20ad813c/config.json#L1274
https://github.com/exercism/python/blob/7e5a83d333424c615b9a4886fc11ce23d4d1c1e0/config.json#L401

manumafe98:
Oh another thing I want to mention @sanderploegsma  is that locally my tests were failing in the ``checkstyle`` task
````
> Task :practice:resistor-color-trio:checkstyleTest FAILED
[ant:checkstyle] [ERROR] E:\repos\java\exercises\practice\resistor-color-trio\src\test\java\ResistorColorTrioTest.java:1: La l├¡nea de finalizaci├│n prevista para el archivo es LF(\n), pero se detecta CRLF(\r\ n). [NewlineAtEndOfFile]
````
And here are working correctly, which was a weird behavior because they were asking me to delete a newline at the end of the file, but when deleted then fails asking me to add it.

sanderploegsma:
> So, what I read is that it should be the same across all tracks

Yes, I got that, and I was wondering where you read that. Because this is what is currently in the [documentation](https://exercism.org/docs/building/tracks/config-json#h-practice-exercises):

> `uuid`: a V4 UUID that uniquely identifies the exercise. **The UUID must be unique both within the track as well as across all tracks**, and must never change.

So if some other resource says otherwise, we may need to notify someone on the staff.

sanderploegsma:
> Oh another thing I want to mention @sanderploegsma is that locally my tests were failing in the `checkstyle` task
> 
> ```
> > Task :practice:resistor-color-trio:checkstyleTest FAILED
> [ant:checkstyle] [ERROR] E:\repos\java\exercises\practice\resistor-color-trio\src\test\java\ResistorColorTrioTest.java:1: La l├¡nea de finalizaci├│n prevista para el archivo es LF(\n), pero se detecta CRLF(\r\ n). [NewlineAtEndOfFile]
> ```
> 
> And here are working correctly, which was a weird behavior because they were asking me to delete a newline at the end of the file, but when deleted then fails asking me to add it.

My Spanish is not great, but it looks like it's complaining about the line endings being CRLF (which is common on Windows, equal to `\r\n`) while it wants LF (`\n`, mostly used in the Unix world).

manumafe98:
> > So, what I read is that it should be the same across all tracks
> 
> Yes, I got that, and I was wondering where you read that. Because this is what is currently in the [documentation](https://exercism.org/docs/building/tracks/config-json#h-practice-exercises):
> 
> > `uuid`: a V4 UUID that uniquely identifies the exercise. **The UUID must be unique both within the track as well as across all tracks**, and must never change.
> 
> So if some other resource says otherwise, we may need to notify someone on the staff.

My bad! I think I misunderstood it, because reading it again you're right

manumafe98:
> > Oh another thing I want to mention @sanderploegsma is that locally my tests were failing in the `checkstyle` task
> > ```
> > > Task :practice:resistor-color-trio:checkstyleTest FAILED
> > [ant:checkstyle] [ERROR] E:\repos\java\exercises\practice\resistor-color-trio\src\test\java\ResistorColorTrioTest.java:1: La l├¡nea de finalizaci├│n prevista para el archivo es LF(\n), pero se detecta CRLF(\r\ n). [NewlineAtEndOfFile]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And here are working correctly, which was a weird behavior because they were asking me to delete a newline at the end of the file, but when deleted then fails asking me to add it.
> 
> My Spanish is not great, but it looks like it's complaining about the line endings being CRLF (which is common on Windows, equal to `\r\n`) while it wants LF (`\n`, mostly used in the Unix world).

Sorry I forgot to translate it! hahaha, but yeah is basically what you said, maybe is because im running the tests on windows, and here the pipeline uses some unix container or something like that.

PR #2640, opened 2024-01-13
Sync tests for practice exercise ``space age``

manumafe98:
# pull request

This one required a major reformat, due to tests passing a new parameter and also usage of a deprecated method ``setScale``
````
{
  "uuid": "57b96e2a-1178-40b7-b34d-f3c9c34e4bf4",
  "description": "invalid planet causes error",
  "property": "age",
  "input": {
    "planet": "Sun",
    "seconds": 680804807
  },
  "expected": {
    "error": "not a planet"
  }
}
````

I used as reference the ``go`` track that is updated.

Reviewer Resources:

### DIFF
--- a/exercises/practice/space-age/.meta/src/reference/java/SpaceAge.java
+++ b/exercises/practice/space-age/.meta/src/reference/java/SpaceAge.java
@@ -1,73 +1,37 @@
-import java.math.BigDecimal;
+import java.util.Map;
+import java.util.HashMap;
 
 class SpaceAge {
 
-    private enum Planet {
-        EARTH(1.0),
-        MERCURY(0.2408467),
-        VENUS(0.61519726),
-        MARS(1.8808158),
-        JUPITER(11.862615),
-        SATURN(29.447498),
-        URANUS(84.016846),
-        NEPTUNE(164.79132);
-
-        private final double relativeOrbitalPeriod;
-
-        Planet(double relativeOrbitalPeriod) {
-            this.relativeOrbitalPeriod = relativeOrbitalPeriod;
-        }
-
-        private double getRelativeOrbitalPeriod() {
-            return relativeOrbitalPeriod;
-        }
-    }
+    private Map<String, Double> planetsOrbitalPeriod = new HashMap<>() {{
+            put("Mercury", 0.2408467);
+            put("Venus", 0.61519726);
+            put("Mars", 1.8808158);
+            put("Jupiter", 11.862615);
+            put("Saturn", 29.447498);
+            put("Uranus", 84.016846);
+            put("Neptune", 164.79132);
+            put("Earth", 1.0);
+        }};
 
     private static final double EARTH_ORBITAL_PERIOD_IN_SECONDS = 31557600.0;
-    private static final int PRECISION = 2;
-
     private double seconds;
+    private String planet;
 
-    SpaceAge(double seconds) {
+    SpaceAge(double seconds, String planet) {
         this.seconds = seconds;
+        this.planet = planet;
+
+        if (planetsOrbitalPeriod.get(planet) == null) {
+            throw new IllegalArgumentException("not a planet");
+        }
     }
 
-    double onEarth() {
-        return calculateAge(Planet.EARTH);
-    }
-
-    double onMercury() {
-        return calculateAge(Planet.MERCURY);
-    }
-
-    double onVenus() {
-        return calculateAge(Planet.VENUS);
-    }
-
-    double onMars() {
-        return calculateAge(Planet.MARS);
-    }
-
-    double onJupiter() {
-        return calculateAge(Planet.JUPITER);
-    }
-
-    double onSaturn() {
-        return calculateAge(Planet.SATURN);
-    }
-
-    double onUranus() {
-        return calculateAge(Planet.URANUS);
-    }
-
-    double onNeptune() {
-        return calculateAge(Planet.NEPTUNE);
-    }
-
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by sanderploegsma — CHANGES_REQUESTED** [review]
This is a good learning opportunity to show that sometimes we can choose _not_ to implement a test from the canonical data: the design of the exercise is such that it doesn't allow any unknown planets, so the extra test case isn't applicable at all.

I double-checked how the [C# implementation](https://github.com/exercism/csharp/tree/main/exercises/practice/space-age) handled this and they chose not to include this test case. In my opinion we should follow their lead here, since the exercises are very similar.

So to conclude, please revert all changes and instead mark the new test as "not included" in `.meta/tests.toml`, along with a comment explaining why. It should look something like this:


```toml
[57b96e2a-1178-40b7-b34d-f3c9c34e4bf4]
description = "invalid planet causes error"
include = false
comment = "Excluded because the design of the exercise does not allow for arbitrary planets"
```

manumafe98:
Perfect! Would be great if Exercism does the same across all tracks i guess? Because for example in the [Go implementation](https://github.com/exercism/go/blob/main/exercises/practice/space-age/.meta/tests.toml) it is applied.

sanderploegsma:
> Perfect! Would be great if Exercism does the same across all tracks i guess? Because for example in the [Go implementation](https://github.com/exercism/go/blob/main/exercises/practice/space-age/.meta/tests.toml) it is applied.

Well that is because they chose to implement it with a different method signature. If our exercise would have looked similar to that one, we might have chosen to include the test that checks for an invalid planet.

It is up to each track to choose how to implement each practice exercise according to that programming language's best practices. For Go, the choice of accepting a `planet` argument makes sense as it's not as object-oriented as Java and C#. It also doesn't really have the concept of enums, which would have been another way to restrict the values it accepts at compile-time.

manumafe98:
> > Perfect! Would be great if Exercism does the same across all tracks i guess? Because for example in the [Go implementation](https://github.com/exercism/go/blob/main/exercises/practice/space-age/.meta/tests.toml) it is applied.
> 
> Well that is because they chose to implement it with a different method signature. If our exercise would have looked similar to that one, we might have chosen to include the test that checks for an invalid planet.
> 
> It is up to each track to choose how to implement each practice exercise according to that programming language's best practices. For Go, the choice of accepting a `planet` argument makes sense as it's not as object-oriented as Java and C#. It also doesn't really have the concept of enums, which would have been another way to restrict the values it accepts at compile-time.

Great! just in case i've maintained the removal of the depracated method ``setScale`` and applied what you suggested me to the tests in the ``.meta/tests.toml``

exercises/practice/space-age/.meta/src/reference/java/SpaceAge.java
```diff
@@ -65,9 +61,9 @@ private double getRelativeOrbitalPeriod() {
     }
 
     private double calculateAge(Planet planet) {
-        double age = seconds / (EARTH_ORBITAL_PERIOD_IN_SECONDS * planet.getRelativeOrbitalPeriod());
-
-        return new BigDecimal(age).setScale(PRECISION, BigDecimal.ROUND_HALF_UP).doubleValue();
+        double planetYears = (seconds / EARTH_ORBITAL_PERIOD_IN_SECONDS) / planet.getRelativeOrbitalPeriod();
+        
+        return (double) Math.round(planetYears * 100) / 100;
```
sanderploegsma, line 66: This change is due to the fact that `BigDecimal.setScale(int, int)` is deprecated, correct?

I looked around and it seems like the recommended way is to use [`BigDecimal.setScale(int, RoundingMode)`](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#setScale-int-java.math.RoundingMode-). Although I don't mind this approach either.

Up to you to choose which way you prefer!

sanderploegsma (approved):

exercises/practice/space-age/.meta/src/reference/java/SpaceAge.java
manumafe98: I don't have a strong opinion on this one, so if you see the BigDecimal approach better tell me and I will make it work hahaha

sanderploegsma: To be honest my opinion is that the exercise should have to round the results at all, instead the test cases should be implemented so that they compare the results only for the first two decimals...

Nothing in the instructions talks about rounding, I've always found it strange that this is something the students should figure out...

manumafe98: Yeah I agree, we could perform the rounding or decimal check in the tests, should be taken as a separate issue or you want me to try solve it in this PR?

sanderploegsma: Feel free to change it in this PR. I suggest having a look at the [`AbstractDoubleAssert.isEqualTo(double, Offset<Double>)`](https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractDoubleAssert.html#isEqualTo(double,%20org.assertj.core.data.Offset)) method.

manumafe98: Thanks for the advice! I will give it a try!

manumafe98: Done!

sanderploegsma (approved):
This looks great!

PR #2658, opened 2024-01-20
Enable representer and analyzer

sanderploegsma:
- The representer was recently disabled because it kept crashing; this bug has been fixed here: https://github.com/exercism/java-representer/pull/82
- The analyzer is still in an early stage: it only has an implementation for the `hamming`, `leap` and `two-fer` exercises. However, due to a recent addition that provides some basic general feedback to _all_ exercises I think it makes sense to enable it. Especially since there are a lot of questions on Discord from people who are struggling with getting tests to pass because they are printing the answers instead of returning them (which is something the analyzer informs them of).

---

Reviewer Resources:


ErikSchierboom (approved):
I did a local test and the analyzer and representer seemed to work well.

ErikSchierboom:
@sanderploegsma There seems to be an issue with the representer: https://github.com/exercism/java-representer/issues/104

Issue #2049, opened 2021-10-23
squeaky-clean: change tasks to not include unicode handling


jaywritescode:
I noticed in the `squeaky-clean` problem, there's a test as follows:
```java
//  test/java/SqueakyCleanTest.java

    @Test
    public void string_with_no_letters() {
        assertThat(SqueakyClean.clean("\uD83D\uDE00\uD83D\uDE00\uD83D\uDE00")).isEmpty();
    }
```
However, there's no corresponding [instruction](https://exercism.org/tracks/java/exercises/squeaky-clean) to remove "non-standard" characters from the input string, so the test suite defines a different spec than the instructions. 

I think the intent of the test is to remove any non-alphanumeric character or underscore from the input string, but I personally feel going too far into the details of Unicode (i.e. what is a "character" anyway?) distracts from the purpose of the exercise and can be discouraging. Perhaps the instructions can be clarified or the test can be removed or ignored. 

## Tasks

After some discussion (see comment below), an agreement was reached to modify the exercise to not include any unicode handling. Here are the tasks to do this:

- [ ] Update the current tasks and their examples to not include non-ascii characters.
- [ ] Change the tests to use non-ascii characters too.
- [ ] Remove the final task concerning greek letters.

## Contributing to this task

- If you'd like to contribute to this task, make a comment below saying that you'd like to work on this issue.
- After that, feel free to make a PR fixing the issue. Don't forget to [link the PR to this issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

njhanley:
This undocumented test is part of a larger issue: the stated goal, written tasks, hints, and tests all seem to disagree on what we're trying to accomplish. If the purpose of `clean` is to produce strings composed of zero or more letters and underscores, why don't we simply strip the other characters? What is the purpose of the replacements? Why is `isWhitespace` [recommended](https://github.com/exercism/java/blob/65faa744c9f9acb860f47e9241302b7b5241364f/exercises/concept/squeaky-clean/.docs/hints.md?plain=1#L9) when we're only instructed to replace spaces? Why remove Greek letters when `"àḃç"` is passed through unaltered?

Moreover, is it really a good idea to introduce Unicode support alongside `char`s without discussing supplementary characters, especially when there are tests containing surrogates? If "What is a character anyway?" is the question being asked, it isn't being adequately addressed by this exercise. In my opinion, that question is beyond the scope of simple `char` manipulation.

[...]

PR #2664, opened 2024-01-29
Updating Squeaky Clean

manumafe98:
Related issue: [#2049](https://github.com/exercism/java/issues/2049)

The goal of this PR is to update Squeaky clean to not ask the students to take into count control characters, unicode and greek letters.

Reviewer Resources:

### DIFF
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -10,11 +10,11 @@ In all cases the input string is guaranteed to be non-null. Note that the `clean
 
 ## 1. Replace any spaces encountered with underscores
 
-Implement the (_static_) `SqueakyClean.clean()` method to replace any spaces or numbers with underscores. This also applies to leading and trailing spaces.
+Implement the (_static_) `SqueakyClean.clean()` method to replace any spaces with underscores. This also applies to leading and trailing spaces.
 
 ```java
-SqueakyClean.clean("my   Id 1");
-// => "my___Id__"
+SqueakyClean.clean("my   Id");
+// => "my___Id"
 ```
 
 ## 2. Convert kebab-case to camelCase
@@ -26,7 +26,16 @@ SqueakyClean.clean("a-bc");
 // => "aBc"
 ```
 
-## 3. Omit characters that are not letters
+## 3. Convert leetspeak to normal text
+
+Modify the (_static_) `SqueakyClean.clean()` method to convert [leetspeak][leet-speak] to normal text.
+
+```java
+SqueakyClean.clean("H3ll0 W0rld");
+// => "Hello_World"
+```
+
+## 4. Omit characters that are not letters
 
 Modify the (_static_) `SqueakyClean.clean()` method to omit any characters that are not letters.
 
@@ -34,3 +43,5 @@ Modify the (_static_) `SqueakyClean.clean()` method to omit any characters that 
 SqueakyClean.clean("a$#.b");
 // => "ab"
 ```
+
+[leet-speak]: https://en.wikipedia.org/wiki/Leet
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -29,7 +29,7 @@ SqueakyClean.clean("a-bc");
 ## 3. Convert leetspeak to normal text
 
 Modify the (_static_) `SqueakyClean.clean()` method to convert [leetspeak][leet-speak] to normal text.
-For simplicity will only be using `4`, `3`, `0`, `1` and `7` from the table.
+For simplicity we will only be using `4`, `3`, `0`, `1` and `7` from the table.
 
 ```java
 SqueakyClean.clean("H3ll0 W0rld");
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -10,45 +10,27 @@ In all cases the input string is guaranteed to be non-null. Note that the `clean
 
 ## 1. Replace any spaces encountered with underscores
 
-Implement the (_static_) `SqueakyClean.clean()` method to replace any spaces with underscores. This also applies to leading and trailing spaces.
+Implement the (_static_) `SqueakyClean.clean()` method to replace any spaces or numbers with underscores. This also applies to leading and trailing spaces.
 
 ```java
-SqueakyClean.clean("my   Id");
-// => "my___Id"
+SqueakyClean.clean("my   Id 1");
+// => "my___Id__"
 ```
 
-## 2. Replace control characters with the upper case string "CTRL"
-
-Modify the (_static_) `SqueakyClean.clean()` method to replace control characters with the upper case string `"CTRL"`.
-
-```java
-SqueakyClean.clean("my\0Id");
-// => "myCTRLId",
-```
-
-## 3. Convert kebab-case to camelCase
+## 2. Convert kebab-case to camelCase
 
 Modify the (_static_) `SqueakyClean.clean()` method to convert kebab-case to camelCase.
 
 ```java
-SqueakyClean.clean("à-ḃç");
-// => "àḂç"
+SqueakyClean.clean("a-bc");
+// => "aBc"
 ```
 
-## 4. Omit characters that are not letters
+## 3. Omit characters that are not letters
 
 Modify the (_static_) `SqueakyClean.clean()` method to omit any characters that are not letters.
 
 ```java
-SqueakyClean.clean("a1😀2😀3😀b");
+SqueakyClean.clean("a$#.b");
 // => "ab"
-```
-
-## 5. Omit Greek lower case letters
-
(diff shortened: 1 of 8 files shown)


manumafe98:
I was thinking after the reformat, are we falling short with this concept exercise? do we need to add another point so students understand better? I was thinking on something to do is the string already contains an uppercase letter maybe.

manumafe98:
Also the pipeline seems to be failing because of the practice exercise ``strain`` which is weird

manumafe98:
> Perhaps if you want to add another task we could make the digits part a bit more interesting, since it's now grouped with converting whitespace. What about if we ask students to convert from [leetspeak](https://en.wikipedia.org/wiki/Leet) to normal text? So they would be required to change H3ll0 W0rld into Hello_World or something.

Uhh I really like this idea!! but also I think if we do this it should be with a couple of words, mostly the easy ones like 
````
e = 3
a = 4
o = 0
l = 1
t = 7
````

and probably no more than that.

**Comment by sanderploegsma** [comment]
Sounds good 👍

File: exercises/concept/squeaky-clean/.meta/src/reference/java/SqueakyClean.java [hunk 1]
```diff
@@ -1,11 +1,24 @@
...
+import java.util.HashMap;
+
 class SqueakyClean {
+    private static Map<Character, Character> leetSpeakMap = new HashMap<>() {{
+            put('3', 'e');
+            put('4', 'a');
+            put('0', 'o');
+            put('1', 'l');
+            put('7', 't');
+        }};
```
sanderploegsma, line 11: The only concern I have here is that this file is considered to be the "exemplar" solution for this exercise. It is shown to mentors with the remark that they should steer students towards an equivalent solution.

Using a `Map` here effectively means that the `maps` concept (which doesn't exist yet btw) becomes a prerequisite for this exercise. This would be a little much IMO.

Can we maybe dumb down the exemplar solution a bit by converting the map to a helper method instead? Something like this:

```java
class SqueakyClean {
    private static char replaceDigit(char digit) {
        if (digit == '3') return 'e';
        if (digit == '4') return 'a';
        // etc
    }
}
```

exercises/concept/squeaky-clean/.docs/instructions.md
```diff
@@ -26,11 +26,22 @@ SqueakyClean.clean("a-bc");
 // => "aBc"
 ```
 
-## 3. Omit characters that are not letters
+## 3. Convert leetspeak to normal text
+
+Modify the (_static_) `SqueakyClean.clean()` method to convert [leetspeak][leet-speak] to normal text.
```
sanderploegsma, line 31: I think we should clarify in the instructions exactly which numbers they should account for, since the Wiki page contains a _huge_ table.

exercises/concept/squeaky-clean/.docs/introduction.md
sanderploegsma, line 1: I believe this file is generated (notice the `introduction.md.tpl` file next to it). It takes its source from `concepts/chars/introduction.md`, so you should update that file instead and then re-generate this one using

```sh
bin/configlet generate
```

exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
```diff
@@ -43,50 +43,50 @@ public void leading_and_trailing_spaces() {
 
     @Test
     @Tag("task:2")
-    @DisplayName("The clean method replaces control characters with CTRL")
-    public void ctrl() {
-        assertThat(SqueakyClean.clean("my\0\r\u007FId")).isEqualTo("myCTRLCTRLCTRLId");
+    @DisplayName("The clean method converts kebab to camel case after removing a dash")
+    public void kebab_to_camel_case() {
+        assertThat(SqueakyClean.clean("à-ḃç")).isEqualTo("àḂç");
```
sanderploegsma, line 48: This test should probably use `a-bc` and `aBc` like in the exercise instructions.

manumafe98:
I added the new task, I'm not sure if I should add a note that says that we are not using all the numbers from leetspeak or if the students should realize this checking the tests.

exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
```diff
@@ -43,50 +43,50 @@ public void leading_and_trailing_spaces() {
...
+        assertThat(SqueakyClean.clean("a-C ")).isEqualTo("aC_");
     }
 
     @Test
-    @Tag("task:3")
-    @DisplayName("The clean method converts kebab to camel case after removing a dash")
-    public void kebab_to_camel_case() {
-        assertThat(SqueakyClean.clean("à-ḃç")).isEqualTo("àḂç");
+    @Tag("task:2")
+    @DisplayName("The clean method returns a string in camel case after removing a dash and replaces whitespaces")
```
sanderploegsma, line 60: The display name of this test and the one before are practically the same. Can we make them a little more distinguishable?

sanderploegsma (changes requested):
This is starting to look good!

exercises/concept/squeaky-clean/.docs/instructions.md
sanderploegsma: Ah you commented the same thing:

> I added the new task, I'm not sure if I should add a note that says that we are not using all the numbers from leetspeak or if the students should realize this checking the tests.

I think that in general it's fine for _practice_ exercises to require students to look at the tests in order to find out what is required of them. For _concept_ exercises however, I'd like to hold their hands a little more, as it's all about learning a new concept here. So I guess that is the judgement call I would make in most cases.

exercises/concept/squeaky-clean/.meta/src/reference/java/SqueakyClean.java
manumafe98: Done!

exercises/concept/squeaky-clean/.docs/instructions.md
manumafe98: > For concept exercises however, I'd like to hold their hands a little more, as it's all about learning a new concept here. So I guess that is the judgement call I would make in most cases.

Yeah I agree! we go for this one then

exercises/concept/squeaky-clean/.docs/introduction.md
manumafe98: Done!

exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
manumafe98: Done!

exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
manumafe98: Done!

exercises/concept/squeaky-clean/.docs/instructions.md
```diff
@@ -17,38 +17,32 @@ SqueakyClean.clean("my   Id");
...
+SqueakyClean.clean("a-bc");
+// => "aBc"
 ```
 
-## 3. Convert kebab-case to camelCase
+## 3. Convert leetspeak to normal text
 
-Modify the (_static_) `SqueakyClean.clean()` method to convert kebab-case to camelCase.
+Modify the (_static_) `SqueakyClean.clean()` method to convert [leetspeak][leet-speak] to normal text.
+For simplicity will only be using `4`, `3`, `0`, `1` and `7` from the table.
```
sanderploegsma, line 32: Small typo:

```suggestion
For simplicity we will only be using `4`, `3`, `0`, `1` and `7` from the table.
```

exercises/concept/squeaky-clean/src/test/java/SqueakyCleanTest.java
```diff
@@ -43,50 +43,50 @@ public void leading_and_trailing_spaces() {
...
+    public void kebab_to_camel_case_and_spaces() {
+        assertThat(SqueakyClean.clean(" hello-world ")).isEqualTo("_helloWorld_");
     }
 
     @Test
     @Tag("task:3")
-    @DisplayName("The clean method returns a string in camel case after removing a dash and a number")
-    public void kebab_to_camel_case_no_letter() {
-        assertThat(SqueakyClean.clean("a-1C")).isEqualTo("aC");
+    @DisplayName("The clean method converts leetspeak to normal text after reeplacing numbers with chars")
```
sanderploegsma, line 67: Another small typo
```suggestion
    @DisplayName("The clean method converts leetspeak to normal text after replacing numbers with chars")
```

sanderploegsma:
Also, since your changes to the exercise are substantial, feel free to add yourself as an author/contributor to the exercise's `.meta/config.json` 👍

sanderploegsma (approved):
🚀

sanderploegsma:
I just realized that we forgot to update the exercise's `.meta/design.md` file. We changed the scope of the exercise quite a bit and that should be reflected in the design file. Could you open another PR to address this?

manumafe98:
> I just realized that we forgot to update the exercise's `.meta/design.md` file. We changed the scope of the exercise quite a bit and that should be reflected in the design file. Could you open another PR to address this?

Oh yeah, we totally forgot

PR #2695, opened 2024-01-31
Adding analyzer feedback to design.md for bird watcher concept exercise

manumafe98:
# pull request

Related issue: [#2670](https://github.com/exercism/java/issues/2670)

Reviewer Resources:

### DIFF
--- a/exercises/concept/bird-watcher/.meta/design.md
+++ b/exercises/concept/bird-watcher/.meta/design.md
@@ -41,8 +41,12 @@ This exercise's prerequisites Concepts are:
 
 This exercise could benefit from the following rules in the [analyzer]:
 
-- `essential`: Verify that the solution does not hard-code the arrays used by the tests (`{2, 5, 0, 7, 4, 1 }`, `{0, 2, 5, 3, 7, 8, 4}`)
-- `actionable`: If the student did not use `clone` in the constructor to make a copy of the array, instruct them to do so.
-- `informative`: If the student used a different loop than a `For-Each` one in the methods `hasDayWithoutBirds` and `getBusyDays`, inform them that the solution can be improve using it.
+- `essential`: Verify that the solution does not hard-code the array passed in the constructor of the class `{2, 5, 0, 7, 4, 1 }`.
+- `essential`: The solution requires that the user uses at least once a `For` loop, the method `getCountForFirstDays()` could be a great place to do it.
+- `essential`: The solution requires that the user uses at least once a `For-Each` loop, the method `getBusyDays()` could be a great place to do it.
+- `actionable`: If the student did not use `clone` in the constructor to make a copy of the array, instruct them to do so. This is because if not, allows code outside the class to mutate the contents of the array.
+
+If the solution does not receive any of the above feedback, it must be exemplar.
+Leave a `celebratory` comment to celebrate the success!
 
 [analyzer]: https://github.com/exercism/java-analyzer
--- a/exercises/concept/bird-watcher/.meta/design.md
+++ b/exercises/concept/bird-watcher/.meta/design.md
@@ -36,3 +36,13 @@ This exercise's prerequisites Concepts are:
 - `classes`: know how to work with fields.
 - `booleans`: know what a `boolean` is.
 - `basics`: know how to work with `integers` and how to assign and update variables.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution does not hard-code the arrays used by the tests (`{2, 5, 0, 7, 4, 1 }`, `{0, 2, 5, 3, 7, 8, 4}`)
+- `actionable`: If the student did not use `clone` in the constructor to make a copy of the array, instruct them to do so.
+- `informative`: If the student used a different loop than a `For-Each` one in the methods `hasDayWithoutBirds` and `getBusyDays`, inform them that the solution can be improve using it.
+
+[analyzer]: https://github.com/exercism/java-analyzer
--- a/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java
+++ b/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java
@@ -7,7 +7,7 @@ class BirdWatcher {
     }
 
     public int[] getLastWeek() {
-        return birdsPerDay.clone();
+        return new int[] { 0, 2, 5, 3, 7, 8, 4 };
     }
 
     public int getToday() {
@@ -15,7 +15,7 @@ class BirdWatcher {
     }
 
     public void incrementTodaysCount() {
-        birdsPerDay[birdsPerDay.length - 1] = birdsPerDay[birdsPerDay.length - 1] + 1;
+        birdsPerDay[birdsPerDay.length - 1]++;
     }
 
     public boolean hasDayWithoutBirds() {


manumafe98:
I have a couple of doubts on the reference resolution:

https://github.com/exercism/java/blob/02d488e7ad4c6bed91b277006753ab5a14fa8585/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java#L10

The clone in the method `getLastWeek` is necessary? Because it's already making the copy in the constructor.

https://github.com/exercism/java/blob/02d488e7ad4c6bed91b277006753ab5a14fa8585/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java#L18

Also I was thinking that this could be simplified to 

````java
    public void incrementTodaysCount() {
        birdsPerDay[birdsPerDay.length - 1]++;
    }
````

But this one is more peaky.

**Comment by sanderploegsma** [comment]
> I have a couple of doubts on the reference resolution:
> 
> https://github.com/exercism/java/blob/02d488e7ad4c6bed91b277006753ab5a14fa8585/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java#L10
> 
> The clone in the method `getLastWeek` is necessary? Because it's already making the copy in the constructor.

I think the author added the `.clone()` here because the array is a reference value, and by returning a reference to the field, the caller could make changes to it:

```java
birdWatcher.getLastWeek()[0] = 0;
```

Adding the `.clone()` makes the method return a copy of the array to prevent this type of mis-use.

> https://github.com/exercism/java/blob/02d488e7ad4c6bed91b277006753ab5a14fa8585/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java#L18
> 
> Also I was thinking that this could be simplified to
> 
> ```java
>     public void incrementTodaysCount() {
>         birdsPerDay[birdsPerDay.length - 1]++;
>     }
> ```
> 
> But this one is more peaky.

Yeah I think you're right. The exercise's `.meta/config.json` indicates that this exercise was forked from the C# track, and they seem to be doing this too: https://github.com/exercism/csharp/blob/main/exercises/concept/bird-watcher/.meta/Exemplar.cs.

sanderploegsma:
Reading the exercise instructions again, I think that the implementation of `getLastWeek()` in the exemplar solution is actually wrong!

> For comparison purposes, you always keep a copy of last week's counts nearby, which were: 0, 2, 5, 3, 7, 8 and 4. Implement the `BirdWatcher.getLastWeek()` method that returns last week's counts:
>
> ```java
> BirdWatcher.getLastWeek();
> // => [0, 2, 5, 3, 7, 8, 4]
> ```

From this I conclude that the solution must _always_ return those values. The array passed to the constructor contains the values for _this_ week, not _last_ week. So the exemplar solution should implement `getLastWeek()` similar to the C# track: https://github.com/exercism/csharp/blob/63dd1aa3fe0bd2b59033158f243d096bd556b9f4/exercises/concept/bird-watcher/.meta/Exemplar.cs#L14

exercises/concept/bird-watcher/.meta/design.md
```diff
@@ -36,3 +36,13 @@ This exercise's prerequisites Concepts are:
 - `classes`: know how to work with fields.
 - `booleans`: know what a `boolean` is.
 - `basics`: know how to work with `integers` and how to assign and update variables.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution does not hard-code the arrays used by the tests (`{2, 5, 0, 7, 4, 1 }`, `{0, 2, 5, 3, 7, 8, 4}`)
+- `actionable`: If the student did not use `clone` in the constructor to make a copy of the array, instruct them to do so.
```
sanderploegsma, line 45: Maybe add a note on explaining why this is preferred: it's because arrays are passed by reference, so not using `clone` in the constructor allows code outside the class to mutate the contents of the array.

exercises/concept/bird-watcher/.meta/design.md
```diff
@@ -36,3 +36,13 @@ This exercise's prerequisites Concepts are:
...
+- `informative`: If the student used a different loop than a `For-Each` one in the methods `hasDayWithoutBirds` and `getBusyDays`, inform them that the solution can be improve using it.
```
sanderploegsma, line 46: Actually I think it's reasonable to require the solution to contain at least one `For` loop and one `For-Each` loop, given the scope of the exercise. So I would add two `essential` feedback points:

- One to verify that the solution uses at least one `For` loop. If not, instruct them to use at least one, and mention that `getCountForFirstDays()` might be a good place to use it.
- One to verify that the solution uses at least one `For-Each` loop. If not, instruct them to use at least one, and mention that `getBusyDays()` might be a good place to use it.

exercises/concept/bird-watcher/.meta/design.md
```diff
@@ -36,3 +36,13 @@ This exercise's prerequisites Concepts are:
...
+
```
sanderploegsma, line 47: I think that for concept exercises it's a good idea to celebrate the fact that a student submitted a solution that receives no other feedback on the implementation, because they did a good job in that case. See https://github.com/exercism/java/pull/2696

sanderploegsma (changes requested):

exercises/concept/bird-watcher/.meta/design.md
manumafe98: Done!

exercises/concept/bird-watcher/.meta/design.md
manumafe98: Done!

exercises/concept/bird-watcher/.meta/design.md
manumafe98: Done!

sanderploegsma (approved):
:shipit:

exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java
```diff
@@ -7,15 +7,15 @@ public BirdWatcher(int[] birdsPerDay) {
     }
 
     public int[] getLastWeek() {
-        return birdsPerDay.clone();
+        return new int[] { 0, 2, 5, 3, 7, 8, 4 };
```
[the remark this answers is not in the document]
phoenix-1729: Why should the test cases fail?(Can you please elaborate a little bit)
The array which getLastWeek() is returning is the same array which is being passed while creating the object of Birdwatcher.

`private static final int DAY1 = 0;
    private static final int DAY2 = 2;
    private static final int DAY3 = 5;
    private static final int DAY4 = 3;
    private static final int DAY5 = 7;
    private static final int DAY6 = 8;
    private static final int TODAY = 4;

    private BirdWatcher birdWatcher;
    private int lastWeek[] = {DAY1, DAY2, DAY3, DAY4, DAY5, DAY6, TODAY};

    @BeforeEach
    public void setUp() {
        birdWatcher = new BirdWatcher(lastWeek);
    }

    @Test
    @Tag("task:1")
    @DisplayName("The getLastWeek method correctly returns last week's counts")
    public void itTestGetLastWeek() {
        assertThat(birdWatcher.getLastWeek())
            .containsExactly(DAY1, DAY2, DAY3, DAY4, DAY5, DAY6, TODAY);
    }`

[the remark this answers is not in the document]
manumafe98: Hey @phoenix-1729 I recommend to ask the questions in the issue itself

PR #2700, opened 2024-01-31
Add analyzer feedback for Annalyns Infiltration

### LINKED ISSUE #2669: annalyns-infiltration: describe Analyzer feedback in `.meta/design.md`
The `.meta/design.md` file of each concept exercise should contain a section describing what feedback the [analyzer](https://github.com/exercism/java-analyzer) should provide and when. This is a prerequisite for implementing the analyzer for each of these exercises.

The section should clearly describe *what* should be checked by the analyzer, what feedback should be given to the student and the *type* of each comment (`essential`, `actionable`, `informative` or `celebratory`, see [the documentation](https://exercism.org/docs/building/tooling/analyzers/interface#h-type-optional)).

Here is a good example of what it might look like: https://github.com/exercism/javascript/blob/main/exercises/concept/mixed-juices/.meta/design.md

When writing the analyzer section for this concept exercise, make sure to keep the following guidelines in mind: https://exercism.org/docs/building/tooling/analyzers/comments

kahgoh:
# pull request

closes #2669

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/annalyns-infiltration/.meta/design.md
+++ b/exercises/concept/annalyns-infiltration/.meta/design.md
@@ -23,17 +23,19 @@ Nothing to report
 This exercise could benefit from the following rules in the [analyzer]:
 
 - `essential`: If student returns a boolean literal, tell them it is possible to directly return the result of a expression. For example:
-                ```java
-                // instead of
-                if (knightIsAwake) {
-                    return true;
-                } else {
-                    return false;
-                }
 
-                // ... return the expression directly
-                return knightIsAwake;
-                ```
+  ```java
+  // instead of
+  if (knightIsAwake) {
+    return true;
+  } else {
+    return false;
+  }
+
+  // ... return the expression directly
+  return knightIsAwake;
+  ```
+
 - `essential`: If the student compares a boolean variable with a boolean literal (e.g. `knightIsAwake == true` or `archerIsAwake == false`), tell them this can be simplified to just the variables (e.g. `knightIsAwake` or `archerIsAwake`).
 - `essential`: If the student uses an `if` statement or the ternary operator, tell them this exercise was to explore booleans and boolean operators and this exercise can be solved without them.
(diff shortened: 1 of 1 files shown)


manumafe98 (changes requested):
Checking on the Javascript analyzer [approach](https://github.com/exercism/javascript/blob/main/exercises/concept/annalyns-infiltration/.meta/design.md#analyzer) I would say that we could be more explicit on some actionables, I like the topics already created, but maybe we could add a couple more, I also would like your opinion on this one @sanderploegsma 

The if statement in the method seem to be treated as essential as  well, check [this](https://github.com/exercism/javascript-analyzer/blob/main/src/analyzers/concept/annalyns-infiltration/index.ts)

sanderploegsma:
I think that given the scope of the concept and the exercise, we should steer students towards a solution that _only_ uses boolean operators (`&&`, `||` and `!`). If a solution uses if-statements, ternary expressions or comparison operators and boolean literals, I think the analyzer comments should be of type `essential`.

In terms of additional comments, I can think of only one thing, and that is unnecessary parentheses in the final method. If there is a consistent way to check this, could be an `informative` comment if the rest of the solution doesn't receive any feedback.

kahgoh:
Making them essential sounds good to me ;) 

Thanks for pointing me to the Javascript analyzer @manumafe98. I've tried taking some inspiration from there for the wording and added something about the unnecessary parentheses.

manumafe98:
> Thanks for pointing me to the Javascript analyzer @manumafe98. I've tried taking some inspiration from there for the wording and added something about the unnecessary parentheses.

Np! this is already looking great!

exercises/concept/annalyns-infiltration/.meta/design.md
```diff
@@ -17,3 +17,28 @@ Nothing to report
...
+                // instead of
+                if (knightIsAwake) {
+                    return true;
+                } else {
+                    return false;
+                }
+
+                // ... return the expression directly
+                return knightIsAwake;
+                ```
```
sanderploegsma, line 36: This snippet is not rendering correctly: https://github.com/exercism/java/blob/c6c6889a24742828d4904b6b8284181359cbc29c/exercises/concept/annalyns-infiltration/.meta/design.md

Try indenting it with 2 spaces and leave a blank line before and after it. That should make it part of the list item and fix the rendering.

manumafe98 (approved):

PR #2702, opened 2024-01-31
Adding analyzer feedback for need for speed concept exercise

manumafe98:
# pull request

Related issue: [#2682](https://github.com/exercism/java/issues/2682)

As the exercise does not have loops as a prerequisites, and we do not have a concept that introduces to while loops I made some slight changes to the reference resolution as well

Also I got my inspiration for the [go analyzer](https://github.com/exercism/go/blob/main/exercises/concept/need-for-speed/.meta/design.md#analyzer)

Reviewer Resources:

### DIFF
--- a/exercises/concept/need-for-speed/.meta/design.md
+++ b/exercises/concept/need-for-speed/.meta/design.md
@@ -17,6 +17,7 @@
 - Inheritance.
 - Finalizers.
 - Method overloading.
+- Loops.
 
 ## Concepts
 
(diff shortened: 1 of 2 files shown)


sanderploegsma:
I'm reading through the original `.meta/design.md` and it seems to be a copy of the one from `elons-toy-car`. It stood out because it mentions that constructors are out-of-scope, while the exercise teaches the concept of a constructor. 😂

So I would suggest updating the entire file in this PR to address this.

exercises/concept/need-for-speed/.meta/design.md
```diff
@@ -17,6 +17,7 @@
 - Inheritance.
 - Finalizers.
 - Method overloading.
+- Loops.
```
sanderploegsma, line 20: I don't think we need to list this here. The out-of-scope section is typically used to clarify which parts of the concept being taught are considered out of scope for the exercise. Loops are not really part of the constructors concept, so no need to mention it.

manumafe98:
> I'm reading through the original `.meta/design.md` and it seems to be a copy of the one from `elons-toy-car`. It stood out because it mentions that constructors are out-of-scope, while the exercise teaches the concept of a constructor. 😂
> 
> So I would suggest updating the entire file in this PR to address this.

Oh, good catch, okey I would update the entire file then

exercises/concept/need-for-speed/.meta/design.md
manumafe98: 👍

sanderploegsma (approved):
👍

Issue #2642, opened 2024-01-14
Dominoes java track exercise violates equals x hashcode.


abmpicoli:
The code for Domino.java has a violation between equals x hashcode code.

The code for equals accepts rotated tiles: that is , a Domino where left=3 and right=1 equals to a Domino where left=1 and right=3.

However the hashcode is a hash of left,right: this means that you can have a class that equals, and have a different hashcode.

sanderploegsma:
Thanks for pointing this out! You are right of course, and it's definitely something we should address since we should be setting an example.

PRs welcome!

abmpicoli:
Created pull request with the solution https://github.com/exercism/java/pull/2704

PR #2704, opened 2024-02-01
Issue #2642 solution - hashcode for dominoes.

abmpicoli:
* Add a comment in the Domino class making explicit that changes in the Domino class are not accepted by the solution. (Believe me,
I was really surprised during my first submission that my changes in the Domino class were rejected. As well as my mentee ).

* Change the way equals and hashcode are implemented to use the min+max value of each side instead of left+right.

* Added a constraint validation within the Domino class constructor, that tiles must
  have a maximum number of 9 on each side (this is needed for the hash implementation).

* In the equals code, a comparison between a Domino and an Object would generate a
  class cast exception instead of the expected value, false.

# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/dominoes/.meta/src/reference/java/Domino.java
+++ b/exercises/practice/dominoes/.meta/src/reference/java/Domino.java
@@ -27,11 +27,11 @@ class Domino {
     
     @Override
     public boolean equals(Object o) {
-    if (!o instanceof Domino ) {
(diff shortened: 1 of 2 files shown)


abmpicoli:
When you code that last minute thing and forget to test it... Boom! :D Sorry about the previous commit.

sanderploegsma (approved):
LGTM, thanks for fixing!

PR #2716, opened 2024-02-02
`pop-count`: rename to `eliuds-eggs`

ErikSchierboom:
_We are renaming the `pop-count` exercise to `eliuds-eggs`. See https://github.com/exercism/problem-specifications/pull/2382 for more information._

---

This PR renames `pop-count` to `eliuds-eggs`. We have attempted to rename everything that is automatable, but you may need to make specific changes in the files.

If you could prioritise finishing and merging this, we would be grateful, so we can get the exercise in the correct state across Exercism. Please let me know if you have any questions.

Thanks! 🙂


manumafe98:
I guess it doesn't matter that the source url used to be called pop count right?
https://github.com/exercism/java/blob/de5908615c21193a37d1ca2c416f7cdac75e6c31/exercises/practice/pop-count/.meta/config.json#L21

sanderploegsma (approved):
Good to go!

PR #2722, opened 2024-02-05
Adding analyzer feedback for karls language concept exercise

manumafe98:
# pull request

closes [#2678](https://github.com/exercism/java/issues/2678)

To be consistent with lasagna design.md, I made a couple of changes, so instead of using a loop make the user reuse code seem better for me.

Reviewer Resources:

### DIFF
--- a/exercises/concept/karls-languages/.docs/hints.md
+++ b/exercises/concept/karls-languages/.docs/hints.md
@@ -2,37 +2,32 @@
 
 ## 1. Define a function to check if the language list is empty
 
-- Try using the [`isEmpty()`][isempty] method.
+- One of the [methods][list] on the List type can be used to check if it is empty.
 
 ## 2. Define a function to add a language to the list
 
-- Try using the [`add(E element)`][add] method.
+- One of the [methods][list] on the List type can be used to add elements.
 - Reminder: methods that return `void` do not need any `return` statements.
 
 ## 3. Define a function to remove a language from the list
 
-- Try using the [`remove(Object o)`][remove] method.
+- One of the [methods][list] on the List type can be used to remove elements.
 - Reminder: methods that return `void` do not need any `return` statements.
 
 ## 4. Define a function to return the first item in the list
 
-- Try using the [`get(int index)`][get] method.
+- One of the [methods][list] on the List type can be used to get elements on a certain index.
 
 ## 5. Define a function to return how many languages are in the list
 
-- Try using the [`size()`][size] method.
+- One of the [methods][list] on the List type can be used to get the size of the list.
 
 ## 6. Define a function to determine if a language is in the list
 
-- Try using the [`contains(Object o)`][contains] method.
+- One of the [methods][list] on the List type can be used to check if an element is contained in the list.
 
 ## 7. Define a function to determine if the list is exciting
 
-- Try using the `containsLanguage` method from the previous step.
+- Try using a method already defined in the class.
 
-[isempty]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#isEmpty()
-[add]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#add(E)
-[remove]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#remove(java.lang.Object)
-[get]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#get(int)
-[size]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#size()
-[contains]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#contains(java.lang.Object)
+[list]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html
--- a/exercises/concept/karls-languages/.docs/hints.md
+++ b/exercises/concept/karls-languages/.docs/hints.md
@@ -2,31 +2,37 @@
 
 ## 1. Define a function to check if the language list is empty
 
-- Try using the [`isEmpty()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#isEmpty()) method.
+- Try using the [`isEmpty()`][isempty] method.
 
 ## 2. Define a function to add a language to the list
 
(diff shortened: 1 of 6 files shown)


manumafe98:
I have also a couple of questions:
- maybe the method called ``isEmpty`` is literally giving the solution to the student, maybe we should rename it?
- Also seeing the hints file, and checking other files we are using documentation of different versions on java, we don't have a standard, I was thinking that we could maybe create a variable and assign it to the latest version, so the documentation that the student has is always updated? I'm open to ideas

exercises/concept/karls-languages/.docs/hints.md
sanderploegsma, line 1: I think these hints may give away a bit too much. Based on [the documentation](https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-hints-md) I'd say they could be just a little bit more cryptic. For example, instead of

> Try using the `isEmpty()` method.

it might say

> One of the [methods on the `List` type](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html) can be used to check if it is empty.

Same for the other hints.

exercises/concept/karls-languages/.meta/design.md
```diff
@@ -43,6 +43,19 @@ This Concepts Exercise's Concepts are:
...
-- `for-loops`: know how to use a for-loop to iterate over a collection.
 - `arrays`: know of the array collection type and that it has a fixed length.
 - `strings`: data types used in this exercise
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not reuse the implementation of the `containsLanguage` method in the `isExciting` method, instruct them to do so.
+  Explain that reusing existing code instead of copy-pasting can help make code easier to maintain.
```
sanderploegsma, line 54: I'm not sure that this is essential for this exercise. Although it's good practice to reuse code as much as possible, that's not the goal of the exercise. I'd change this to `actionable`.

manumafe98:
If we 100% want to make ``for-loops`` a prerequisite we could replace the last method  ``isExciting`` that currently for me seems pointless for something like this

````java
    public <T> String isInstanceOf(List<T> list) {
        for (T element : list) {
            if (element instanceof Integer) {
                return "Integer";
            }
        }
        return "String";
    }
````
And introduce the concept ``instanceof``

File: exercises/concept/karls-languages/.meta/design.md [hunk 2]
```diff
@@ -43,6 +43,19 @@ This Concepts Exercise's Concepts are:
...
+- `actionable`: If the solution did not use `contains` in the method `containsLanguage`, instruct the student to do so.
+- `actionable`: If the solution did not use `isEmpty` in the method `isEmpty`, instruct the student to do so.
+
```
sanderploegsma, line 57: Based on some of the solution sets I found in the community solutions we may want to add some (`informative`?) feedback on using

```java
return languages.contains(language);
```

instead of

```java
if (languages.contains(language) {
    return true;
}
return false;
```

exercises/concept/karls-languages/.meta/design.md
sanderploegsma: Actually I see that most community solutions use this:

```java
class LanguageList {
    boolean isExciting() {
        return languages.contains("Java") || languages.contains("Kotlin");
    }
}
```

In this case I would even go as far as saying that reusing `containsLanguage` instead of `language.contains()` becomes subjective enough that the comment might be `informative`...

sanderploegsma:
> Also seeing the hints file, and checking other files we are using documentation of different versions on java, we don't have a standard, I was thinking that we could maybe create a variable and assign it to the latest version, so the documentation that the student has is always updated? I'm open to ideas

How would that work exactly? AFAIK we can't define variables in Markdown files.

sanderploegsma:
> If we 100% want to make `for-loops` a prerequisite we could replace the last method `isExciting` that currently for me seems pointless for something like this
> 
> ```java
>     public <T> String isInstanceOf(List<T> list) {
>         for (T element : list) {
>             if (element instanceof Integer) {
>                 return "Integer";
>             }
>         }
>         return "String";
>     }
> ```
> 
> And introduce the concept `instanceof`

Nah I think removing the `for-loops` prerequisite is the way to go here, the concept is really not needed for this exercise. Also based on the community solutions almost nobody uses a loop for the last task.

exercises/concept/karls-languages/.docs/hints.md
manumafe98: Done!

exercises/concept/karls-languages/.meta/design.md
manumafe98: Percect, I agree with this change!

exercises/concept/karls-languages/.meta/design.md
manumafe98: Done!

manumafe98:
> How would that work exactly? AFAIK we can't define variables in Markdown files.

Yeah, I thought that the variables accept plain text, but it seems that they only accept URLs. However, I still think we should find a way to handle this, because currently, we have a ton of documentation for different Java versions – 7, 8, 9, 11, 12, 13, 14, etc. Perhaps a policy to only add documentation for the latest current version, but that wouldn't eliminate the need to update the older ones at some point.

sanderploegsma:
> > How would that work exactly? AFAIK we can't define variables in Markdown files.
> 
> 
> 
> Yeah, I thought that the variables accept plain text, but it seems that they only accept URLs. However, I still think we should find a way to handle this, because currently, we have a ton of documentation for different Java versions – 7, 8, 9, 11, 12, 13, 14, etc. Perhaps a policy to only add documentation for the latest current version, but that wouldn't eliminate the need to update the older ones at some point.

Can you add a new issue to discuss this?

exercises/concept/karls-languages/.meta/src/reference/java/LanguageList.java
```diff
@@ -4,7 +4,7 @@
 public class LanguageList {
     private final List<String> languages = new ArrayList<>();
 
-    public boolean isEmpty() {
+    public boolean checkIfEmpty() {
```
sanderploegsma, line 7: One doubt I have about this change is that it is conventional to name boolean methods `isX`. Such conventions are best taught by example, so I'm not sure I see the benefit of renaming this method.

manumafe98: Ok, I will rollback that change then

manumafe98:
> Can you add a new issue to discuss this?

Sure

sanderploegsma (approved):

Issue #2718, opened 2024-02-02
Wizards and Warriors exercise reformat 


manumafe98:
As this concept exercises introduces to the topic of ``inheritance`` for me, seems a bit weird to give the student the class with the inherited parent class already in the code.
I was thinking that we can create a tasks that ask the student to implement the classes ``Wizard`` and ``Warrior`` and then that they extend also this classes with the parent class ``Fighter``.
Maybe as well give a brief introduction on what's ``Override`` doing in the code.

PR #2729, opened 2024-02-08
Reformat concept exercise Wizards and Warriors

manumafe98:
# pull request

closes [#2718](https://github.com/exercism/java/issues/2718)

The goal is to update the concept exercise so the student is not directly handled the code and has to work with it a bit more.

Reviewer Resources:


manumafe98:
@sanderploegsma I would like your feedback here, I'm not sure why does tests are failing, locally when I run ``gradle check```all seem to be working ok

sanderploegsma:
The reason the tests are failing is because we made the decision in the past that every stub solution for every exercise should compile cleanly with the tests, so that students can use the test suite from the moment they start instead of having to diagnose compilation errors.

These changes break that setup; the tests can't compile if the necessary classes are not implemented yet.

If you feel strongly that this exercise should be refactored, we need to think about another way to make sure the tests can run on the starter solution. One way of doing so could be to use reflection to invoke the methods being tested, like in the `lasagna` exercise.

manumafe98:
> If you feel strongly that this exercise should be refactored, we need to think about another way to make sure the tests can run on the starter solution. One way of doing so could be to use reflection to invoke the methods being tested, like in the `lasagna` exercise.

Well I definitely think this need a reformat, what are your thoughts about the change?

Maybe we could remove the first task, the one that the user creates the class, and only leave the second one that the user extends with the parent class? Or that will fail as well?

sanderploegsma:
Well you would need to change the tests a bit because statements like these still won't compile because a `Warrior` cannot be assigned to a variable of type `Fighter`:

https://github.com/exercism/java/blob/89308ce1d646f3de5316d4a04f45d25701b2166a/exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java#L51

Also, if you don't predefine the `isVulnerable()` and `damagePoints()` methods on both classes, the later tests won't be able to compile either.

sanderploegsma:
I tend to agree with your opinion on this exercise though, I think that the starter solution is too complete and would like for the student to have to do a little bit of work to get the inheritance part working.

I ran into the same thing when adding the `logs-logs-logs` exercise, where the tests need to check for specific enum cases but I wanted the student to have to add those themselves, so i couldn't reference them directly. In that case it was pretty easy to solve this with reflection. And we chose to reuse the approach from `lasagna` in the `wizards-and-warriors-2` exercise, again because we wanted the student to define the methods themselves.

concepts/inheritance/introduction.md
```diff
@@ -25,13 +25,19 @@ Consider an animal named `Lion`, having a class like,
...
 
+    @Override
     public void bark() {
         System.out.println("Lion here!!");
     }
 
 }
 ```
 
+```exercism/note
```
sanderploegsma, line 36: I believe the admonitions should be prefixed by `~~~~` instead of three backtics

exercises/concept/wizards-and-warriors/.docs/hints.md
```diff
@@ -2,41 +2,51 @@
...
-Detailed explanation of inheritance can be found at [Inheritance][inheritance-main].
+Detailed explanation of inheritance can be found at [Inheritance][inheritance-concept].
 The whole inheritance concept has a lot to do with the concepts around [overriding][java-overriding].
 
-## 1. Describe a Fighter
+## 1. Create the Warrior and Wizard classes
+
+- Review the [class][class-concept] concept.
+
+## 2. Extend Wizard and Warrior classes with Figther class
```
sanderploegsma, line 12: Small typo:

```suggestion
## 2. Extend Wizard and Warrior classes with Fighter class
```

exercises/concept/wizards-and-warriors/.docs/hints.md
```diff
@@ -2,41 +2,51 @@
...
 ## General
 
-Detailed explanation of inheritance can be found at [Inheritance][inheritance-main].
+Detailed explanation of inheritance can be found at [Inheritance][inheritance-concept].
 The whole inheritance concept has a lot to do with the concepts around [overriding][java-overriding].
 
-## 1. Describe a Fighter
+## 1. Create the Warrior and Wizard classes
+
+- Review the [class][class-concept] concept.
```
sanderploegsma, line 10: This might be a good place to use [widgets](https://exercism.org/docs/building/markdown/widgets) to link to the concept on the website

exercises/concept/wizards-and-warriors/.docs/instructions.md
```diff
@@ -16,19 +16,39 @@ For a Wizard, these are the rules:
...
+
+## 2. Extend Wizard and Warrior classes with Figther class
+
+Use the `extends` keyword to inherit the `Fighter` class.
+
+```java
+public class Warrior extends Fighter {
+    
+}
+```
```
sanderploegsma, line 39: This instruction is a bit too verbose, it literally shows what to type into your editor. I suggest instructing _what_ to do, not _how_ to do it.

We want the student to make the `Warrior` and `Wizard` classes to inherit from the `Fighter` class.

manumafe98:
> I tend to agree with your opinion on this exercise though, I think that the starter solution is too complete and would like for the student to have to do a little bit of work to get the inheritance part working.
> 
> I ran into the same thing when adding the `logs-logs-logs` exercise, where the tests need to check for specific enum cases but I wanted the student to have to add those themselves, so i couldn't reference them directly. In that case it was pretty easy to solve this with reflection. And we chose to reuse the approach from `lasagna` in the `wizards-and-warriors-2` exercise, again because we wanted the student to define the methods themselves.

Great, thanks for the feedback, then I would give it a try with `reflection` and those two exercises as an example to make it work.

exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
```diff
@@ -40,15 +78,15 @@ public void testWizardVulnerable() {
...
+    @Tag("task:6")
     @DisplayName("The isVulnerable method of the Wizard returns true by default")
     public void testWizardVulnerableByDefault() {
         Wizard wizard = new Wizard();
         assertThat(wizard.isVulnerable()).isTrue();
     }
 
     @Test
-    @Tag("task:4")
+    @Tag("task:6")
```
sanderploegsma, line 89: Looks like the tests were incorrectly tagged, starting with this one. It seems to test task 7 (previously task 5). The next three tests should be tagged with 7, 8, 8 respectively.

sanderploegsma (changes requested):

concepts/inheritance/introduction.md
manumafe98: Just in case I copied it from https://github.com/exercism/java/blob/832c70349b3ce74f36f046cd88c6a0ffacff3452/concepts/datetime/introduction.md?plain=1#L37

Maybe is wrong there as well?

exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
manumafe98: Done!

exercises/concept/wizards-and-warriors/.docs/hints.md
manumafe98: Done!

concepts/inheritance/introduction.md
manumafe98: Done!

exercises/concept/wizards-and-warriors/.docs/instructions.md
manumafe98: Done!

exercises/concept/wizards-and-warriors/.docs/hints.md
manumafe98: Seems like the linter check does not like the widgets or I'm doing something wrong

manumafe98: > The link reference, i.e. the contents of (), is ignored by the website. You can point them at whatever you want for maintenance purposes.

Oh yeah, It's because it should point to somewhere, where do you think I should point it ?

sanderploegsma: I'd say point it to the source on GitHub, so https://github.com/exercism/java/tree/main/concepts/classes

concepts/inheritance/introduction.md
sanderploegsma: Hmm thanks for pointing that out. I double checked the [page on the website](https://exercism.org/tracks/java/concepts/datetime) and it seems to render correctly with three backticks as well. Still, the [documentation](https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions) says to use `~~~~` so let's follow that.

exercises/concept/wizards-and-warriors/.docs/instructions.md
sanderploegsma: I think we should not include the code snippets either

sanderploegsma:
Perhaps the exercise could benefit from a bit more restructuring: what if we reorder them so that the student first implements the `Warrior` class fully, then followed by the `Wizard` class? So something like this:

1. Create the Warrior type (this includes inheriting from `Fighter`)
2. Describe a Warrior (this covers overriding the `toString()` method on the `Warrior` class)
3. Calculate the damage dealt by a Warrior (this covers overriding the `damagePoints()` method on the `Warrior` class)
4. Create the Wizard type (this includes inheriting from `Fighter`)
5. Describe a Wizard (this covers overriding the `toString()` on the `Wizard` class)
6. Allow a Wizard to prepare a spell (this covers adding the `prepareSpell()` method on the `Wizard` class)
7. Make a Wizard vulnerable unless they prepared a spell (this covers overriding the `isVulnerable()` method on the `Wizard` class)
8. Calculate the damage dealt by a Wizard (this covers overriding the `damagePoints()` method on the `Wizard` class)

sanderploegsma:
Also there is one problem with the current design: Because the `Fighter.damagePoints` method is `abstract`, students can't really finish the second step without also implementing the `damagePoints` method. If they don't, their solution won't compile.

Maybe abstract classes deserve their own concept, especially since the `inheritance` introduction does not even mention what they are. If we then make the `Fighter` type non-abstract and provide a default implementation of the `Fighter.damagePoints` method the above issue will be solved.

I have the following in mind:

```java
class Fighter {
    boolean isVulnerable() {
        return true;
    }

    int getDamagePoints(Fighter target) {
        return 1;
    }
}
```

Note:
- I renamed `damagePoints` to `getDamagePoints` since the former seems to indicate that an action is taken while there is not
- I made the `isVulnerable` return `true` by default because that makes a bit more sense in the context of the exercise:
  - A `Warrior` is never vulnerable
  - A `Wizard` is sometimes vulnerable
  - Any other `Fighter` is vulnerable by default

exercises/concept/wizards-and-warriors/.docs/instructions.md
manumafe98: Oh I thought we should always add them, I will remove them then

manumafe98:
> Perhaps the exercise could benefit from a bit more restructuring: what if we reorder them so that the student first implements the `Warrior` class fully, then followed by the `Wizard` class? So something like this:
> 
> 1. Create the Warrior type (this includes inheriting from `Fighter`)
> 2. Describe a Warrior (this covers overriding the `toString()` method on the `Warrior` class)
> 3. Calculate the damage dealt by a Warrior (this covers overriding the `damagePoints()` method on the `Warrior` class)
> 4. Create the Wizard type (this includes inheriting from `Fighter`)
> 5. Describe a Wizard (this covers overriding the `toString()` on the `Wizard` class)
> 6. Allow a Wizard to prepare a spell (this covers adding the `prepareSpell()` method on the `Wizard` class)
> 7. Make a Wizard vulnerable unless they prepared a spell (this covers overriding the `isVulnerable()` method on the `Wizard` class)
> 8. Calculate the damage dealt by a Wizard (this covers overriding the `damagePoints()` method on the `Wizard` class)

I like this I added an extra task, to make the warrior invulnerable

manumafe98:
@sanderploegsma I will wait for your review before applying the ``reflection proxy`` just in case

sanderploegsma:
I think I want to have another look at the copy in the instructions.md, but other than that it all looks pretty neat so far. I suggest finishing this PR so that all checks pass again, and then we can do some finishing touches on the copy.

exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
```diff
@@ -1,87 +1,217 @@
...
+    }
+
+    @Test
+    @Tag("task:7")
+    @DisplayName("The prepareSpell method makes the Wizard not vulnerable")
+    void testWizardVulnerable() {
         wizard.prepareSpell();
-        assertThat(warrior.damagePoints(wizard)).isEqualTo(6);
+        assertThat(wizard.isVulnerable()).isFalse();
+    }
```
sanderploegsma, line 156: Hmm it's a bit strange that this test requires the `isVulnerable` method to be implemented correctly, but that is part of the next task. Maybe tasks 7 and 8 should be merged into one? So adding the `prepareSpell()` method and overriding the `isVulnerable()` method are bundled into a single task? This might make sense as they are very much related.

manumafe98:
@sanderploegsma I was thinking, should we add [TODO's](https://github.com/exercism/java/blob/main/exercises/concept/lasagna/src/main/java/Lasagna.java) like the lasagna and wizard-and-warriors-2 to the given code to the student

exercises/concept/wizards-and-warriors/src/test/java/FighterTest.java
manumafe98: Sure, I agree!

manumafe98: Done!

sanderploegsma:
@manumafe98 I pushed some extra commits to clean up the exercise a bit more. Let me know what you think of the changes, IMO it should be good to go now.

sanderploegsma (approved):

manumafe98:
> @manumafe98 I pushed some extra commits to clean up the exercise a bit more. Let me know what you think of the changes, IMO it should be good to go now.

Yep I would say that looks great as well! Now seems more like a interactive concept exercise for a student!

PR #2731, opened 2024-02-16
Adding analyzer feedback for wizard-and-warriors concept exercise

manumafe98:
# pull request

closes [#2688](https://github.com/exercism/java/issues/2688)

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -217,7 +217,8 @@
         "prerequisites": [
           "classes",
           "strings",
-          "if-else-statements"
+          "if-else-statements",
+          "ternary-operators"
         ],
         "status": "active"
       },
--- a/exercises/concept/wizards-and-warriors/.meta/design.md
+++ b/exercises/concept/wizards-and-warriors/.meta/design.md
@@ -27,6 +27,7 @@ This exercise's prerequisites Concepts are:
 - `classes`
 - `strings`
 - `if-else-statements`
+- `ternary-operators`
 
 ## Representer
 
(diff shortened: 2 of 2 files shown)


config.json
```diff
@@ -217,7 +217,8 @@
         "prerequisites": [
           "classes",
           "strings",
-          "if-else-statements"
+          "if-else-statements",
+          "ternary-operators"
```
sanderploegsma, line 221: I'm not sure we should add `ternary-operators` as prerequisite. Although it's possible to use one, it's not like the exercise can't be solved without using ternary expressions.

In general I think we should only add prerequisites if the concepts are really required in order to solve the exercise.

exercises/concept/wizards-and-warriors/.meta/design.md
```diff
@@ -27,14 +27,22 @@ This exercise's prerequisites Concepts are:
...
 ## Representer
 
 This exercise does not require any specific representation logic to be added to the [representer][representer-java].
 
 ## Analyzer
 
-This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer-java].
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `actionable`: If the solution uses an `if` statement in the `getDamagePoints` method for Wizard and Warrior classes, instruct the student to use a ternary expression instead.
```
sanderploegsma, line 40: Not sure about this one to be honest. It's also pretty off-topic for this concept exercise, so maybe we should leave it out.

exercises/concept/wizards-and-warriors/.meta/design.md
sanderploegsma, line 1: None of these comments are really related to the inheritance concept, so here are some starting points for other comments:

- The student should not make any changes to the `Fighter` class, perhaps we can do something with that. Not sure what exactly, but maybe you have an idea?
- Maybe we can add an `informative` comment about the `@Override` annotation.

sanderploegsma (changes requested):

exercises/concept/wizards-and-warriors/.meta/design.md
manumafe98: The ``@Override`` annotation comment should be ok, I totally forgot that I had that idea as well hahaha, but the ``Fighter`` one for me seems kinda pointless, it's difficult to make comments about the inheritance topic mainly because is already tested and its a requirement to pass the exercise tests.

sanderploegsma: OK yeah I agree

sanderploegsma (approved):

PR #2736, opened 2024-02-21
Adding analyzer feedback for ``secrets`` concept exercise

manumafe98:
# pull request

closes [#2685](https://github.com/exercism/java/issues/2685)

This [solution](https://exercism.org/mentoring/automation/ef0f7c0053ec476c86f7144833e6f049/edit?source%5Bcriteria%5D=secrets&source%5Bpage%5D=1&source%5Btrack_slug%5D=java) could be a great example to see if we want to ask the operator `~` to be used 100% or be flexible and only ask  for the `&` one

Reviewer Resources:

### DIFF
--- a/exercises/concept/secrets/.meta/design.md
+++ b/exercises/concept/secrets/.meta/design.md
@@ -25,7 +25,8 @@ This exercise could benefit from the following rules in the [analyzer]:
 - `essential`: If the student did not use the `>>>` operator in the `shiftBack` method, instruct the student to do so.
 - `essential`: If the student did not use the `|` operator in the `setBits` method, instruct the student to do so.
 - `essential`: If the student did not use the `^` operator in the `flipBits` method, instruct the student to do so.
-- `essential`: If the student did not use the `&` or `~` operator in the `clearBits` method, instruct the student to do so.
+- `essential`: If the student did not use the `&` operator in the `clearBits` method, instruct the student to do so.
+- `informative`: If the solution did not use the `~` operator in the `clearBits` method, inform the student that with it will achieve a more concise solution.
 
 If the solution does not receive any of the above feedback, it must be exemplar.
 Leave a `celebratory` comment to celebrate the success!
--- a/exercises/concept/secrets/.meta/design.md
+++ b/exercises/concept/secrets/.meta/design.md
@@ -17,3 +17,17 @@ The goal of this exercise is to teach the student about bitwise operations in Ja
 This exercise's prerequisites Concepts are:
 
 - 'numbers'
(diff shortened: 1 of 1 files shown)


exercises/concept/secrets/.meta/design.md
```diff
@@ -17,3 +17,17 @@ The goal of this exercise is to teach the student about bitwise operations in Ja
 This exercise's prerequisites Concepts are:
 
 - 'numbers'
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `>>>` operator in the `shiftBack` method, instruct the student to do so.
```
sanderploegsma, line 25: Technically using the `>>` operator is also fine, right? I don't think they are _required_ to use the unsigned shift right operator, a normal shift right will work too.

sanderploegsma (changes requested):

sanderploegsma:
@kahgoh you recently authored this exercise, do you have any suggestions to add or improve?

exercises/concept/secrets/.meta/design.md
kahgoh: Currently, no `>>` wouldn't be fine. One of the tests will fail if they use it and the instructions lead the students to choose `>>>`.

kahgoh:
I think it looks good. I can't think of anything else to add or improve.

manumafe98:
@kahgoh I don't think you could see the content of the link on the description of the PR, just in case I was referring to this

A student solve the exercise like this:
````java
    public static int clearBits(int value, int mask) {
        int and = value & mask;
        return and ^ value;
    }
````

Instead of the most common solution, that would be like this:

````java
    public static int clearBits(int value, int mask) {
        return value & ~mask ;
    }
````

Do you think we should allow it, or remark to use ``~`` in the analyzer?

exercises/concept/secrets/.meta/design.md
sanderploegsma: Ok thanks for clearing that up!

sanderploegsma (approved):

kahgoh:
@manumafe98 Sorry, I did miss the link in the PR's description. I think that is certainly a valid solution as it is still using bit operators, even though I didn't think of it when I created the exercise. From that perspective, I think if they use `value & mask ^ value` it should be allowed, although (if possible) we could suggest to try using `&` and `~` for a shorter solution.

sanderploegsma (approved):

kahgoh (approved):

PR #2744, opened 2024-03-01
Changing String format for String concatenation

manumafe98:
# pull request

As we are generally giving feedback on preferring string concatenation over string.format it makes sense to me that all our reference resolutions (at least for concept exercises) comply with that. Changing the practice ones seemed like a bit overkill, because normally we should give more freedom in those cases to students

Reviewer Resources:


manumafe98:
@sanderploegsma a bit weird, but I had to re run the workflows pipeline because it was stuck for 25 mins in the configlet one

sanderploegsma (approved):

Issue #2713, opened 2024-02-02
#48in24: Add analyzer feedback for Parallel Letter Frequency


sanderploegsma:
The Parallel Letter Frequency exercise on the Java track [is being featured in the #48in24 challenge](https://exercism.org/challenges/48in24/implementation_status), which means that we expect an influx of students attempting to solve it during that week.

It would be nice if the exercise contains some more content by that time.
One example of this is to add [analyzer](https://exercism.org/docs/building/tooling/analyzers) feedback for the exercise, to provide automated feedback on submitted solutions.

Adding analyzer feedback is done in three steps:

1. Create (or update) the `.meta/design.md` file for the exercise explaining what type of feedback the analyzer should provide ([example for the Leap exercise](https://github.com/exercism/java/blob/main/exercises/practice/leap/.meta/design.md)).
2. Add the copy for the analyzer comments to [exercism/website-copy](https://github.com/exercism/website-copy/tree/main/analyzer-comments/java).
3. Implement the analyzer rules in [exercism/java-analyzer](https://github.com/exercism/java-analyzer).

yogan:
Just as a reminder, the exercise text currently states that:

> Single-threaded (non-concurrent) solutions can pass all tests [but the last.](https://www.youtube.com/watch?v=mJZZNHekEQw) Your solution will be tested for concurrency by submitting it as a [Runnable](https://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html) to an [ExecutorService.](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) Your solution must leverage multiple [Threads](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html) to pass the final test.

However, there is not (yet) such a test case.

Maybe for now remove that section and replace it with a note that participants should take care themselves that they are using some kind of parallelism in their solutions.

sanderploegsma:
Thanks for the feedback, good to know! PR is welcome if you're willing to fix it yourself 😁

yogan:
Can you give me a hint where I can the find the text for the exercise? I assumed it would be in this repository, but I have cloned it and searched for the mentioned paragraph without any matches.

sanderploegsma:
Sure! Here it is: https://github.com/exercism/java/blob/main/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md

yogan:
[...]

PR #2752, opened 2024-03-18
Update instructions for parallel-letter-frequency (#2713)

yogan:
See #2713 

The exercise docs suggested that the submitted code is automatically
tested for the usage of parallelism. This is not the case, so the
paragraph rewritten to instead suggest students to check for parallelism
themselves.

To help, a few links to resources about concurrency and parallelism in
Java are added.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/parallel-letter-frequency/.docs/hints.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/hints.md
@@ -1,6 +1,6 @@
 # Hints
 
-The following resources may help you with this:
+The following resources may help you solve this exercise:
 
 - [Concurrency (Oracle Java Tutorials)](https://docs.oracle.com/javase/tutorial/essential/concurrency)
 - [Java Concurrency (Baeldung)](https://www.baeldung.com/java-concurrency)
--- a/exercises/practice/parallel-letter-frequency/.docs/hints.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/hints.md
@@ -1,0 +1,7 @@
+# Hints
+
+The following resources may help you with this:
+
+- [Concurrency (Oracle Java Tutorials)](https://docs.oracle.com/javase/tutorial/essential/concurrency)
+- [Java Concurrency (Baeldung)](https://www.baeldung.com/java-concurrency)
+- [Java 8 Parallel Streams (Baeldung)](https://www.baeldung.com/java-8-streams#parallel-streams)
--- a/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
@@ -1,11 +1,3 @@
 # Instructions append
 
-The automated tests are currently not checking for parallelism. Please ensure yourself that your solution is using some form of parallelism, like threads or parallel streams.
-
-The following resources may help you with this:
-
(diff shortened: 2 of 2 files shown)


exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
```diff
@@ -1,7 +1,11 @@
 # Instructions append
 
-Single-threaded (non-concurrent) solutions can pass all tests [but the last.](https://www.youtube.com/watch?v=mJZZNHekEQw)  Your solution will be tested for concurrency by submitting it as a [Runnable](https://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html) to an [ExecutorService.](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) Your solution must leverage multiple [Threads](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html) to pass the final test.
+The automated tests are currently not checking for parallelism. Please ensure yourself that your solution is using some form of parallelism, like threads or parallel streams.
```
manumafe98, line 3: I'm not sure if adding that the tests do not check for parallelism is a great move, maybe we should remove the 

> Your solution will be tested for concurrency by submitting it as a [Runnable](https://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html) to an [ExecutorService.](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html) Your solution must leverage multiple [Threads](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html) to pass the final test

And then when we create the analyzer for this exercise then try to check that they are using parellelism and if not leave a comment, @sanderploegsma  what are your thoughts on this one?

sanderploegsma: Yes I agree, I don't think the instructions should explicitly mention that parallelism isn't tested. My 2cts:

1. Remove the first paragraph completely.
2. The list of resources look good, but I would move them to a new file `hints.md`, so that students can try to solve the exercise on their own (without spoilers, if you will) if they want to.

yogan: Sounds good, I can take care of that.

Will a new file `hints.md` automatically be picked up and included, or does it have to be configured somewhere?

yogan: I guess `hints.md` will get picked up, others exercises are using it too.

I've updated the PR to remove the paragraph that mentions that parallelism isn't tested and move the links over to `hints.md`.

sanderploegsma: Yup, the `hints.md` file will automatically be picked up by the Exercism tooling, no need to configure it anywhere!

Here's a reference to the relevant documentation in case you're interested: https://exercism.org/docs/building/tracks/practice-exercises#h-documentation-files

exercises/practice/parallel-letter-frequency/.docs/hints.md
```diff
@@ -0,0 +1,7 @@
+# Hints
+
+The following resources may help you with this:
```
sanderploegsma, line 3:
```suggestion
The following resources may help you solve this exercise:
```

sanderploegsma (approved):

PR #2753, opened 2024-03-20
Adding analyzer feedback for  `booking-up-for-beauty` concept exercise

manumafe98:
# pull request

closes [#2672](https://github.com/exercism/java/issues/2672)

The goal is to add the analyzer information for booking-up-for-beauty and update the reference resolution to a more concise solution

Reviewer Resources:

### DIFF
--- a/exercises/concept/booking-up-for-beauty/.meta/design.md
+++ b/exercises/concept/booking-up-for-beauty/.meta/design.md
@@ -22,3 +22,19 @@
 
 - `strings`: know how to use string formatting.
 - `numbers`: know how to apply basic mathematical operators.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use `ofPattern` and `parse` methods in the `schedule` method, instruct the student to do so.
+- `essential`: If the student did not use `isBefore` in the `hasPassed` method, instruct the student to do so.
+- `essential`: If the student did not use `getHour` in the `isAfternoonAppointment` method, instruct the student to do so.
+- `essential`: If the student did not use `ofPattern` and `DateTimeFormatter.format` methods in the `getDescription` method, instruct the student to do so.
+- `essential`: If the student did not use `getYear` in the `getAnniversaryDate` method, instruct the student to do so.
+- `informative`: If the solution uses `String.format` in the `getDescription` method, inform the student that this cause a small performance penalty compared to string concatenation.
+
+If the solution does not receive any of the above feedback, it must be exemplar.
+Leave a `celebratory` comment to celebrate the success!
+
+[analyzer]: https://github.com/exercism/java-analyzer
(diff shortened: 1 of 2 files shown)


exercises/concept/booking-up-for-beauty/.meta/design.md
```diff
@@ -22,3 +22,19 @@
...
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use `ofPattern` and `parse` methods in the `schedule` method, instruct the student to do so.
+- `essential`: If the student did not use `isBefore` in the `hasPassed` method, instruct the student to do so.
+- `essential`: If the student did not use `getHour` in the `isAfternoonAppointment` method, instruct the student to do so.
+- `essential`: If the student did not use `ofPattern` and `DateTimeFormatter.format` methods in the `getDescription` method, instruct the student to do so.
+- `essential`: If the student did not use `getYear` in the `getAnniversaryDate` method, instruct the student to do so.
```
sanderploegsma, line 34: Are we sure that all of these comments are essential? I would argue that a comment should be essential if the student seems to have missed the point of an exercise, for instance by not applying the concept (correctly). However, if the solution is valid and does apply the concept being taught but has room for improvement, I would mark those comments as actionable instead.

For example, if a student implements the `hasPassed` method like this:

```java
public boolean hasPassed(LocalDateTime appointmentDate) {
    return appointmentDate.compareTo(LocalDateTime.now()) < 0;
}
```

I don't think it warrants an `essential` comment. Using `isBefore` could be considered more idiomatic, but the comment should at most be `actionable`.

manumafe98: You're right, at most should be `actionables`, I see this case similar to the `log-levels`, we want to guide the students to the reference resolution or they can solve it in similar ways?

sanderploegsma: The goal is to guide the students to a solution that is somewhat equivalent to the reference solution. A bit of a vague statement, but that is because the exercises are not (and cannot be) set up to exactly lead to the same solution.

From the [docs](https://exercism.org/docs/building/tracks/concept-exercises#h-exercise-files) (emphasis mine):

> Exemplar implementation: provides an **idiomatic implementation** that passes all the tests

As long as someone experienced in Java can look at a solution and think: "I don't think there is much left to improve on", I would argue it's idiomatic and therefore equivalent to the reference solution.

Still vague, I know. I hope you see what I'm getting at though 😉

manumafe98: Okey, but in this kinda cases for example the `substring` one in `log-levels` it can be solved in multiple "correct" ways, this case I would say is the same, so we shouldn't make students implement a certain method, right? maybe we could create like a range of methods that could achieve the same and allow it, like [split,substring] or in this case [compareTo, isBefore]

sanderploegsma: Yeah that sounds more sophisticated indeed!

sanderploegsma (approved):

Issue #2687, opened 2024-01-30
tim-from-marketing: describe Analyzer feedback in `.meta/design.md`


sanderploegsma:
The `.meta/design.md` file of each concept exercise should contain a section describing what feedback the [analyzer](https://github.com/exercism/java-analyzer) should provide and when. This is a prerequisite for implementing the analyzer for each of these exercises.

The section should clearly describe *what* should be checked by the analyzer, what feedback should be given to the student and the *type* of each comment (`essential`, `actionable`, `informative` or `celebratory`, see [the documentation](https://exercism.org/docs/building/tooling/analyzers/interface#h-type-optional)).

Here is a good example of what it might look like: https://github.com/exercism/javascript/blob/main/exercises/concept/mixed-juices/.meta/design.md

When writing the analyzer section for this concept exercise, make sure to keep the following guidelines in mind: https://exercism.org/docs/building/tooling/analyzers/comments

Issue #2555, opened 2023-11-05
Add concept: `optionals`


sanderploegsma:
The goal of this issue is to add a new concept to the Java track that introduces `optionals`. The corresponding concept exercise can probably be based on the "Role Playing Game" exercise from the [Rust track](https://github.com/exercism/rust/tree/main/exercises/concept/role-playing-game) or the [Elm track](https://github.com/exercism/elm/tree/main/exercises/concept/role-playing-game). 

Instructions on how to contribute to the Java track and how to implement concept exercises can be found in the [contributing guide](https://github.com/exercism/java/blob/main/CONTRIBUTING.md).

As always, leave a comment here if you want to work on this exercise, and we'll assign it to you. In case you have any further questions, feel free to ask here.

sanderploegsma:
After adding the concept, it should probably be added as a prerequisite to the following practice exercises:

- Error Handling
- Word Search

alex-md:
I'm interested in addressing this issue. My experience with Java optionals is fairly solid, and I believe I'm up to the task of developing a concept exercise on this topic. I've reviewed the "Role Playing Game" exercise from both the Rust and Elm tracks, and I see potential in adapting it for the Java track. Please consider assigning this issue to me, and I'd welcome any feedback or suggestions you might have. Thanks.

sanderploegsma:
@alex-md that's great! I'll assign the issue to you.

Make sure to check out the contributing guide in this repository if you haven't already, and feel free to comment here in case you have any questions!

manumafe98:
I would like to try on this one!

sanderploegsma:
Great, looking forward to it! I'll assign you.

sanderploegsma:
@manumafe98 just checking; are you still working on this? Or did it perhaps fall off your radar while tackling the other issues? 😉

manumafe98:
@sanderploegsma This is still under my radar! just I was trying to get more knowledge on how to create exercises to make a better job

manumafe98:
@sanderploegsma I was thinking on how to tackle this concept, and checking the rust example that you passed in the description, they introduce in the same exercise the ``nullability`` concept. We already have a our own concept exercise for nullability and my idea was maybe to introduce ``Optionals`` in that same exercise as we have done for example with the exercise that introduces ``List`` and ``Generics``, I would like to know your opinion on this idea.

sanderploegsma:
[...]

PR #2754, opened 2024-03-21
Adding analyzer feedback for `tim-from-marketing` concept exercise

manumafe98:
# pull request

closes [#2687](https://github.com/exercism/java/issues/2687)

This could be an [example](https://exercism.org/mentoring/automation/cd618e6f14334c5ba7f87a6a2f3612c2/edit?source%5Bcriteria%5D=tim&source%5Bpage%5D=2&source%5Btrack_slug%5D=java) that triggers both comments

Reviewer Resources:

### DIFF
--- a/exercises/concept/tim-from-marketing/.meta/design.md
+++ b/exercises/concept/tim-from-marketing/.meta/design.md
@@ -24,6 +24,7 @@
 This exercise could benefit from the following rules in the [analyzer]:
 
 - `essential`: If the student did not use the `null` expression in the `print` method, instruct the student to do so.
+- `actionable`: If the solution uses `Optionals` to solve the exercise, encourage the student to try solving it differently.
 - `informative`: If the solution uses `String.format`, instruct the student to use simple string concatenation instead.
   Explain that `String.format` is significantly slower than concatenating strings and should be used in more complex scenarios.
 
--- a/exercises/concept/tim-from-marketing/.meta/design.md
+++ b/exercises/concept/tim-from-marketing/.meta/design.md
@@ -18,3 +18,16 @@
 
 - `strings`: strings will be compared to null and basic methods from strings will be called.
 - `if-else-statements`: using a conditional statement.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `null` expression in the `print` method, instruct the student to do so.
+- `informative`: If the solution uses `String.format`, instruct the student to use simple string concatenation instead.
+  Explain that `String.format` is significantly slower than concatenating strings and should be used in more complex scenarios.
+
+If the solution does not receive any of the above feedback, it must be exemplar.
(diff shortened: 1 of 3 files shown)


sanderploegsma:
The example you linked is an interesting case: clearly the author has enough Java knowledge to understand the Nullable concept, or they wouldn't have chosen to use `java.util.Optional`. In these cases I'm not sure they will appreciate if the analyzer is being a wise-ass by pointing out that they should use `null` instead. On the other hand, the goal of this exercise is to practice using `null`. 

I think I would prefer it if we leave an `actionable` comment that acknowledges that using `java.util.Optional` is a great way to deal with nullable arguments, but nudges the author to also try to solve the exercise without it. Once #2555 is completed, we could even point out that optionals are covered in a later concept.

What do you think?

manumafe98:
> I think I would prefer it if we leave an `actionable` comment that acknowledges that using `java.util.Optional` is a great way to deal with nullable arguments, but nudges the author to also try to solve the exercise without it. Once #2555 is completed, we could even point out that optionals are covered in a later concept.
> 
> What do you think?

Sure, I like that!

exercises/concept/tim-from-marketing/.meta/design.md
```diff
@@ -18,3 +18,17 @@
 
 - `strings`: strings will be compared to null and basic methods from strings will be called.
 - `if-else-statements`: using a conditional statement.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `null` expression in the `print` method, instruct the student to do so.
+- `actionable`: If the solution uses `Optionals` to solve the exercise, encourage the student to try solving it differently.
```
sanderploegsma, line 27:
```suggestion
- `actionable`: If the solution uses `Optionals` to solve the exercise, encourage the student to try solving it using `null` instead.
```

exercises/concept/tim-from-marketing/.meta/design.md
```diff
@@ -18,3 +18,17 @@
 
 - `strings`: strings will be compared to null and basic methods from strings will be called.
 - `if-else-statements`: using a conditional statement.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `null` expression in the `print` method, instruct the student to do so.
```
sanderploegsma, line 26: Note that this comment currently would still apply to the example solution you linked to in the PR. 

To be honest I think we can remove this comment completely now that we have the one covering optionals: I don't think there is a way to pass the tests without using either `null` or `java.util.Optional`.

sanderploegsma (changes requested):

exercises/concept/tim-from-marketing/.meta/design.md
manumafe98: 👍

sanderploegsma (approved):

PR #2755, opened 2024-03-22
Adding analyzer feedback for `captains-log` concept exercise

manumafe98:
# pull request

closes [#2674](https://github.com/exercism/java/issues/2674)

Reviewer Resources:

### DIFF
--- a/exercises/concept/captains-log/.meta/design.md
+++ b/exercises/concept/captains-log/.meta/design.md
@@ -19,3 +19,16 @@
 - `strings`: know how to use string formatting.
 - `numbers`: know how to apply basic mathematical operators.
 - `arrays`: know how to retrieve array elements by their index.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `nextInt` method in the `randomShipRegistryNumber` method, instruct the student to do so.
+- `essential`: If the student did not use the `nextDouble` method in the `randomStardate` method, instruct the student to do so.
+- `informative`: If the solution uses `String.format` in the `randomShipRegistryNumber` method, inform the student that this cause a small performance penalty compared to string concatenation.
+
+If the solution does not receive any of the above feedback, it must be exemplar.
+Leave a `celebratory` comment to celebrate the success!
+
+[analyzer]: https://github.com/exercism/java-analyzer
--- a/exercises/concept/captains-log/.meta/design.md
+++ b/exercises/concept/captains-log/.meta/design.md
@@ -24,11 +24,16 @@
 
 This exercise could benefit from the following rules in the [analyzer]:
 
-- `essential`: If the student did not use the `nextInt` method in the `randomShipRegistryNumber` or `randomPlanetClass` methods, instruct the student to do so.
-- `essential`: If the student did not use the `nextDouble` method in the `randomStardate` method, instruct the student to do so.
+- `actionable`: If the solution hardcodes the number of planet classes in `randomPlanetClass`, instruct the student to use `length` method instead.
+- `informative`: If the solution uses [`RandomGenerator.nextDouble(double)`][nextDouble(double)], inform the student that they could use [`RandomGenerator.nextDouble(double, double)`][nextDouble(double, double)] to achieve a clearer solution.
+- `informative`: If the solution uses [`RandomGenerator.nextInt(int)`][nextInt(int)], inform the student that they could use [`RandomGenerator.nextInt(int, int)`][nextInt(int, int)] to achieve a clearer solution.
 - `informative`: If the solution uses `String.format` in the `randomShipRegistryNumber` method, inform the student that this cause a small performance penalty compared to string concatenation.
 
 If the solution does not receive any of the above feedback, it must be exemplar.
 Leave a `celebratory` comment to celebrate the success!
 
 [analyzer]: https://github.com/exercism/java-analyzer
+[nextDouble(double)]: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/RandomGenerator.html#nextDouble(double)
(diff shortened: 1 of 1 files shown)


exercises/concept/captains-log/.meta/design.md
```diff
@@ -19,3 +19,16 @@
 - `strings`: know how to use string formatting.
 - `numbers`: know how to apply basic mathematical operators.
 - `arrays`: know how to retrieve array elements by their index.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: If the student did not use the `nextInt` method in the `randomShipRegistryNumber` or `randomPlanetClass` methods, instruct the student to do so.
+- `essential`: If the student did not use the `nextDouble` method in the `randomStardate` method, instruct the student to do so.
```
sanderploegsma, line 28: Given that a solution needs to pass all tests in order to receive feedback from the analyzer; do you have an example of a solution that does _not_ use the `nextInt` or `nextDouble` methods?

manumafe98: My train of thought was similar as the secrets concept exercise, that its going to be pretty rare to see someone solving it with another methods, but if it happens the analyzer comment triggers

sanderploegsma: Well my reason for asking was mainly because I want to avoid the same situation as in Log Levels, where feedback is given even though the solution needs no improvement.

I think the best way to avoid this is to narrow down the cases in which certain comments are triggered. 

So in this case we can choose to only leave the `Random.nextInt` comment if a solution uses `Random.next` for example. Which I doubt will happen often, and if it does it's probably more of an actionable comment instead of an essential one.

manumafe98: Do you see someone using `Random.next` to solve the ones with `nextInt`? hmm maybe remove those comments and only left the `String.format` one

sanderploegsma:
I've been going through the community solutions and these solutions stood out to me:

| Solution | Observations |
| ------ | ----- |
| [Link](https://exercism.org/mentoring/automation/7b1204d25e8848a18edcdf7b810adc12/edit?source%5Bcriteria%5D=captain&source%5Bpage%5D=1&source%5Btrack_slug%5D=java) | Hard-coding the number of planet classes (`10`) instead of using the length of the array, could be an `actionable` comment. |
| [Link](https://exercism.org/mentoring/automation/edafb85b8a6c438d9624cfc3657e4e01/edit?source%5Bcriteria%5D=captain&source%5Bpage%5D=1&source%5Btrack_slug%5D=java) | This solution uses a Java version >= 17, judging by the usage of the [`RandomGenerator.nextDouble(double)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/RandomGenerator.html#nextDouble(double)) method added in Java 17. However, those Java versions also contain [`RandomGenerator.nextDouble(double,double)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/RandomGenerator.html#nextDouble(double,double)) and [`RandomGenerator.nextInt(int,int)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/RandomGenerator.html#nextInt(int,int)) which support passing both a lower and upper bound, which makes the solution a bit easier to read. Could be an `informative` comment. |

exercises/concept/captains-log/.meta/design.md
sanderploegsma: So far no, I have not seen any. I posted other findings below.

manumafe98: I removed those comments finally

exercises/concept/captains-log/.meta/design.md
```diff
@@ -19,3 +19,21 @@
...
 - `numbers`: know how to apply basic mathematical operators.
 - `arrays`: know how to retrieve array elements by their index.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `actionable`: If the solution hardcodes the number of planet classes in `randomPlanetClass`, instruct the student to use `length` method instead.
+- `informative`: If the solution uses [`RandomGenerator.nextDouble(double)`][nextDouble(double)], inform the student that they could use [`RandomGenerator.nextDouble(double, double)`][nextDouble(double, double)] to achieve a clearer solution.
+- `informative`: If the solution uses [`RandomGenerator.nextInt(int)`][nextInt(int)], inform the student that they could use [`RandomGenerator.nextInt(int, int)`][nextInt(int, int)] to achieve a clearer solution.
```
sanderploegsma, line 29: Note that this depends on the Java version used by the student. `nextInt(int)` is available in Java 8 and above, but `nextInt(int,int)` is not available until Java 17.

So this feedback may confuse students who are using older Java versions because their code will not compile after applying the suggestion.

sanderploegsma (changes requested):

exercises/concept/captains-log/.meta/design.md
manumafe98: This is because of the exercism cli, right? because on the ui you cannot choose the java version, I never used it before, and I guess we cannot check for the java version priori to leave the comment, so well maybe leave only the actionable for the length and the string,format ones

sanderploegsma: Yeah I don't have a clear-cut answer for this either. I suppose it's safe to assume that when a solution uses `RandomGenerator.nextDouble(double, double)` they must be using Java 17 or higher, but other than that I don't think we can manage it right now.

Keep in mind though that the design document contains the _intentions_ behind the analyzer, that doesn't necessarily mean that everything in it should be (or will be) implemented. IMO it's fine if this `design.md` describes that the analyzer could point solutions to the `RandomGenerator` methods, but only when using Java 17 and above. If we ever come to a point where we can determine this in the analyzer we have a record here reminding us to add those rules.

manumafe98: Okey so I think with this clarification that I added to those comments we are great then.

sanderploegsma (approved):

Issue #2625, opened 2024-01-03
Reformat concept ``Loops``


manumafe98:
I was checking the ``Python`` exercism track, and I saw that it covers all the loops in a single concept called ``Loops`` and as the Java track does not cover ``While`` loops and has ``For-each`` and ``For`` loops separated on why do not take the same approach as in the Python track. 

What do you guys think?

sanderploegsma:
Another good observation, thanks! In fact, I checked my notes and it seems like I agree with you:

> Concepts to add:
> - ...
> - While loops. Actually, shouldn't we just combine for/foreach/while/do-while into Loops or Looping?

One thing to keep in mind is that the concept should be taught through an exercise. Merging the existing concepts into one big one that covers For, For-Each and While means that we would have to update the existing exercise or come up with a new one that covers all of these constructs.

Feel free to comment any ideas on how to do this, as I'm inclined to agree with you on having a single concept on looping constructs.

manumafe98:
Well maybe we could re format the exercise used in the python track [making-the-grade](https://github.com/exercism/python/tree/main/exercises/concept/making-the-grade) to put all in one, what do you think?

manumafe98:
I was thinking on adding this to the forum as I did with the ``Functional Programming`` topic as well

josealonso:
@manumafe98, glad to help you wit this task.

PR #2773, opened 2024-04-04
Starting to reformat loops into a unified concept

manumafe98:
# pull request

closes [#2625](https://github.com/exercism/java/issues/2625)

This is a work in progress

Reviewer Resources:


concepts/loops/about.md
```diff
@@ -0,0 +1,178 @@
+# About
+
+In Java there are four looping constructs, two that are condition centric: `while` and `do-while` and the other two are iteration centric: `for` and `for-each`.
```
sanderploegsma, line 3: Maybe it makes sense to rewrite this sentence to list the constructs in the order in which they are described below, so `for`, `for-each`, `while`, `do-while`.

concepts/loops/about.md
```diff
@@ -0,0 +1,178 @@
...
+
+## For
+
+A for loop provides a mechanism to execute a group of statements repeatedly until some condition is met.
```
sanderploegsma, line 7: Ironically this sentence talks about `for` using a condition, where the first paragraph lists `while` and `do-while` as being condition-centric and `for`/`for-each` being something else 😜

concepts/loops/about.md
```diff
@@ -0,0 +1,178 @@
...
+The output would be:
+
+```text
+square of 1 is 1
+square of 2 is 4
+square of 3 is 9
+square of 4 is 16
+```
+
+If iterating through every element in a collection, a `for-each` loop is preferred, but it can be done with a `for` loop like this:
```
sanderploegsma, line 58:
```suggestion
Iterating through every element in a collection is usually performed using a `for-each`, but it can be done with a `for` loop like this:
```

concepts/loops/introduction.md
sanderploegsma, line 1: I know a lot of other concepts just reuse the contents of `about.md` in their respective `introduction.md`, but I think in this case the concept introduction is pretty verbose. I'm wondering whether we can make the introduction a bit shorter so that they actually give a short introduction to the concept and nothing more.

config.json
```diff
@@ -61,9 +61,7 @@
         "name": "Bird Watcher",
         "uuid": "60600885-8684-45c3-b8a9-a2f9e3ce2d97",
         "concepts": [
-          "arrays",
-          "for-loops",
-          "foreach-loops"
+          "arrays"
```
sanderploegsma, line 64: One thing that worries me here is that the exemplar implementation of `bird-watcher` actually uses loops. And we can't make that a prerequisite here because that would lead to a dependency cycle, because the new exercise depends on `lists` which in turn depends on `arrays`

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    static int getAverageScoreOfStudents(List<Integer> studentScores, int numberOfStudents) {
+        int count = 0;
+        int sum = 0;
+
+        while (count < numberOfStudents) {
+            sum += studentScores.get(count);
+            count++;
+        }
```
sanderploegsma, line 25: What makes this method a good candidate for a `while` loop? I could just as easily (and probably would) implement this using a `for` loop:

```java
int sum = 0;

for (int i = 0; i < numberOfStudents; i++) {
    sum += studentScores.get(i);
}

return sum / numberOfStudents;
```

exercises/concept/making-the-grade/.meta/config.json
```diff
@@ -0,0 +1,23 @@
...
+      ".meta/src/reference/java/MakingTheGrade.java"
+    ],
+    "invalidator": [
+      "build.gradle"
+    ]
+  },
+  "forked_from": [
+    "python/making-the-grade"
+  ],
+  "blurb": "Learn to use different kinds of loops by automating the tasks of an assitant teacher."
```
sanderploegsma, line 22:
```suggestion
  "blurb": "Learn to use different kinds of loops by automating the tasks of an assistant teacher."
```

exercises/concept/making-the-grade/.docs/hints.md
```diff
@@ -0,0 +1,17 @@
+# Hints
+
+## 1. Non-Passing Students
+
+- Try using one of the loops that is iteration centric.
+
+## 2. Get the average grade of students
+
+- Try using one of the loops that is condition centric using numberOfStudents as one of the core conditions.
```
sanderploegsma, line 9:
```suggestion
- Try using one of the loops that is condition centric using `numberOfStudents` as one of the core conditions.
```

exercises/concept/making-the-grade/.docs/instructions.md
```diff
@@ -0,0 +1,60 @@
...
+```java
+MakingTheGrade.countFailedStudents(List.of(40, 70, 80, 20, 39));
+// => 3
+```
+
+## 2. Get the average grade of students
+
+As your second task of the day the professor asks you to get the average score of a certain number of students.
+
+Create the method `getAverageScoreOfStudents(List<Integer> studentScores, int numberOfStudents)` that takes a `List<Integer>` of studentScores and a `int` of numberOfStudents. This method should get the average score of the number of students passed. numberOfStudents it's guaranteed to be lower than the size of the list of scores.
```
sanderploegsma, line 25:
```suggestion
Create the method `getAverageScoreOfStudents(List<Integer> studentScores, int numberOfStudents)` that takes a `List<Integer>` of studentScores and a `int` of numberOfStudents.
This method should get the average score of the number of students passed.
The number of students is guaranteed to be lower than the size of the list of scores.
```

exercises/concept/making-the-grade/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+## Learning objectives
+
+- Know how to repeteadly execute code using a `for` loop.´
+- Know how to repeteadly execute code using a `for-each` loop.
+- Know how to repeteadly execute code using a `while` loop.
+- Know how to repeteadly execute code using a `do-while` loop.
+
+## Out of scope
+
+- Specific iteration over a `Map`
```
sanderploegsma, line 12: Why is this listed here? When I think about loops I don't necessarily think of maps.

Things that are related and seem to be out of scope of this exercise:

- Breaking out of loops using the `break` keyword
- Skipping iterations using the `continue` keyword

concepts/loops/about.md
sanderploegsma, line 1: I think this content should also talk about the `break` and `continue` keywords, even if it's not covered by the exercise. They are a vital part of looping constructs.

concepts/loops/about.md
manumafe98: Done!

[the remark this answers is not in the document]
manumafe98: Done!

config.json
manumafe98: I guess probably the solution would be to reformat `bird-watcher` as well to avoid needing loops? what do you suggest?

concepts/loops/introduction.md
manumafe98: Yeah! why not, let me think on what can I add to the introduction!

exercises/concept/making-the-grade/.meta/design.md
manumafe98: Instead of adding those to the out of scope I prefer mention them and maybe try to add exercise to cover those topics now that you say it, do you have in mind some other topics that could be out of scope? or I should remove it?

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
manumafe98: So my train of thought was that as I'm giving `numberOfStudents` as a parameter for the method and `while` is condition centric loop that would guide students to use `while` instead of `for` but yeah it can be solved the way you said as well. Do you have another ideas?

concepts/loops/about.md
manumafe98: Done!

concepts/loops/introduction.md
manumafe98: Done!

concepts/loops/about.md
manumafe98: Done!

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
```diff
@@ -53,4 +53,32 @@ static List<String> studentRanking(List<Integer> studentScores, List<String> stu
...
+
+        for (int score : studentScores) {
+            if (score <= 40) {
+                break;
+            }
+
+            count++;
+        }
+
+        return count;
```
sanderploegsma, line 82: This actually makes for a nice `while` task:

```java
class MakingTheGrade {

    static int evaluateChallengingExam(List<Integer> studentScores) {
        int index = 0;

        while (studentScores.get(index) > 40) {
            index++;
        }

        return index + 1;
    }
}
```

😆

exercises/concept/making-the-grade/src/test/java/MakingTheGradeTest.java
```diff
@@ -127,4 +127,22 @@ void testEmptyLists() {
...
+    void testThreeOddScores() {
+        assertThat(
+            MakingTheGrade.countOddScores(List.of(20, 35, 40, 10, 39, 77))
+            ).isEqualTo(3);
+    }
+
+    @Test
+    @Tag("task:6")
+    @DisplayName("Two students passed the exam before the first non-passing score")
+    void testTwoStudentsPassedTheExamBeforeTheNonPassingScore() {
```
sanderploegsma, line 143: Maybe also add tests where:

- No students passed
- All students passed

exercises/concept/making-the-grade/src/test/java/MakingTheGradeTest.java
```diff
@@ -127,4 +127,22 @@ void testEmptyLists() {
             MakingTheGrade.studentRanking(List.of(), List.of())
             ).isEqualTo(List.of());
     }
+
+    @Test
+    @Tag("task:5")
+    @DisplayName("Three student scores were odd after the correction of the math tests")
+    void testThreeOddScores() {
```
sanderploegsma, line 134: Maybe also add tests where:

- No scores are odd
- All scores are odd

exercises/concept/making-the-grade/.meta/design.md
manumafe98: 6 tasks in a not very complex concept for me seems acceptable

config.json
manumafe98: > Maybe, I don't know to be honest. That would mean that the exercise is limited to just array creation and array indexing?

Yeah, could be something like this:
- How to create them
- How to access the elements in them
- How to access the length of the array
- How to create multi dimensional arrays
- Maybe how to make a copy of an array

concepts/loops/about.md
```diff
@@ -0,0 +1,236 @@
...
+} while (counter < 2);
+```
+
+which outputs:
+
+```text
+1
+```
+
+## Break
```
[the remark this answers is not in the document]
manumafe98: Done!

exercises/concept/making-the-grade/src/test/java/MakingTheGradeTest.java
manumafe98: Done!

exercises/concept/making-the-grade/src/test/java/MakingTheGradeTest.java
manumafe98: Done!

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
manumafe98: Do you suggest to change this one for the one that is actually the while task? and make a new one for break ?

manumafe98:
Hey @kahgoh Do you mind helping me review this PR? Sander is taking a break from exercism so I will like your opinions on this topics

kahgoh:
Hey @manumafe98, I'd be happy to help :) I'll hope to have a look at it within a day or two.

manumafe98:
> Hey @manumafe98, I'd be happy to help :) I hope to have a look at it within a day or two.

Thanks! no rush

exercises/concept/making-the-grade/.docs/instructions.md
```diff
@@ -0,0 +1,87 @@
...
+This method should count up the number of odd students scores and return that count as an integer.
+
+```java
+MakingTheGrade.countOddScores(List.of(20, 35, 40, 10, 39, 77));
+// => 3
+```
+
+## 6. Challenging Exam
+
+The teacher has prepared a challenging exam and wants to analyze the results. As his assistant you'll evaluate the exams in the same order students handed them in. The teacher is particularly interested in knowing how many students passed the exam before the first non-passing student.
```
kahgoh, line 78: I notice there are some places where the paragraph is written on one continuous line, but it is inconsistent with most of the document is written with one sentence per line. Did you mean to put it a sentence per line?

```suggestion
The teacher has prepared a challenging exam and wants to analyze the results. 
As his assistant you'll evaluate the exams in the same order students handed them in. 
The teacher is particularly interested in knowing how many students passed the exam before the first non-passing student.
```

config.json
manumafe98: @kahgoh I would like to discuss this topic as well, before continuing with the changes to the loop concept, because we will probably have to reformat how the array concept is done as well. 

I think what I proposed seems okey, what do you think?

kahgoh: Yeah, I think it be fine if we simplify the bird watcher exercise as you've suggested. If we want students the chance to  loop an over arrays, I think we could do it in Making the Grade if we wanted to.

manumafe98: Okey I will prioritize that simplification then, and then we continue discussing the loop concept exercises, thanks for your opinions!

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
josealonso: @manumafe98, @kahgoh, happy to work with you on this one, guys.

josealonso:
@manumafe98, @kahgoh, happy to work with you on this one, guys.

manumafe98:
> @manumafe98, @kahgoh, happy to work with you on this one, guys.

Sure @josealonso , if you want to continue with this one go ahead, what remains to be done is mainly the rework of the arrays concept exercise

exercises/concept/making-the-grade/.meta/src/reference/java/MakingTheGrade.java
josealonso: I would use the for-each loop in both methods. I think the *traditional* for loop should only be used when you need to control the index value, like in C. That's what I said to some of my mentees.
On the other hand, creating an exercise where the exemplar solution uses the Stream API is crucial, in my opinion. In almost any interview for a Java junior role you are going to be asked about **Streams**. 
I know it because I have been a Java programmer for the last five years.  

@manumafe98 @kahgoh @sanderploegsma

josealonso:
> > @manumafe98, @kahgoh, happy to work with you on this one, guys.
> 
> Sure @josealonso , if you want to continue with this one go ahead, what remains to be done is mainly the rework of the arrays concept exercise

Thanks for your reply @manumafe98. I just went through all the files you created and you have done a lot of work already!!
I'll try to get familiar with all the files needed to create an exercise.

concepts/loops/about.md
```diff
@@ -0,0 +1,240 @@
...
+
+which outputs:
+
+```text
+1
+2
+3
+4
+5
+```
```
josealonso, line 151: I would add a small explanation on the similarity of both loops --->
--------

Notice that while loop is equivalent to the following:

```java
for (int counter = 1; counter <= 5; counter++) {
    System.out.println(counter);
}
```

concepts/loops/about.md
```diff
@@ -0,0 +1,240 @@
...
+    if (i < 3) {
+        continue;
+    }
+
+    System.out.println(i);
+}
+```
+
+which outputs:
+
```
josealonso, line 231: ## Using `break` and `continue` is discouraged

However, using a `break` inside a loop is not recommended, as it might make the code difficult to read. A boolean variable, usually called a `flag`, can be used instead:

```java
boolean showNumber = true;
for (int i = 0; i < 10; i++) {
    if (i == 5) {
        showNumber = false;
    }

    if (showNumber) {
        System.out.println(i);
    }
}
```

or using a while loop:

```java
int i = 0;
boolean showNumber = true;
while (i < 10) {
    if (i == 5) {
        showNumber = false;
    }

    if (showNumber) {
        System.out.println(i);
    }
    i++;
}
```

concepts/loops/about.md
josealonso: In general, the same control flow can be expressed using either a `while` or a `for` loop.

concepts/loops/about.md
kahgoh: I think it is fine the way it is now. I have definitely heard we shouldn't use `break` and `continue` before, but I don't think this is universal and they are used in projects. Examples can be seen in some well known Java open source projects, like [Spring](https://github.com/spring-projects/spring-data-jpa/blob/0187162bfaae9eb3033093e9041b989312f4d036/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/QueryTransformers.java#L52), [Maven](https://github.com/apache/maven/blob/69980eeafcb1c0371e48df89953390b6d939d61a/impl/maven-impl/src/main/java/org/apache/maven/impl/AbstractNode.java#L41) and even [Eclipse](https://github.com/eclipse-platform/eclipse.platform/blob/083f7afd5a2665b4418ae74503823f44093b40fc/ua/org.eclipse.help.webapp/src/org/eclipse/help/internal/webapp/servlet/DynamicXHTMLFilter.java#L120). Even when I try Google on this subject I get answers for both sides, like in the discussion on [this](https://softwareengineering.stackexchange.com/questions/58237/are-break-and-continue-bad-programming-practices) one or [this](https://stackoverflow.com/questions/18188123/is-it-bad-practice-to-use-break-to-exit-a-loop-in-java) one. Given the differing opinions and the usages out there, I'm wary of saying don't use `breaks` and `continue`.

manumafe98:
@josealonso What was left to do on this PR was mainly related to the array concept, to reformat it as well, as those exercise are heavy related on loops and it kinda does not make sense to be like that, but we need ideas because if we remove the exercises that contain loop the array concept will be only about indexing and creation

concepts/loops/about.md
manumafe98: I agree with @kahgoh I think the addition of `break` and `continue` are valuable for the exercise and the topic

josealonso:
> @josealonso What was left to do on this PR was mainly related to the array concept, to reformat it as well, as those exercise are heavy related on loops and it kinda does not make sense to be like that, but we need ideas because if we remove the exercises that contain loop the array concept will be only about indexing and creation

But you're using a List already. And the concept introduced in this exercise is `loops`. I may be missing something. :thinking:

kahgoh:
> but we need ideas because if we remove the exercises that contain loop the array concept will be only about indexing and creation

Perhaps a few things we could get the students to try in the array exercise:
- getting the length of the array (perhaps  checking to avoid `IndexOutOfBounds`)
- updating elements at specific indexes
- using of [java.util.Arrays](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html) to sort and binary search or to make a [String](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#toString-java.lang.Object:A-) representation

kahgoh:
> But you're using a List already. And the concept introduced in this exercise is loops. I might be missing something. 🤔

@josealonso, the problem is that the loops concept would now depend on the lists concept, but the lists concept currently depend on the arrays concept, which uses currently loops (circular dependency). We'd need to also update the arrays concept exercise to not use loops.

josealonso:
> > But you're using a List already. And the concept introduced in this exercise is loops. I might be missing something. 🤔
> 
> @josealonso, the problem is that the loops concept would now depend on the lists concept, but the lists concept currently depend on the arrays concept, which uses currently loops (circular dependency). We'd need to also update the arrays concept exercise to not use loops.

Thanks for your answer @kahgoh! Presenting arrays and loops at the same time might not be the ideal way to do it, but I think the priority should be the functional features introduced in Java 8. A number of approaches already use Stream methods, so it shouldn't be difficult. The problem is that many students don't even go to the *dig-deeper* sections.

PR #2791, opened 2024-05-01
Add approaches for Knapsack

### LINKED ISSUE #2712: #48in24: Add approaches to Knapsack
The Knapsack exercise on the Java track [is being featured in the #48in24 challenge](https://exercism.org/challenges/48in24/implementation_status), which means that we expect an influx of students attempting to solve it during that week.

It would be nice if the exercise contains some more content by that time.
One example of this is to add one or more [approaches](https://exercism.org/docs/building/tracks/approaches) to the exercise, so that students can dive in deeper after they have solved it themselves.

For an example of how approaches have been added to other practice exercises, take a look at [the approaches for the Leap exercise](https://github.com/exercism/java/tree/main/exercises/practice/leap/.approaches).

kahgoh:
# pull request

Approaches for knapsack, ready for review!

Resolves #2712 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/knapsack/.approaches/config.json
+++ b/exercises/practice/knapsack/.approaches/config.json
@@ -1,0 +1,27 @@
+{
+  "introduction": {
+    "authors": [
+      "kahgoh"
+    ]
+  },
+  "approaches": [
+    {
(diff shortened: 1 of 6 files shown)


sanderploegsma (approved):
What a great write-up, thanks!

PR #2792, opened 2024-05-05
dnd-character: Add test case for ability with lowest number occurring twice

kahgoh:
# pull request

During a mentoring session, I noticed a solution that passed all the existing tests, but would have failed to meet the following example from the instructions. I wonder if it be worth adding a test for it?

> 3, 5, 3, 4: You discard the 3 and sum 5 + 3 + 4 = 12, which you assign to wisdom.

Note, I have put the PR here instead of the [problem specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/dnd-character/canonical-data.json) because I noticed the Java track already has some tests around the `ability` method and the only `ability` test in the spec seems more vague. I think the `ability` tests we have might be Java specific? 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/dnd-character/src/test/java/DnDCharacterTest.java
+++ b/exercises/practice/dnd-character/src/test/java/DnDCharacterTest.java
@@ -148,6 +148,12 @@ public class DnDCharacterTest {
 
     @Disabled("Remove to run test")
     @Test
+    public void testAbilityCalculationsWithTwoLowestNumbers() {
+        assertThat(dndCharacter.ability(List.of(3, 5, 3, 4))).isEqualTo(12);
+    }
+
+    @Disabled("Remove to run test")
+    @Test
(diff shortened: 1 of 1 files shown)


manumafe98:
Maybe it makes more sense to discuss this directly into an issue? how was the student solving the exercise to not meet with that particular test case?

kahgoh:
@manumafe98 I've opened up a post in the forums to discuss: https://forum.exercism.org/t/missing-test-case-for-ability-with-lowest-number-occurring-twice-in-dnd-character/11097

manumafe98:
> @manumafe98 I've opened up a post in the forums to discuss: https://forum.exercism.org/t/missing-test-case-for-ability-with-lowest-number-occurring-twice-in-dnd-character/11097

Great! well we will see when it gets some answers!

ErikSchierboom (approved):

PR #2814, opened 2024-06-22
Fix the definition of unchecked exception in the error-handling exercise instructions

rkabhishek:
# pull request

This PR clarifies the definition of unchecked exception and fixes minor grammar errors in the error-handling exercise instructions.
---

Reviewer Resources:

### DIFF
--- a/exercises/practice/error-handling/.docs/instructions.append.md
+++ b/exercises/practice/error-handling/.docs/instructions.append.md
@@ -22,7 +22,7 @@ This is because checked exceptions are meant to be handled at runtime, i.e. they
 
 They're often used when a method can't return any valid result, for example a search method which hasn't found the item it was searching for.
 
-It's an alternative to returning null or a error code. A checked exception is better than those alternatives because it forces the user of the method to consider the error case.
+It's an alternative to returning null or an error code. A checked exception is better than those alternatives because it forces the user of the method to consider the error case.
 
 ## Unchecked exceptions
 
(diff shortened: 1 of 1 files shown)


rkabhishek:
Hi, I have 2 questions regarding the documentation for this exercise:
1. I think that the line I have edited is not true, right? Out of memory comes under the category of Error which denotes a serious problem whereas unchecked exceptions are the ones which usually are programming bugs that can be fixed and are subclass of RuntimeException. Correct, me if I am wrong. If yes, I believe the docs can be clarified. 

3. The file I edited is called _instructions.append.md_ I also see an _instructions.md_ file. Can we keep the content in one file only?

manumafe98:
Hey @rkabhishek 

> 1. I think that the line I have edited is not true, right? Out of memory comes under the category of Error which denotes a serious problem. In contrast, unchecked exceptions are the ones that usually are programming bugs that can be fixed and are a subclass of RuntimeException. Please let me know if I am wrong. If not, the docs can be clarified.

Yes, you're right that part of the doc is not clear in its current state, if you have an idea on how to make it clearer go ahead!

> 2. The file I edited is called _instructions.append.md_ . I also see an _instructions.md_ file. Can we keep the content in one file only?

Normally the append file is for extra notes and we do not mind to have 2 files, I can see that in this one the majority of the instructions are on the append that is weird.

kahgoh (approved):

kahgoh:
Changes looks good to me. Thanks!

PR #2827, opened 2024-07-26
Fix links for Scrabble Score dig deeper page

annewu109:
# pull request

First pull request on this repository, so forgive me if I did anything wrong.

While I was looking at the Dig Deeper section of Scrabble Score, I saw that the links to each approach were broken because "scrabble-score" was instead written as "scrabble=score". They should now link to the correct pages.

---

Reviewer Resources:


kotp (approved):
Approving to bring attention, not sure if I have more oomph than that to bring it in though.  But it does indeed fix the links.

manumafe98:
Thanks for the PR!

manumafe98 (approved):

PR #2834, opened 2024-08-31
Add dot-dsl exercise

### LINKED ISSUE #1809: dot-dsl: implement exercise
The exercise **dot-dsl** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/main/exercises/dot-dsl). 

How to implement a new exercise for the Java track is described in detail in [CONTRIBUTING.md](https://github.com/exercism/java/blob/main/CONTRIBUTING.md). Please have a look there first before starting working on the exercise. 

kahgoh:
# pull request

Closes #1809

---

Reviewer Resources:


exercises/practice/dot-dsl/src/main/java/DotDsl.java
```diff
@@ -0,0 +1,65 @@
...
+        public Graph edge(String start, String end, Map<String, String> attributes) {
+            throw new UnsupportedOperationException();
+        }
+
+        public Map<String, String> getAttributes() {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    static class Node {
```
manumafe98, line 42: I was thinking as this practice exercise is already one of the hard ones, maybe we can simplify the task of the user only having them to implement `Graph` and `Element` and we give the `Edge` and `Node` already implemented.

We can use the [tree-building](https://github.com/exercism/java/tree/85e331f85bbace9b4227bf7f6e33f5ee12b5fd1b/exercises/practice/tree-building) practice exercise as an example that gives the `Record` and `TreeNode` implementations.

exercises/practice/dot-dsl/src/main/java/DotDsl.java
```diff
@@ -0,0 +1,65 @@
...
+
+    static class Graph {
+        public Graph() {
+        }
+
+        public Graph(Map<String, String> attributes) {
+        }
+
+        public Collection<Node> getNodes() {
+            throw new UnsupportedOperationException();
```
manumafe98, line 14: It will be great to add the following message `Delete this statement and write your own implementation.` to all the `UnsupportedOperationException`
```suggestion
            throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
```

exercises/practice/dot-dsl/src/main/java/DotDsl.java
```diff
@@ -0,0 +1,65 @@
+import java.util.Collection;
+import java.util.Map;
+
+class DotDsl {
+
```
manumafe98, line 5: I will probably also tend to give the `Element` class structure to the user instead of them having it to implement it from scratch

manumafe98 (changes requested):

manumafe98:
Really good job on implementing this one @kahgoh I would say it's one of the tricky ones!

exercises/practice/dot-dsl/src/main/java/DotDsl.java
kahgoh: I thought about that while implementing the exercise. I chose to leave it out because it is still technically possible to implement a solution _without_ the `Element` class (by duplicating the methods in the `Graph`, `Node` and `Edge` classes). I thought this could be discussion point between a mentor and student or, possibly, for the analyzer. 

Let me know if you'd still prefer me to add `Element`. I'm thinking I'd give them the skeleton rather the full `Element` implementation as the student's task in this one is to implement the graph classes. I think this is different to `tree-building`, which is a a _refactoring_ exercise (so it actually starts with a working solution and the student's task is to try make it better). 

I probably should make `inheritance` a prerequisite though. That way the student will have been introduced to the concept before this exercise.

manumafe98: Maybe [dominoes](https://github.com/exercism/java/tree/main/exercises/practice/dominoes) is a better example of what I was trying to say. They give the `Domino` class implemented and the `Dominoes` to be implemented.

For example if we give `Edge` and `Node` already implemented, the user has multiple options to solve the exercise? or only the one that they use the `Element` class?

config.json
```diff
@@ -1802,6 +1802,16 @@
...
         ],
         "difficulty": 5
+      },
+      {
+        "slug": "dot-dsl",
+        "name": "DOT DSL",
+        "uuid": "03e1070d-a3ed-4664-9559-283e535c6bc4",
+        "practices": [],
+        "prerequisites": [
+          "inheritance"
```
manumafe98, line 1812: We probably should add `lists` as prerequisites as well
```suggestion
          "inheritance",
          "lists"
```

manumafe98 (changes requested):

exercises/practice/dot-dsl/src/main/java/DotDsl.java
kahgoh: I have been thinking about this one the last couple of days. I think providing them with the `Edge` and `Node` implementations allows us to make the assertions easier to follow because we can implement their `equals` and `hashCode`. We also wouldn't need `Element` anymore.

Anyway, I thought I try this out and have just pushed the commits for this in. The commit also:
- Reduces the difficulty as I think the exercise has become easier.
- Changes the "inheritance" prerequisite back to "classes" as inheritance is no longer applicable. 

Let me know what you think.

config.json
kahgoh: I've included this one in the latest commit.

exercises/practice/dot-dsl/src/main/java/DotDsl.java
manumafe98: Love this one!

manumafe98 (approved):

PR #2843, opened 2024-10-24
Remove collection from array description

lcsvoj:
# pull request

Updated the description of Java arrays to avoid confusion with the term "collection" as used in the Java Collections Framework. Previously, the text referred to arrays as collections, which is inaccurate in Java's formal terminology.

Native Java arrays are not part of the Collections Framework, as they do not implement the `Collection` interface or its subtypes, such as `List`, `Set`, or `Map`. They are simply fixed-size objects that hold elements of the same type.

I have rephrased the description to make it more accessible for those who are new to Java, avoiding excessive technical jargon while still preventing misunderstandings and aligning the explanation with Java's official definitions.

---

Reviewer Resources:

### DIFF
--- a/concepts/arrays/about.md
+++ b/concepts/arrays/about.md
@@ -1,7 +1,7 @@
 # About Arrays
 
-In Java, data structures that can hold zero or more elements are known as _collections_.
-An **array** is a collection that has a fixed size and whose elements must all be of the same type.
+In Java, arrays are a way to store multiple values of the same type in a single structure.
+Unlike other data structures, arrays have a fixed size once created.
 Elements can be assigned to an array or retrieved from it using an index.
 Java arrays use zero-based indexing: the first element's index is 0, the second element's index is 1, etc.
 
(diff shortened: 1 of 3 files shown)


kahgoh (commented):
Hi @lcsvoj ! The proposed changes look fine to me and I generally agree with removing the references to "collection" from arrays. 

However, there are a couple of other files we should update as they have the same problem. We should be able to copy and paste the proposed change. Could you also update them? They are:
- [concepts/arrays/about.md](https://github.com/exercism/java/blob/main/concepts/arrays/about.md)
- [exercises/concept/bird-watcher/.docs/introduction.md](https://github.com/exercism/java/blob/main/exercises/concept/bird-watcher/.docs/introduction.md)

lcsvoj:
Hello, @kahgoh!

I have made the requested changes to the pull request #2843, updating the cited files.
Sorry for not noticing the need for those changes before being pointed out.

kahgoh (approved):
Looks good. Thanks for raising the PR!

PR #2851, opened 2024-11-02
Remove multiply and divide methods accepting double from ComplexNumber

jagdish-15:
# pull request

## Description
This pull request removes the `multiply` and `divide` methods that accept a double parameter from the `ComplexNumber` class in both the reference solution and the `main.java` file. These methods were deemed unnecessary since the primary focus of the exercise is to operate on complex numbers, and the existing tests do not validate these double overloads.

## Changes Made:
- Removed `multiply(double factor)` and `divide(double divisor)` methods from the `ComplexNumber` class.
- Updated the reference solution accordingly to ensure consistency.

## Rationale
The double overloads for multiplication and division were not accompanied by corresponding test cases in the canonical data, which could lead to situations where these methods are not adequately tested. By removing these methods, we streamline the class and enhance clarity, ensuring that the exercise focuses on complex number operations as intended.

## Next Steps
Please review the changes and let me know if further modifications are needed.
---

Reviewer Resources:

### DIFF
--- a/exercises/practice/complex-numbers/.meta/src/reference/java/ComplexNumber.java
+++ b/exercises/practice/complex-numbers/.meta/src/reference/java/ComplexNumber.java
@@ -1,7 +1,6 @@
 final class ComplexNumber {
 
     private final double real;
-
     private final double imaginary;
 
     ComplexNumber(double real, double imaginary) {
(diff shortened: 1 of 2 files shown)


exercises/practice/complex-numbers/.meta/src/reference/java/ComplexNumber.java
```diff
@@ -46,12 +48,4 @@ ComplexNumber conjugate() {
     ComplexNumber exponentialOf() {
         return new ComplexNumber(Math.cos(imaginary), Math.sin(imaginary)).multiply(Math.exp(real));
     }
-
-    private ComplexNumber divide(double factor) {
-        return new ComplexNumber(real / factor, imaginary / factor);
-    }
-
-    private ComplexNumber multiply(double factor) {
-        return new ComplexNumber(factor * real, factor * imaginary);
```
kahgoh, line 55: There is a compile error because this method is still being used on line 49 (see the [Lint output](https://github.com/exercism/java/actions/runs/11641679071/job/32422104274?pr=2851#step:4:673)).

kahgoh (changes requested):
Thanks for submitting the PR! I think the change in principle looks good. There's just the compile error from the lint output to fix.

PR #2855, opened 2024-11-04
improve explanation on Java array .length property

melonpan:
# pull request

Hello! I've noticed the sentence "The most commonly used property for arrays is its length" might lead to some misunderstandings, as Java only has a single built-in direct property for arrays. To clarify this, I've made a small adjustment. 

## Changes made:
Replaced 

> "The most commonly used property for arrays is its length which can be accessed like this:"

 with 

> "Java has a built-in `length` property that returns the length of an array, which can be accessed like this:".

Please let me know if you have any questions. Thanks in advance.

---

Reviewer Resources:

### DIFF
--- a/concepts/arrays/about.md
+++ b/concepts/arrays/about.md
@@ -44,7 +44,8 @@ int secondElement = twoInts[1];
 Accessing an index that is outside of the valid indexes for the array results in an `IndexOutOfBoundsException`.
 
 Arrays can be manipulated by either calling an array instance's methods or properties, or by using the static methods defined in the `Arrays` class (typically only used in generic code).
-Java has a built-in `length` property that returns the length of an array, which can be accessed like this:
+The `length` property holds the length of an array.
+It can be accessed like this:
 
 ```java
 int arrayLength = someArray.length;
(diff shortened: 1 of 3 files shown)


concepts/arrays/about.md
```diff
@@ -44,7 +44,7 @@ int secondElement = twoInts[1];
 Accessing an index that is outside of the valid indexes for the array results in an `IndexOutOfBoundsException`.
 
 Arrays can be manipulated by either calling an array instance's methods or properties, or by using the static methods defined in the `Arrays` class (typically only used in generic code).
-The most commonly used property for arrays is its length which can be accessed like this:
+Java has a built-in `length` property that returns the length of an array, which can be accessed like this:
```
kahgoh, line 47: I think the wording sounds like the property is on Java itself. I'd also suggest removing the word "returns" since it is property rather than a method that returns something. For example:

```suggestion
The `length` property holds the length of an array.
It can be accessed like this:
```

kahgoh (changes requested):
Thanks for the PR. I think I agree with the general idea behind this change. I just have a suggestion for the wording.

In addition, could you also update the following files? They also describe the array length:
- the arrays introduction (located in [concepts/arrays/about.md](https://github.com/exercism/java/blob/main/concepts/arrays/introduction.md)
- introduction to the Bird Watcher exercise (located in [exercises/concepts/bird-watcher/.docs/introduction.md](https://github.com/exercism/java/blob/main/exercises/concept/bird-watcher/.docs/introduction.md))

melonpan:
Thanks for the suggestion, I've changed the wording and updated the files.

kahgoh (approved):
Thanks! Changes looks good to me.

PR #2856, opened 2024-11-05
Fix grammar and clarity in reference types explanation

armanh3k:
# pull request

Small change to the grammar to make more sense.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/tim-from-marketing/.docs/introduction.md
+++ b/exercises/concept/tim-from-marketing/.docs/introduction.md
@@ -7,8 +7,8 @@ In Java, the `null` literal is used to denote the absence of a value.
 Primitive data types in java all have a default value and therefore can never be `null`.
 By convention, they start with a lowercase letter e.g `int`.
 
-Reference types contain the memory address of an object can have a value of null.
-These variables usually start with an uppercase e.g `String`.
+Reference types contain the memory address of an object and can have a value of null. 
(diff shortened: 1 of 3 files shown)


exercises/concept/tim-from-marketing/.docs/introduction.md
```diff
@@ -7,8 +7,8 @@ In Java, the `null` literal is used to denote the absence of a value.
 Primitive data types in java all have a default value and therefore can never be `null`.
```
kahgoh, line 7: This is outside your change, but suggest capitalizing the J in Java for consistency. 

```suggestion
Primitive data types in Java all have a default value and therefore can never be `null`.
```

kahgoh (changes requested):
Thanks for pointing out those errors and submitting a PR to fix them! I have a couple minor suggestions. 

In addition, could you also update [concepts/nullability/introduction.md](https://github.com/exercism/java/blob/main/concepts/nullability/introduction.md?plain=1) and [concepts/nullability/about.md](https://github.com/exercism/java/blob/main/concepts/nullability/about.md)? They have a similar problem.

kahgoh (commented):
Thanks for making those updates! I spotted a couple more inconsistencies though.

kahgoh (approved):
Looks good now! Thanks for fixing this!

Issue #2710, opened 2024-02-02
#48in24: Add approaches to Parallel Letter Frequency


sanderploegsma:
The Parallel Letter Frequency exercise on the Java track [is being featured in the #48in24 challenge](https://exercism.org/challenges/48in24/implementation_status), which means that we expect an influx of students attempting to solve it during that week.

It would be nice if the exercise contains some more content by that time.
One example of this is to add one or more [approaches](https://exercism.org/docs/building/tracks/approaches) to the exercise, so that students can dive in deeper after they have solved it themselves.

For an example of how approaches have been added to other practice exercises, take a look at [the approaches for the Leap exercise](https://github.com/exercism/java/tree/main/exercises/practice/leap/.approaches).

masiljangajji:
Hello @sanderploegsma 
Are you still looking for a contributor?
If so, I would like to contribute to this.

kahgoh:
Hi @masiljangajji! Feel free to go for it! I've assigned to you now.

masiljangajji:
Hi @kahgoh , could you review this for me? I’m planning to talk to my friends about open source contribution this Friday, so I’d appreciate it if you could take a look!

kahgoh:
Hi @masiljangajji, thanks for working on this one! Your PR is next on my list and I should get to it later today.

PR #2863, opened 2024-11-12
Add approaches for Parallel Letter Frequency

masiljangajji:
# pull request

Approaches for Parallel Letter Frequency, ready for review!

Resolves https://github.com/exercism/java/issues/2710

---

Reviewer Resources:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,8 +2,6 @@
 
 <!-- Your content goes here: -->
 
-
-
 <!-- DO NOT EDIT BELOW THIS LINE! -->
 ---
 
--- a/exercises/practice/parallel-letter-frequency/.approaches/fork-join/content.md
+++ b/exercises/practice/parallel-letter-frequency/.approaches/fork-join/content.md
@@ -1,0 +1,90 @@
+# `Fork/Join` 
+
+```java
+import java.util.Map;
+import java.util.List;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.RecursiveTask;
+
+class ParallelLetterFrequency {
+
+    List<String> texts;
+    ConcurrentMap<Character, Integer> letterCount;
+
+    ParallelLetterFrequency(String[] texts) {
+        this.texts = List.of(texts);
+        letterCount = new ConcurrentHashMap<>();
+    }
+
+    Map<Character, Integer> countLetters() {
+        if (texts.isEmpty()) {
+            return letterCount;
+        }
+
+        ForkJoinPool forkJoinPool = new ForkJoinPool(); 
+        forkJoinPool.invoke(new LetterCountTask(texts, 0, texts.size(), letterCount));
+        forkJoinPool.shutdown();
+
+        return letterCount;
+    }
+
+    private static class LetterCountTask extends RecursiveTask<Void> {
+        private static final int THRESHOLD = 10; 
+        private final List<String> texts;
+        private final int start;
+        private final int end;
+        private final ConcurrentMap<Character, Integer> letterCount;
+
+        LetterCountTask(List<String> texts, int start, int end, ConcurrentMap<Character, Integer> letterCount) {
+            this.texts = texts;
+            this.start = start;
+            this.end = end;
+            this.letterCount = letterCount;
+        }
+
+        @Override
+        protected Void compute() {
+            if (end - start <= THRESHOLD) {
+                for (int i = start; i < end; i++) {
+                    for (char c : texts.get(i).toLowerCase().toCharArray()) {
+                        if (Character.isAlphabetic(c)) {
+                            letterCount.merge(c, 1, Integer::sum);
+                        }
+                    }
+                }
+            } else {
+                int mid = (start + end) / 2;
+                LetterCountTask leftTask = new LetterCountTask(texts, start, mid, letterCount);
+                LetterCountTask rightTask = new LetterCountTask(texts, mid, end, letterCount);
+                invokeAll(leftTask, rightTask); 
+            }
+            return null;
+        }
+    }
+}
+```
+
+Using [`ConcurrentHashMap`][ConcurrentHashMap] ensures that frequency counting and updates are safely handled in a parallel environment.
+
+If there are no strings, a validation step prevents unnecessary processing.
+
+A [`ForkJoinPool`][ForkJoinPool] is then created. The core of [`ForkJoinPool`][ForkJoinPool] is the Fork/Join mechanism, which divides tasks into smaller units and processes them in parallel.
+
+THRESHOLD is the criterion for task division. If the range of texts exceeds the THRESHOLD, the task is divided into two subtasks, and [`invokeAll`][invokeAll](leftTask, rightTask) is called to execute both tasks in parallel.
+
+Each subtask in LetterCountTask will continue calling compute() to divide itself further until the range is smaller than or equal to the threshold.
+
+For tasks that are within the threshold, letter frequency is calculated. The [`isAlphabetic`][isAlphabetic] method is used to identify all characters classified as alphabetic in Unicode, covering various languages like English, Korean, Japanese, Chinese, etc., returning true for alphabetic characters and false for numbers, special characters, spaces, and others.
+
+Additionally, since uppercase and lowercase letters are treated as the same character (e.g., A and a), each character is converted to lowercase.
+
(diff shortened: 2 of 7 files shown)


exercises/practice/parallel-letter-frequency/.approaches/introduction.md
```diff
@@ -0,0 +1,137 @@
+# Introduction
+
+There are multiple ways to solve the Parallel Letter Frequency problem.
+One approach is to use parallelStream, and another involves using ForkJoinPool.
```
kahgoh, line 4: Looking at the existing approaches for other exercises, such as [grains](https://github.com/exercism/java/blob/main/exercises/practice/grains/.approaches/introduction.md), I think it would be good to format the methods and classes with backticks (``) to maintain consistency in the formatting (example in the suggestion below). Could you go through and update?

```suggestion
One approach is to use `Stream.parallelStream`, and another involves using `ForkJoinPool`.
```

exercises/practice/parallel-letter-frequency/.approaches/introduction.md
```diff
@@ -0,0 +1,137 @@
...
+
+## General guidance
+
+To count occurrences of items, a map data structure is often used, though arrays and lists can work as well. A map, being a key-value pair structure, is suitable for recording frequency by incrementing the value for each key.
```
kahgoh, line 8: We generally format our Markdown one sentence per line. Could you please go through and update the format?

```suggestion
To count occurrences of items, a map data structure is often used, though arrays and lists can work as well. 
A map, being a key-value pair structure, is suitable for recording frequency by incrementing the value for each key.
```

exercises/practice/parallel-letter-frequency/.approaches/fork-join/content.md
```diff
@@ -0,0 +1,88 @@
...
+
+If there are no strings, a validation step prevents unnecessary processing.
+
+A [`ForkJoinPool`][ForkJoinPool] is then created. The core of [`ForkJoinPool`][ForkJoinPool] is the Fork/Join mechanism, which divides tasks into smaller units and processes them in parallel.
+
+THRESHOLD is the criterion for task division. If the range of texts exceeds the THRESHOLD, the task is divided into two subtasks, and [`invokeAll`][invokeAll](leftTask, rightTask) is called to execute both tasks in parallel.
+
+Each subtask in LetterCountTask will continue calling compute() to divide itself further until the range is smaller than or equal to the threshold.
+
+For tasks that are within the threshold, letter frequency is calculated. The [`isAlphabetic`][isAlphabetic] method is used to identify all characters classified as alphabetic in Unicode, covering various languages like English, Korean, Japanese, Chinese, etc., returning true for alphabetic characters and false for numbers, special characters, spaces, and others.
```
kahgoh, line 79: Suggest using `Character.isAlphabetic` here to make it clear the method is from the `Character` class. 

```suggestion
For tasks that are within the threshold, letter frequency is calculated. 
The [`Character.isAlphabetic`][isAlphabetic] method is used to identify all characters classified as alphabetic in Unicode, covering various languages like English, Korean, Japanese, Chinese, etc., returning `true` for alphabetic characters and `false` for numbers, special characters, spaces, and others.
```

kahgoh (changes requested):
Nice, I think this generally looks good! There's just a couple of things to fix in the Markdown.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,8 +2,6 @@
 
 <!-- Your content goes here: -->
 
-
-
```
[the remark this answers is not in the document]
masiljangajji: > I don't think this related to the PR. Could you please undo this change?

the following error will occur.
~~~
.github/PULL_REQUEST_TEMPLATE.md:5 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
.github/PULL_REQUEST_TEMPLATE.md:6 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 3]
~~~

Should we go ahead and cancel it anyway?

masiljangajji:
Hi @kahgoh Thanks for reviewing! 

I’ve incorporated the feedback and committed the changes. However, when I run markdownlint-cli2 on PULL_REQUEST_TEMPLATE.md encounter the error , Please check it

exercises/practice/parallel-letter-frequency/.approaches/introduction.md
```diff
@@ -0,0 +1,150 @@
...
+To count occurrences of items, a map data structure is often used, though arrays and lists can work as well.
+
+A [`map`][map], being a key-value pair structure, is suitable for recording frequency by incrementing the value for each key.
+
+If the data being counted has a limited range (e.g., characters or integers), an `int[] array` or [`List<Integer>`][list] can be used to record frequencies.
+
+Parallel processing typically takes place in a multi-[`thread`][thread] environment.
+
+The Java 8 [`stream`][stream] API provides methods that make parallel processing easier, including the parallelStream() method.
+
```
kahgoh, line 18:
```suggestion
```

kahgoh (commented):
Thanks for the updates! There are still some places where I think we should place backticks. I've gone ahead and added them as suggestions.

I wasn't sure if there was misunderstanding in regards my earlier comment about having a sentence per line as I can see there are now empty lines between each sentence. While each sentence should start on its own line, there should only be empty lines between _paragraphs_.

For example, this from your first commit was fine:
```

```

It looks like this:
For example, this from your first commit was fine:
```

There are multiple ways to solve the Parallel Letter Frequency problem.
One approach is to use `parallelStream`, and another involves using `ForkJoinPool`.
```

It renders like this:
> There are multiple ways to solve the Parallel Letter Frequency problem.  One approach is to use `parallelStream`, and another involves using `ForkJoinPool`.

But this makes two paragraphs:
```
There are multiple ways to solve the Parallel Letter Frequency problem.

One approach is to use `parallelStream`, and another involves using `ForkJoinPool`.
```

It looks like this:
> There are multiple ways to solve the Parallel Letter Frequency problem.
>
> One approach is to use `parallelStream`, and another involves using `ForkJoinPool`.

Notice the first example looks like one paragraph, whereas the second looks like two.

See other approaches, like the [collatz conjecture](https://github.com/exercism/java/blob/main/exercises/practice/collatz-conjecture/.approaches/introduction.md?plain=1) or [Difference of squares](https://github.com/exercism/java/blob/main/exercises/practice/difference-of-squares/.approaches/formula/content.md?plain=1) for examples on how they are formatted. 

I thought the arrangement of paragraphs or empty lines from your first commit was fine - we just needed to move the sentences to start on its own line. I have added some suggestions where the empty lines can be removed to turn them back to paragraphs. Feel free to go through and update other spots accordingly.

masiljangajji:
Hi @kahgoh 
I'm not very familiar with English, so I made a few mistakes. Thank you for the detailed review!

exercises/practice/parallel-letter-frequency/.approaches/fork-join/content.md
```diff
@@ -0,0 +1,91 @@
...
+Using [`ConcurrentHashMap`][ConcurrentHashMap] ensures that frequency counting and updates are safely handled in a parallel environment.
+
+If there are no strings, a validation step prevents unnecessary processing.
+
+A [`ForkJoinPool`][ForkJoinPool] is then created.
+The core of [`ForkJoinPool`][ForkJoinPool] is the Fork/Join mechanism, which divides tasks into smaller units and processes them in parallel.
+
+`THRESHOLD` is the criterion for task division.
+If the range of texts exceeds the `THRESHOLD`, the task is divided into two subtasks, and [`invokeAll(leftTask, rightTask)`][invokeAll] is called to execute both tasks in parallel.
+Each subtask in LetterCountTask will continue calling compute() to divide itself further until the range is smaller than or equal to the `THRESHOLD`.
```
kahgoh, line 78: Suggest clarifying that `compute` is called through `invokeAll`. Your example only shows `invokeAll` being called, not `compute`, so it might be confusing. Also, suggest putting the `LetterCountTask` and `compute` in backticks for formatting.

```suggestion
Each subtask in `LetterCountTask` will continue calling `compute()` (via `invokeAll(leftTask, rightTask)`) to divide itself further until the range is smaller than or equal to the `THRESHOLD`.
```

kahgoh (commented):
Thanks! I think we're almost there! Just spotted a couple of minor things.

masiljangajji:
@kahgoh  Thank you! Is there anything else that could be improved?

masiljangajji:
Thank you so much for the detailed review. This is my first time contributing to open source, and since I'm not very familiar with English, I made a lot of mistakes. Thanks to you, I was able to have a great experience!

PR #2877, opened 2024-12-10
Update tests two bucket

jagdish-15:
# Pull Request

This PR adds 3 new tests to the Two Bucket exercise to sync with the `problem-specifications` repository. Two of the newly added tests check for error handling, which was not tested before. I am also unsure about the error message—should I keep it as it is in the `problem-specifications` repo or modify it to be more informative? 

Let me know your thoughts on this.

---

Reviewer Resources:


kahgoh:
I think with the error messages I think it is fine to leave them out of the exception. But, I think we should make our own exception instead of reusing `IllegalArgumentException`. `IllegalArgumentException` is used to indicate when one of the arguments to the method is bad (like if if one of them is negative). It doesn't suit cases where the arguments are good, but we can't reached the required amount. If we can't reach the value, I'd use a different exception, perhaps making a custom one for the exercise.

jagdish-15:
@kahgoh, thank you for the suggestion—it does make a lot of sense to create a custom exception for this purpose.
I imagine it would look something like this:

```java
public class UnreachableGoalException extends RuntimeException {
    public UnreachableGoalException() {
        super();
    }

    public UnreachableGoalException(String message) {
        super(message);
    }

    public UnreachableGoalException(String message, Throwable cause) {
        super(message, cause);
    }
}
```

That said, when I think about this, we've used `IllegalArgumentException` in similar scenarios before—like when the target amount couldn't be reached with the available denominations in the "Change" exercise. I've noticed this approach across many other exercises as well.

While your suggestion is undoubtedly a more "proper" and robust approach, it might feel a little inconsistent with the conventions used elsewhere. What are your thoughts on this? How do you think I should proceed?

kahgoh:
I think there is also precedence for making our own exceptions for our exercises, like [BinarySearch](https://github.com/exercism/java/blob/main/exercises/practice/binary-search/src/test/java/BinarySearchTest.java) and [TreeBuilding](https://github.com/exercism/java/blob/main/exercises/practice/tree-building/src/test/java/BuildTreeTest.java). But yeah, the think the exception would look something like that.

jagdish-15:
Sure, I'll get right on it as soon as possible, since I already have the body of the custom Exception ready. However, I'll implement it the way it's done in exercises like "Binary Search" and "Tree Building," instead of following the approach in "Calculator Conundrum," as that causes build failure issues for Windows users (as highlighted in the forum discussion).  

I was also considering implementing similar custom Exceptions in exercises like "Change," since an unassignable change value shouldn’t technically throw an `IllegalArgumentException`. If you have other exercises in mind that would benefit from custom Exceptions, feel free to let me know. We could even extend this improvement across the entire track—I’d be more than happy to take on that task!

kahgoh:
> However, I'll implement it the way it's done in exercises like "Binary Search" and "Tree Building," instead of following the approach in "Calculator Conundrum," as that causes build failure issues for Windows users (as highlighted in the forum discussion). 

Sounds good. Feel free to update "Calculator Conundrum" as well.

> I was also considering implementing similar custom Exceptions in exercises like "Change," since an unassignable change value shouldn’t technically throw an `IllegalArgumentException`

I'm reluctant to make this change in the other exercises. On one hand, I agree that there are other exercises where a custom exception makes sense. _But_ on the other hand, wouldn't it mean solutions that students have already submitted become invalid if they use the custom exception? A lot of the community solutions would become invalid or outdated. The change would likely require the system to re-run tests on existing solutions, which incurs a cost (if you're interested, there is a note about expensive operations in [Avoiding triggering unnecessary test runs](https://exercism.org/docs/building/tracks#h-avoiding-triggering-unnecessary-test-runs)). I'm not quite sure if the benefit is worth the cost from making the change.

jagdish-15:
> Sounds good. Feel free to update "Calculator Conundrum" as well.

Great! I'll get to updating "Calculator Conundrum" and this exercise as soon as my exams end—sorry for the delay!


Regarding the change to custom exceptions in other exercises, I completely agree with your point about the cost and the potential impact on existing solutions. Updating all exercises to use custom exceptions would indeed risk invalidating many community solutions, requiring a system-wide test rerun, which can be resource-intensive.

For this particular exercise, however, the situation seems different. Since the tests have been updated from the problem-specification repository, community solutions for this exercise are already going to be invalidated. If updates from the problem-specification repo are frequent and similar changes would cause solutions to become outdated anyway, we could consider making this adjustment now.

That said, it really comes down to the priority and whether the benefits of introducing custom exceptions outweigh the potential disruption. Let me know what you decide!

kahgoh:
> For this particular exercise, however, the situation seems different. Since the tests have been updated from the problem-specification repository, community solutions for this exercise are already going to be invalidated. If updates from the problem-specification repo are frequent and similar changes would cause solutions to become outdated anyway, we could consider making this adjustment now.

I agree, I think updating the Two Buckets exercise to meet the problem specifications and using the custom exception is fine. Its only updating the _other_ exercises to use a custom exception like "Change" I'm reluctant to if they currently "work" with `IllegalArgumentException`.

kahgoh:
> For this particular exercise, however, the situation seems different. Since the tests have been updated from the problem-specification repository, community solutions for this exercise are already going to be invalidated. If updates from the problem-specification repo are frequent and similar changes would cause solutions to become outdated anyway, we could consider making this adjustment now.

I agree, I think updating the Two Buckets exercise to meet the problem specifications and using the custom exception is fine. Its only updating the _other_ exercises to use a custom exception like "Change" I'm reluctant to if they currently "work" with `IllegalArgumentException`.

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -75,4 +76,52 @@ public void testBucketOneSizeTwoBucketTwoSizeThreeStartWithOne() {
...
 
     }
+
+    @Disabled("Remove to run test")
+    @Test
+    public void testReachingGoalIsImpossible() {
+
+        TwoBucket twoBucket = new TwoBucket(6, 15, 5, "one");
+
+        assertThatExceptionOfType(IllegalArgumentException.class)
```
kahgoh, line 86: I think it is a bit odd to have to have to assert that we get an exception three times in the test (I was expecting only one assertion). What do you think about refactoring to assert the exception just once? In the [csharp](https://github.com/exercism/csharp/blob/main/exercises/practice/two-bucket/TwoBucketTests.cs) implementation, the `TwoBucket` class has a `measure` method that the tests calls. It returns a "Result" object containing the number of moves, the final bucket and the other bucket or throws an exception if the goal can be reached.

jagdish-15: > I think it is a bit odd to have to have to assert that we get an exception three times in the test

Even I found it a bit odd to assert the exception three times while implementing the tests but couldn’t think of a better alternative. However, your suggestion of returning a `Result` object and handling everything within a single method is an excellent idea! This approach not only simplifies the implementation but also ensures the custom exception can be thrown effectively.

kahgoh: Yeah, I think we can go ahead with this one as we both seem to agree it would be better this way.

kahgoh: Yeah, I think we can go ahead with this one as we both seem to agree it would be better this way.

jagdish-15:
@kahgoh, 

I've updated the solution so the main class now returns a `Result` object instead of using three separate methods to retrieve individual variables. Additionally, the constructor now throws an exception for impossible bucket configurations, which I believe is a more intuitive approach compared to delegating this responsibility to the `getResult` method.

I'm currently working on implementing the custom exception as we discussed. Let me know your thoughts!

jagdish-15:
@kahgoh,

I’ve updated the solution to include the custom exception as discussed. The constructor now throws this exception for impossible bucket configurations.

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -2,77 +2,110 @@
...
+
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    public void testReachingGoalIsImpossible() {
+        
+        assertThatExceptionOfType(UnreachableGoalException.class)
+                .isThrownBy(() -> new TwoBucket(6, 15, 5, "one"))
+                .withMessage("impossible");
```
kahgoh, line 86: Now that we have the `UnreachableGoalException`, I think it would be enough to assert on the exception type.

```suggestion
```

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -2,77 +2,110 @@
...
+        assertThat(bucketResult.getOtherBucket()).isEqualTo(2);
+
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    public void testReachingGoalIsImpossible() {
+        
+        assertThatExceptionOfType(UnreachableGoalException.class)
+                .isThrownBy(() -> new TwoBucket(6, 15, 5, "one"))
```
kahgoh, line 85: I think we want to call the `getResult()` in the assertion because I think it is also valid to do calculations in the `getResults` method instead of the constructor.

```suggestion
                .isThrownBy(() -> new TwoBucket(6, 15, 5, "one").getResult())
```

jagdish-15: I’ve updated the assertion to call `getResult()` as suggested. Let me know if there’s anything else you’d like adjusted!

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
jagdish-15: I included the error message initially because the `problem-specifications` repository specified it, but I can update it if you'd prefer. Let me know what you think!

jagdish-15: I see your point about the cross-language nature of the problem specification. It makes sense to focus on asserting the exception type since "impossible" as a message doesn't provide much context and might not translate well to other languages or implementations.

While I agree a more descriptive message like "Cannot reach target" would be helpful, it might not be strictly necessary here. Sticking to `UnreachableGoalException` seems cleaner and still conveys the intent clearly. Thanks for clarifying—it feels like the right balance to avoid overcomplicating the assertion.

kahgoh (approved):
Looks good now. Thanks @jagdish-15!

PR #2879, opened 2024-12-17
Add text clarifying test executable permissions

gs-101:
# pull request

This PR adds a section on how to add executable permissions to the `gradlew` file, used to run tests.

I find it interesting how no one else seems to be facing this issue, but I always face it on Java when downloading exercises (which I do by using [exercism.el](https://github.com/anonimitoraf/exercism.el)).

---

Reviewer Resources:

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -81,6 +81,20 @@ Good luck! Have fun!
 
    _(Don't worry about the tests failing, at first, this is how you begin each exercise.)_
 
+   If you get the following error:
+
+   ```sh
+   ./gradlew: Permission denied
+   ```
+
+   Then the file is missing the execute permission. To fis this, run:
+
+   ```sh
+   chmod +x ./gradlew
+   ```
+
+   And now you should be able to run the previous command.
+
 4. Solve the exercise. Find and work through the `instructions.append.md` guide ([view on GitHub][hello-world-tutorial]).
 
 Good luck! Have fun!


docs/TESTS.md
```diff
@@ -81,6 +81,16 @@ Good luck! Have fun!
 
    _(Don't worry about the tests failing, at first, this is how you begin each exercise.)_
 
+   If you get the following error:
+
+   ./gradlew: Permission denied
```
kahgoh, line 86: Could we put this in a code block (using the triple back ticks ```), like the earlier snippets?

docs/TESTS.md
```diff
@@ -81,6 +81,16 @@ Good luck! Have fun!
...
+
+   Then that means there's an issue regarding permissions, that the file isn't executable. To fix this, run:
+
+   chmod +x ./gradlew
```
kahgoh, line 90: Similar, could we put this in a code block?

kahgoh (commented):
Thanks for the PR! Just a couple of minor comments to make the formatting more consistent.

docs/TESTS.md
gs-101: Sure! Let me just rebase this.

gs-101: Ah, the previous ones had blocks, I didn't see it because I have markup turned off 😅.

docs/TESTS.md
```diff
@@ -81,6 +81,20 @@ Good luck! Have fun!
 
    _(Don't worry about the tests failing, at first, this is how you begin each exercise.)_
 
+   If you get the following error:
+
+   ```sh
+   ./gradlew: Permission denied
+   ```
+
+   Then that means there's an issue regarding permissions, that the file isn't executable. To fix this, run:
```
kahgoh, line 90: I wanted to suggest a more direct wording for the first sentence. What do you think? 

```suggestion
   Then the file is missing the execute permission. To fix this, run:
```

gs-101: I agree, this wording seems better.

kahgoh (approved):
Thanks, it looks good now!

PR #2883, opened 2024-12-28
Fix test name for binary search exercise

### LINKED ISSUE #2881: Binary Search Exercise: Fix test name
The test name `aValueLargerThanTheArraysSmallestValueIsNotFound` should be changed to `aValueLargerThanTheArraysLargestValueIsNotFound`, because the value 13 to be tested is larger than the arrays largest value 11.

akmaekki:
# 2881

In JAVA track improve test names for binary search exercise.

Fixes #2881 

Reviewer Resources:

### DIFF
--- a/exercises/practice/binary-search/src/test/java/BinarySearchTest.java
+++ b/exercises/practice/binary-search/src/test/java/BinarySearchTest.java
@@ -119,7 +119,7 @@ public class BinarySearchTest {
     @Disabled("Remove to run test")
     @Test
     public void nothingIsFoundWhenTheLeftAndRightBoundCross() throws ValueNotFoundException {
-        List<Integer> sortedList = List.of(2, 1);
+        List<Integer> sortedList = List.of(1, 2);
 
         BinarySearch search = new BinarySearch(sortedList);
 
--- a/exercises/practice/binary-search/src/test/java/BinarySearchTest.java
+++ b/exercises/practice/binary-search/src/test/java/BinarySearchTest.java
@@ -94,7 +94,7 @@ public class BinarySearchTest {
 
     @Disabled("Remove to run test")
     @Test
-    public void aValueLargerThanTheArraysSmallestValueIsNotFound() throws ValueNotFoundException {
+    public void aValueLargerThanTheArraysLargestValueIsNotFound() throws ValueNotFoundException {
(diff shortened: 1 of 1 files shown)


exercises/practice/binary-search/src/test/java/BinarySearchTest.java
```diff
@@ -119,7 +119,7 @@ public void nothingIsFoundInAnEmptyArray() throws ValueNotFoundException {
     @Disabled("Remove to run test")
     @Test
     public void nothingIsFoundWhenTheLeftAndRightBoundCross() throws ValueNotFoundException {
-        List<Integer> sortedList = List.of(1, 2);
+        List<Integer> sortedList = List.of(2, 1);
```
akmaekki, line 122: I can revert this again, but the test name suggests to me that right boundary (maximum) is less than the left boundary (minimum).

kahgoh: Hey @akmaekki, the data for the tests comes from the [problem specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/binary-search/canonical-data.json). As far as I can tell, the `List.of(1, 2)` matches the problem specification and so does the name. I think the name may be referring to way binary search works - left and right starts at the end and beginning of the list. At each step, compare the value you're looking for and update the "left" and "right" indexes depending on the comparison result. Repeat. If the value isn't in the list, the "left" and "right" will eventually cross over. Although, I'm little unsure why the value 0 is chosen as the value to find - it looks like the same case as `aValueSmallerThanTheArraysSmallestValueIsNotFound`.

If you feel there is still something wrong with the data for this test and would like to change, could you please open a discussion in the [forums](https://forum.exercism.org/)? Generally, we should fix the data in the problem specifications so that it can eventually be fixed in all tracks with this exercise, but we usually discuss changes to the problem specifications in the forums first. 

For this PR, I'd suggest backing this out for now to unblock it. If we agree to fix it in the forums, we can also do another PR.

akmaekki: Hi @kahgoh 
Thanks for the elaborate discussion and tip with the forum. I am fine with backing this out. Will remove the changes done on `nothingIsFoundWhenTheLeftAndRightBoundCross` and refresh this PR.

kahgoh (approved):
This looks good now! Thanks!

PR #2884, opened 2024-12-28
Write constants in UPPER_CASE 

josealonso:
# Fix constants casing in the functional approach for the Darts exercise

---

Reviewer Resources:


josealonso:
@kahgoh, let me know if I'm missing something.
Thanks :)

kahgoh (approved):
Thanks for raising this PR! I agree with the changes here, as I think it is more idiomatic to use SCREAMING_CASE for static constants and to write `private static final` instead of `final private static`.

PR #2893, opened 2025-01-02
Explain modern switch block

josealonso:
# Explain modern switch block

Text added to the switch block concept, explaining the modern switch statement and the switch expression.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/football-match-reports/.docs/introduction.md
+++ b/exercises/concept/football-match-reports/.docs/introduction.md
@@ -35,7 +35,7 @@ switch (direction) {
 
 ## Modern Switch Statements
 
-The switch statement was improved in the latest Java versions. 
+The switch statement was improved in the latest Java versions.
 
 - The `break` keyword is not needed and the arrow operator is used instead of the semicolon.
 - Multiple case values can be provided in a single case statement.
--- a/exercises/concept/football-match-reports/.docs/introduction.md
+++ b/exercises/concept/football-match-reports/.docs/introduction.md
@@ -32,3 +32,63 @@ switch (direction) {
         break;
 }
 ```
+
+## Modern Switch Statements
+
+The switch statement was improved in the latest Java versions. 
+
+- The `break` keyword is not needed and the arrow operator is used instead of the semicolon.
+- Multiple case values can be provided in a single case statement.
+
+```java
+String direction = getDirection();
+switch (direction) {
+    case "left" -> goLeft();
+    case "right" -> goRight();
+    case "top", "bottom" -> goStraight();
+    default -> markTime();
+}
+```
+
+The first LTS (Long Term Support) version that had these improvements was Java 17, released on September, 2021.
+
+Other improvement is that the case values can be any object.
+
+## Switch Expressions
+
+Going even further, the switch block is now an expression, meaning it returns a value.
+
+```java
+return switch (day) {   // switch expression
+    case "Monday", "Tuesday", "Wednesday", "Thursday", "Friday" -> "Week day";
+    case "Saturday", "Sunday" -> "Weekend";
+    default -> "Unknown";
+};
+```
+
+instead of using a switch statement:
+
+```java
+String day = "";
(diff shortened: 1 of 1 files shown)


josealonso:
I didn't want to include `pattern matching`. I think it's quite easy to read.
I welcome your feedback, @kahgoh @manumafe98 

Thanks :)

manumafe98:
Hey @josealonso Thanks for the contribution! You have to actually make the update [here](https://github.com/exercism/java/blob/main/concepts/switch-statement/introduction.md) as you can see in the .docs directory we have a [.md.tpl file](https://github.com/exercism/java/blob/main/exercises/concept/football-match-reports/.docs/introduction.md.tpl) that points to that one.

What i'm thinking is that the usage of switch would depend on what version of java the students use, makes sense to teach both the newer and older version but seems weird to called them like modern or older version, maybe specify in what versions of java the `modern` one can be applied and in which of them not?

manumafe98 (changes requested):
Update the introduction in the correct file

kahgoh:
I think it might be confusing to mention both versions in the instructions - are students expected to use the "old" or the "new" switch statements in the exercise? If we're adding to this exercise, I'd suggest we could add/update the exercise to practice both styles. Alternatively, we've already mention switch expressions in the [about.md](https://github.com/exercism/java/blob/main/concepts/switch-statement/about.md). The `about.md` is for adding extra information like this (see [about.md documentation](https://exercism.org/docs/building/tracks/concepts#h-file-about-md)).

josealonso:
> Hey @josealonso Thanks for the contribution! You have to actually make the update [here](https://github.com/exercism/java/blob/main/concepts/switch-statement/introduction.md) as you can see in the .docs directory we have a [.md.tpl file](https://github.com/exercism/java/blob/main/exercises/concept/football-match-reports/.docs/introduction.md.tpl) that points to that one.
> 
> What i'm thinking is that the usage of switch would depend on what version of java the students use, makes sense to teach both the newer and older version but seems weird to called them like modern or older version, maybe specify in what versions of java the `modern` one can be applied and in which of them not?

Thanks @manumafe98. I'm sorry I didn't modify the correct document. Regarding the Java versions, that is explained in the text I wrote.

josealonso:
> I think it might be confusing to mention both versions of the `switch` statement in the instructions - are students expected to use the "old" or the "new" switch statements in the exercise? If we're adding to this exercise, I'd suggest we could add/update the exercise to practice both styles. Alternatively, we've already mention switch expressions in the [about.md](https://github.com/exercism/java/blob/main/concepts/switch-statement/about.md). The `about.md` is for adding extra information about the concepts after they finish the exercise (see [about.md documentation](https://exercism.org/docs/building/tracks/concepts#h-file-about-md)).
> 
> Other than, I'd suggest using the terms "switch _statement_" and "switch _expressions_" as I believe they commonly used to distinguish between the two.

That's a good point. I didn't know that was possible, sorry. The best option would be to update the exercise to practice both styles, since the *traditional* switch block is probably the most common one. But that requires to use the Analyzer and I still have to look into that API. 
So for the time being, I'll move the text I wrote to the `about.md` file.
The *modern* switch statement is not an expression, but it's different from the *traditional* switch block. In Java 17 the switch expression is in preview stage, I think.

Thanks for the feedback @kahgoh

kahgoh:
> So for the time being, I'll move the text I wrote to the about.md file.

That sounds good.

> In Java 17 the switch expression is in preview stage, I think.

Ah, I think I missed [Patterns for Switch (JEP  441)](https://openjdk.org/jeps/441), which was added in Java 21. Switch expression came out of preview in Java 14 (according to [this Wikipedia page](https://en.wikipedia.org/wiki/Java_version_history)).

josealonso:
Many thank for the link to the docs @kahgoh. I went trough several pages and I'm impressed it's so well-documented.

josealonso:
Replaced by #2899 @kahgoh

PR #2903, opened 2025-01-14
Update FighterTest.java

NotsOverflow:
fixed  compilation error

# pull request

---

Reviewer Resources:


NotsOverflow:
@kahgoh  yes, on my system it will not compile and wen i sent the result it didn't compile. ( i'm using exercism localy )

kahgoh:
By chance, did you happen to add `public` to the `isVulnerable` and `getDamagePoints` methods in the `Fighter` class in your solution? 
When you start the exercise, they should look like [this file](https://github.com/exercism/java/blob/main/exercises/concept/wizards-and-warriors/src/main/java/Fighter.java) (the methods aren't `public`). If you added the `public` keyword to the methods, then you'll get the compile error. If not, could you please share your solution?

PR #2904, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

kahgoh (changes requested):
Hi @IsaacG , I think this is the incorrect append. It says that the order _is_ important, but the assertions here `containsExactlyInAnyOrder`, which allows the anagrams to be listed in any order.

IsaacG:
> Hi @IsaacG , I think this is the incorrect append. It says that the order _is_ important, but the assertions here `containsExactlyInAnyOrder`, which allows the anagrams to be listed in any order.

Apologies and thank you for catching that. I updated the PR.

kahgoh (approved):

PR #2908, opened 2025-01-28
Update cars-assemble tasks description

tomasnorre:
# pull request

As discussion in the forum: https://forum.exercism.org/t/cars-assemble-exercise-java/15229
the function naming the cars assemble exercise is a little unfortunate, as it doesn't indicate that task one should return working cars, but only produced cars. 

Changing the function name in the exercise, will break ~11K solutions, so I suggest to extend the note, to help the users get this right.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/cars-assemble/.docs/instructions.md
+++ b/exercises/concept/cars-assemble/.docs/instructions.md
@@ -13,14 +13,14 @@ You have two tasks.
 
 ## 1. Calculate the production rate per hour
 
-Implement the `CarsAssemble.productionRatePerHour()` method to calculate the assembly line's production rate per hour, taking into account its current assembly line's speed :
+Implement the `CarsAssemble.productionRatePerHour()` method to calculate the assembly line's production rate per hour, taking into account its current assembly line's speed and success rate:
 
 ```Java
 CarsAssemble.productionRatePerHour(6)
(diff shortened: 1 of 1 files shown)


exercises/concept/cars-assemble/.docs/instructions.md
```diff
@@ -20,7 +20,7 @@ CarsAssemble.productionRatePerHour(6)
 // => 1193.4
 ```
 
-Note that the value returned is a `double`.
+Note that the value returned is a `double`. You will need to take the success rate into account to get the correct result.
```
kahgoh, line 23: Hey @tomasnorre thanks for raising this number and taking into account the number of breaking solutions!

I noticed the end of line 16 also talks about something that students to have to take into account. I suggest updating this task so that its all in one place. For example, instead of adding this sentence here, we could add to line 16 like:

```
Implement the `CarsAssemble.productionRatePerHour()` method to calculate the assembly line's production rate per hour, taking into account its current assembly line's speed _and_ success rate:
```

tomasnorre: I like your suggestion better than mine :)

kahgoh (approved):
Looks good! Thanks again for the PR!

PR #2914, opened 2025-02-17
Add maps concept

### LINKED ISSUE #2583: Add concept: `maps`
The goal of this issue is to add a new concept to the Java track that introduces `maps`. The corresponding concept exercise can probably be based on the "High Score" exercise from the [Elixir track](https://github.com/exercism/elixir/tree/main/exercises/concept/high-score). 

Instructions on how to contribute to the Java track and how to implement concept exercises can be found in the [contributing guide](https://github.com/exercism/java/blob/main/CONTRIBUTING.md).

As always, leave a comment here if you want to work on this exercise, and we'll assign it to you. In case you have any further questions, feel free to ask here.

kahgoh:
# Add maps concept

Resolves #2583 

---

Reviewer Resources:


concepts/maps/about.md
```diff
@@ -0,0 +1,167 @@
+# About Maps
+
+A **Map** is a data structure for storing key value pairs.
+It is similar to dictionaries in other programming languages.
+The [Map][map-javadoc] interface defines operations on a map.
+
+Java has a number Map implementations.
```
josealonso, line 7: Typo: number **of**

josealonso (commented):
@kahgoh, a very complete explanation and well-written exercise!

concepts/maps/about.md
```diff
@@ -0,0 +1,166 @@
...
+Unlike `HashMap`, they populate the map upfront and become read-only once created.
+[Map.copyOf][map-copyof-javadoc] makes a read-only copy of a map.
+
+```java
+Map<String, Integer> readOnlyFruitPrices = Map.copyOf(fruitPrices);
+```
+
+Calling methods like `put`, `remove` or `clear` results in an `UnsupportedOperationException`.
+
+[map-javadoc]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/HashMap.html
```
manumafe98, line 155: I think you referred to this url instead
```suggestion
[map-javadoc]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html
```

concepts/maps/about.md
```diff
@@ -0,0 +1,166 @@
+# About Maps
+
+A **Map** is a data structure for storing key value pairs.
+It is similar to dictionaries in other programming languages.
+The [Map][map-javadoc] interface defines operations on a map.
```
manumafe98, line 5: I think this way its more clear
```suggestion
The [Map][map-javadoc] interface defines the operations you can make with a map.
```

concepts/maps/about.md
```diff
@@ -0,0 +1,166 @@
...
+
+## HashMap
+
+Java has a number of different Map implementations.
+[HashMap][hashmap-javadoc] is a commonly used one.
+
+```java
+// Make an instance
+Map<String, Integer> fruitPrices = new HashMap<>();
```
manumafe98, line 14: Maybe we could add a note or something to clarify why we are using the interface to type the map instead of using ``HashMap`` directly? because it could be that not all the students know it is a best practice in case you want to switch your map implementation, or if it is explain in other part of exercism redirect them to that as well

concepts/maps/introduction.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+You can use the [keys] or [values] methods to obtain the keys or the values in a Map as a Set or collection respectively.studentScores
+
+```java
+fruitPrices.keys();    // Returns "apple" and "pear" in a set
+fruitPrices.values();  // Returns 100 and 80, in a Collection
+```
+
+[map-javadoc]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/HashMap.html
```
manumafe98, line 64: same as the other one I think you referred to this one
```suggestion
[map-javadoc]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html
```

exercises/concept/international-calling-connoisseur/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
+# Instructions
+
+In this exercise you'll be wring code to manage a dictionary of international dialing codes using a `Map`.
```
manumafe98, line 3:
```suggestion
In this exercise you'll be writing code to manage a dictionary of international dialing codes using a `Map`.
```

manumafe98 (changes requested):
Good job! @kahgoh this one will be very useful to the students of exercism!

concepts/maps/about.md
kahgoh: That's a good catch! I definitely meant to refer to `Map` instead of `HashMap`.

exercises/concept/international-calling-connoisseur/src/main/java/DialingCodes.java
```diff
@@ -0,0 +1,37 @@
+import java.util.HashMap;
+import java.util.Map;
+
+public class DialingCodes {
+
+    private final Map<Integer, String> countryByDialingCode = new HashMap<>();
```
kahgoh, line 6: @manumafe98 I'm wondering whether I should remove this? I had originally put this here to make it clearer where the `Map` is defined, but then again `classes` is set as a prerequisite (so students could probably work out to define a field). I would add a hint if removed. 

Another might option is to replace this with something like:
```java
// TODO: Make a Map instance
private final Map<Integer, String> countryByDialingCode = null;
```

Do you have any thoughts on what we should do here?

concepts/maps/about.md
kahgoh: With this one, I decided to add it to the `about.md`, but not the `introduction.md` (to keep the introduction simpler).

exercises/concept/international-calling-connoisseur/src/main/java/DialingCodes.java
manumafe98: Yeah I will remove it too, I think if they reach this part of the track they should know were to define it.
And also it give us leverage to work on an extra actionable if for example a student does something like this

```java
    public Map<Integer, String> getCodes() {
        Map<Integer, String> countryByDialingCode = new HashMap<>();
        return Map.copyOf(countryByDialingCode);
    }
```
To check that they initialize the map as a field and not inside the methods

manumafe98 (approved):

exercises/concept/international-calling-connoisseur/src/main/java/DialingCodes.java
kahgoh: Cool! I'll remove it from the stub and update the hints.

josealonso: I agree with @manumafe98. Students have already defined a list in a previous learning exercise.

josealonso (commented):
Very easy to understand explanation on the `hash` value. Good work!!

josealonso (approved):

kahgoh:
@manumafe98, I added a commit since you approved. Just wanted to check if it looks good to you before merging.

manumafe98 (approved):

PR #2916, opened 2025-02-25
Add missing assertion from updateDialingCode test

gigaSproule:
# pull request

I noticed that the new map concept had a test with a missing assertion. I'm assuming this is the expected outcome based on the test name.

---

Reviewer Resources:


kahgoh (approved):
That's a good spot! Thanks for raising the PR to fix!

PR #2922, opened 2025-03-04
Update the introduction

chrollorifat:
# pull request

This commit will:
- Update the introduction with beginner-friendly information
- Fix some minor grammar

---

Reviewer Resources:

### DIFF
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -24,7 +24,7 @@ class Calculator {
 }
 ```
 
-A function within a class is referred to as a _method_. Each _method_ can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from _methods_ using the `return` keyword. To allow a _method_ to be called by other classes, the `public` access modifier must be added.
+A function within a class is referred to as a _method_. Each method can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Values are returned from methods using the `return` keyword. To allow a method to be called by other classes, the `public` access modifier must be added.
 
 ```java
 class Calculator {
@@ -34,10 +34,10 @@ class Calculator {
 }
 ```
 
-Invoking a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
+Invoking/calling a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
 
 ```java
-int sum = new Calculator().add(1, 2);
+int sum = new Calculator().add(1, 2);  // here the  "add" method has been called to perform the task of addition 
 ```
 
 Scope in Java is defined between the `{` and `}` characters.
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -37,7 +37,8 @@ class Calculator {
 Invoking (or calling) a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
 
 ```java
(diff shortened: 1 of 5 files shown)


jagdish-15 (approved):
I think these changes look good. @kahgoh, any thoughts or feedback?

concepts/basics/introduction.md
```diff
@@ -34,10 +34,10 @@ class Calculator {
 }
 ```
 
-Invoking a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
+Invoking/calling a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
```
kahgoh:
```suggestion
Invoking (or calling) a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
```

concepts/basics/introduction.md
```diff
@@ -1,18 +1,18 @@
 # Introduction
 
-Java is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable is defined by explicitly specifying its type.
+Java is a statically-typed language, which means that the type of a variable is known at compile-time. Assigning a value to a name is referred to as defining a variable. A variable is defined by explicitly specifying its type.
```
kahgoh, line 3: Reading the this paragraph and the one after the example, I thought it might read better we explain the type can't be changed here instead of later.

```suggestion
Java is a statically-typed language, which means that the type of a variable is known at compile-time. Assigning a value to a name is referred to as defining a variable. A variable is defined by explicitly specifying its type. The variable's type can not be changed once it is defined.
```

concepts/basics/introduction.md
```diff
@@ -1,18 +1,18 @@
...
 
 ```java
 int explicitVar = 10;
 ```
 
-Updating a variable's value is done through the `=` operator. Once defined, a variable's type can never change.
+Updating a variable's value is done through the `=` operator. Here, `=` does not represent mathematical equality. It simply assigns a value to a variable. Once defined, a variable's type can never change.
```
kahgoh, line 9: I think it might more friendly if we just explain what `=` does (not sure if the "mathematical equality" will be understood by everyone).  

```suggestion
The `=` operator assigns a value to a variable. It can also be used to update a variable's value.
```

concepts/basics/introduction.md
```diff
@@ -34,10 +34,10 @@ class Calculator {
 }
 ```
 
-Invoking a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
+Invoking/calling a method is done by specifying its class and method name and passing arguments for each of the method's parameters.
 
 ```java
-int sum = new Calculator().add(1, 2);
+int sum = new Calculator().add(1, 2);  // here the  "add" method has been called to perform the task of addition
```
kahgoh: This suggestion is mainly to reduce the width (reducing chance of needing side scrolling)

```suggestion
// call "add" method to perform addition 
int sum = new Calculator().add(1, 2);  
```

concepts/basics/introduction.md
chrollorifat: But we have to make the distinction between the equals to and assigning operator clear though. I've seen people not really grasping the difference between them when they start out and they usually take it as a equals to sign.

jagdish-15:
Yes, the changes have already been implemented in the `main` branch. Some parts are implemented slightly differently, but they essentially achieve the same result.

As @kahgoh mentioned, this PR is no longer needed. It’s likely that the `main` branch was updated after this PR was originally opened (on March 4), which is why it now overlaps with existing changes.

That said, thank you @chrollorifat for taking the time to raise the PR! Your effort is appreciated, and we definitely welcome further contributions in the future! 🚀

PR #2928, opened 2025-03-16
Updating test.toml for Flatten-Array

jagdish-15:
# Pull Request

**@kahgoh**, I noticed that the test descriptions were changed after [this discussion](https://forum.exercism.org/t/flatten-array-mentions-both-lists-and-arrays/15849/14). I'm unsure whether the changes were only meant to update the test descriptions and the `introduction.md` file or if they were intended to modify the entire exercise structure—specifically, replacing lists with arrays and requiring a `flatten` function that accepts an arbitrarily nested array. 

Since Java does not support arrays containing mixed data types (like in dynamically typed languages), I assume these updates are just documentation changes. Let me know what you think!

---

Reviewer Resources:


kahgoh:
This exercise is a bit problematic since we've used Lists instead of arrays and it looks like the problem specs were updated to just say arrays. I think to allow us to resync this exercise (along with the instructions), we should add an `instructions.append.md` to explain we're using a list instead of an array. It doesn't seem right to just update the `test.toml` because it changes the test description to use the word "array' instead of "list".

jagdish-15:
@kahgoh, I've synced the docs for this exercise with **problem-specifications** and added `instructions.append.md` as requested. Let me know what the heading for this file should be, as it's required by `Lint`!

exercises/practice/flatten-array/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+~~~~exercism/note
```
kahgoh, line 1: Looking at other `instructions.append.md`, this is the heading we just use.

```suggestion
# Instructions append
```

Could you also please remove the `~~~` at the end to match the other `instructions.append.md`?

exercises/practice/flatten-array/.meta/config.json
```diff
@@ -9,6 +9,7 @@
     "jagdish-15",
     "jmrunkle",
     "jtigger",
+    "kahgoh",
```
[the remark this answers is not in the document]
jagdish-15: Sure!

exercises/practice/flatten-array/.docs/instructions.append.md
jagdish-15: Sure, I was trying to implement the special blocks we have on Exercism, but staying consistent with other append files is important!

kahgoh (approved):

PR #2941, opened 2025-04-27
strings - Expand concept by newline and text blocks

fapdash:
# pull request

- Add multi-line strings / text blocks (Java 15)
- Add explanation of newlines / end of line chars, so students can understand the motivation behind text blocks

---

Noticed that the String concept was lacking text blocks.
While there are quite some details and intricacies to text blocks, they're not relevant for the level of detail the syllabus is aiming for, therefore it doesn't make sense to have them as an extra concept imo.
I added infos about newlines so students can appreciate the motivation behind text blocks.

Looking forward to your feedback. :)

---

Reviewer Resources:

### DIFF
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -22,6 +22,29 @@ String escaped = "c:\\test.txt";
 // => c:\test.txt
 ```
 
+To put a newline character in a string, use the `\n` escape code (`\r\n` on Windows):
+
+```java
+"<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n"
+```
+
+For code that should work on different operating systems Java offers [`System.lineSeparator()`][system-line-separator], which returns the system-dependent line separator string.
+
+To comfortable work with texts that contain a lot of newlines you can use [Text Blocks](text-blocks).
+These multi-line strings are delimited by triple double quote (`"`) characters.
+
+
+```java
+String multilineHtml = """
+<html>
+    <body>
+        <h1>Hello, World!</h1>
+    </body>
+</html>
+""";
+// => "<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n"
+```
+
 Finally, there are many ways to concatenate a string.
 The simplest one is the `+` operator:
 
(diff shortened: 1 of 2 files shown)


fapdash:
Not sure if system dependent newlines are too much information for this concept?

I've looked at how other tracks are testing multi-line strings. Elixir uses [highschool sweetheart](https://exercism.org/tracks/elixir/exercises/high-school-sweetheart), which asks you to print a multi-line ascii heart. but I'd rather not swap out the whole exercise.

Maybe tell the students that the system should be able to produce a "fancy" message with the log level?

```
logLevelMessage("warning");
=>
┌───
│   Exercism Log System
│   Level: warning
└───
```

concepts/strings/about.md
```diff
@@ -22,6 +22,29 @@ String escaped = "c:\\test.txt";
 // => c:\test.txt
 ```
 
+To put a newline character in a string, use the `\n` escape code (`\r\n` on Windows):
+
+```java
+"<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n"
+```
+
+For code that should work on varying operating systems Java offers [`System.lineSeparator()`][system-line-separator], which returns the system-dependent line separator string.
```
kahgoh: Wouldn't using the `System.lineSeparator` require string concatenation that comes later? I wonder if it would flow better if this section came after string concatenation. Other than that, I think it would be handy to have an example showing how to use this.

concepts/strings/about.md
```diff
@@ -22,6 +22,29 @@ String escaped = "c:\\test.txt";
...
+This is important if you're writing to files that will be read on the same system.
+
+To comfortable work with texts that contain a lot of newlines you can use [Text Blocks][text-blocks].
```
kahgoh:
```suggestion
To comfortably work with texts that contain a lot of newlines you can use [Text Blocks][text-blocks].
```

concepts/strings/about.md
```diff
@@ -22,6 +22,29 @@ String escaped = "c:\\test.txt";
...
+
+```java
+String multilineHtml = """
+<html>
+    <body>
+        <h1>Hello, World!</h1>
+    </body>
+</html>
+""";
+// => "<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n"
```
kahgoh, line 42: I thought the result string line looks a little strange with it all being on one line with the `\n` because the multi-line template is expressed over multiple lines. Perhaps we could remove it?

```suggestion
```

concepts/strings/about.md
```diff
@@ -22,6 +22,29 @@ String escaped = "c:\\test.txt";
 // => c:\test.txt
 ```
 
+To put a newline character in a string, use the `\n` escape code (`\r\n` on Windows):
+
+```java
+"<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n"
```
kahgoh: Thought to assign to variable to keep consistent with the other examples.

```suggestion
String multilineHtml = "<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n";
```

concepts/strings/about.md
fapdash: Mh, I think it's valuable/educational to show how the string ends up looking once evaluated, don't you think?

concepts/strings/about.md
fapdash: I've moved that paragraph further down and added an example in https://github.com/exercism/java/pull/2941/changes/e713aaef239be6879339ae5716035343ae9a7b68

fapdash: But we could expand the log-levels task and add `[no important files changed]`, no?

I also thought you could add optional tasks to a concept, which are not required to progress to the next concept. Can't find it in the docs rn though. Did I dream that up?

Edit: seems what I'm thinking of are practice exercises getting associated with a concept

concepts/strings/about.md
```diff
@@ -22,6 +22,26 @@ String escaped = "c:\\test.txt";
 // => c:\test.txt
 ```
 
+To put a newline character in a string, use the `\n` escape code (`\r\n` on Windows):
+
+```java
+String multilineHtml = "<html>\n    <body>\n        <h1>Hello, World!</h1>\n    </body>\n</html>\n";
+```
+
```
kahgoh, line 30: Something still seems weird about the flow because we revisit line separators later. Perhaps we could add a sentence like _Alternatively, we can use `String.format` or  ``System.lineSeparator()` to add system independent line separator (discussed later)._

concepts/strings/about.md
kahgoh: Hmm ... this is so long ago ... Yeah, I think it would be useful to show how it ends up looking and I'm struggling to think of a better way to show this. Ok, let's leave this for now.

PR #2942, opened 2025-04-28
Add tests to check when a language is added several times

ValMati:
- Test that verifies that a language added twice is counted only once
- Test that proves that a language added twice is correctly deleted

# pull request

Although it is not made explicit in the exercise, I believe that it should be understood that a language should be added only once.

Looking forward to your feedback. :)

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/karls-languages/.meta/config.json
+++ b/exercises/concept/karls-languages/.meta/config.json
@@ -2,6 +2,9 @@
   "authors": [
     "jmrunkle"
   ],
+  "contributors": [
+    "Valmati"
+  ],
   "files": {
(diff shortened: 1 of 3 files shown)


kahgoh:
I am going to decline this one. I don't see the point in adding this requirement. This is an exercise on Lists. In Java, if you want a collection where each item can be appear only once, I'd suggest using a Set instead.

kahgoh:
I am going to decline this one. I don't see the point in adding this requirement. This is an exercise on Lists. In Java, if you want a collection where each item can be appear only once, I'd suggest using a Set instead.

PR #2945, opened 2025-05-25
added non solution files in config.json to skip analysis

Swathibal:
# pull request

This PR fixed this issue https://github.com/exercism/java-analyzer/issues/132.

1. Read only files are added under files.editor to skip the analysis

---

Reviewer Resources:


kahgoh:
Sorry @Swathibal, but this change isn't it. The goal of the issue is to update the analyzer so that it doesn't analyze the files listed in the `editor` in the JSON (or even better, just analyze the solution files), _not_ to add things to them in the JSON.

Issue #2714, opened 2024-02-02
#48in24: Implement analyzer for Bank Account


sanderploegsma:
The Bank Account exercise on the Java track [is being featured in the #48in24 challenge](https://exercism.org/challenges/48in24/implementation_status), which means that we expect an influx of students attempting to solve it during that week.

It would be nice if the exercise contains some more content by that time.
One example of this is to implement an [analyzer](https://exercism.org/docs/building/tooling/analyzers) for the exercise, to provide automated feedback on submitted solutions.

Implementing an exercise analyzer is done in three steps:

1. Create (or update) the `.meta/design.md` file for the exercise explaining what type of feedback the analyzer should provide ([example for the Leap exercise](https://github.com/exercism/java/blob/main/exercises/practice/leap/.meta/design.md)).
2. Add the copy for the analyzer comments to [exercism/website-copy](https://github.com/exercism/website-copy/tree/main/analyzer-comments/java).
3. Implement the analyzer rules in [exercism/java-analyzer](https://github.com/exercism/java-analyzer).

PR #2946, opened 2025-05-29
bank-account: describe Analyzer feedback in .meta/design.md #2714

masiljangajji:
# pull request

closes  https://github.com/exercism/java/issues/2714

The goal is to maintain thread safety while throwing exceptions for any exceptional situations that may arise.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/bank-account/.meta/design.md
+++ b/exercises/practice/bank-account/.meta/design.md
@@ -2,7 +2,7 @@
 
 ## Analyzer
 
-This exercise could benefit from the following rules in the [analyzer](https://github.com/exercism/java-analyzer):
+This exercise could benefit from the following rules in the [analyzer]:
 
 - `essential`: Verify that the solution **guards `deposit`, `withdraw`, and `getBalance` with `synchronized`** to avoid
   race conditions.
@@ -12,3 +12,5 @@ This exercise could benefit from the following rules in the [analyzer](https://g
 - `essential`: If the solution **does not throw an exception when `amount` exceeds `balance` during withdrawal**,
   instruct the student to add this check.
 - `actionable`: Verify that the solution **declares exactly one `boolean`/`Boolean` field to track account state**; if the count is zero or greater than one, instruct the student to use exactly one such field.
+
+[analyzer]: https://github.com/exercism/java-analyzer


exercises/practice/bank-account/.meta/design.md
```diff
@@ -0,0 +1,16 @@
+# Design
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution **guards `deposit`, `withdraw`, and `getBalance` with `synchronized`** to avoid
```
kahgoh, line 7: I think the problem with checking that the methods are `synchronized` it that there are different ways of achieving the same thing with different trade offs. For example:
- [Lock objects](https://docs.oracle.com/javase/tutorial/essential/concurrency/newlocks.html) (`ReentrantLock`/`ReadWriteLock`) (this [solution](https://exercism.org/tracks/java/exercises/bank-account/solutions/liuchang919106840729) uses locks)
- [Synchronized statements](https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html) (using intrinsic locks)

The instructions also hint at this:

> To pass the last test you might find the [synchronized keyword or locks](https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html) useful.

I think each method has their own strengths and weaknesses and there may be more I haven't thought of. We probably should check that they use at least one of these techniques to guard concurrent access, but I'm not sure how easy/hard that will be...

exercises/practice/bank-account/.meta/design.md
```diff
@@ -0,0 +1,16 @@
...
+  race conditions.
+- `essential`: Verify that the solution **checks the account’s open/closed state before every balance operation**
+- `essential`: If the solution **allows negative amounts** in `deposit` or `withdraw`, instruct the student to reject
+  them early.
+- `essential`: If the solution **does not throw an exception when `amount` exceeds `balance` during withdrawal**,
+  instruct the student to add this check.
```
kahgoh, line 13: Looking at the tests, these are already covered by the following test cases (so the analyzer doesn't need to check for them):
- `cannotWithdrawMoreThanWasDeposited`
- `cannotWithdrawNegativeAmount`
- `cannotDepositNegativeAmount`
- `cannotWithdrawMoreThanWasDeposited`

exercises/practice/bank-account/.meta/design.md
```diff
@@ -0,0 +1,16 @@
...
+- `actionable`: Verify that the solution **declares exactly one `boolean`/`Boolean` field to track account state**; if the count is zero or greater than one, instruct the student to use exactly one such field.
```
kahgoh, line 14: I'm somewhat unsure about this one. Most solutions do use a boolean to track whether an account is open or not, but there are also other ways. For example, I have seen [one](https://exercism.org/tracks/java/exercises/bank-account/solutions/Ataro) that use an `AtomicBoolean` and [another](https://exercism.org/tracks/java/exercises/bank-account/solutions/felaz) that uses a string (although, I wouldn't recommend using a string...). Some also prefer [Enums](https://exercism.org/tracks/java/exercises/bank-account/solutions/AndrasLengyel) over booleans (like [this solution](https://exercism.org/tracks/java/exercises/bank-account/solutions/AndrasLengyel))

kahgoh (commented):
Thanks for looking into this and for the suggestions! My thoughts on this one in the comments below...

PR #2956, opened 2025-06-13
Add test for `hasDayWithoutBirds` method

rabestro:
# pull request

- Added a test to validate the behavior of `hasDayWithoutBirds` when the last day has zero visits. 
- Updated the `lastWeek` field to be final and align with best practices.

https://forum.exercism.org/t/bird-watcher-in-java-should-we-add-a-test-case/17933

---

Reviewer Resources:


jagdish-15 (approved):
@kahgoh The forum topic doesn’t appear to have a clear resolution. What do you think, should this be merged with the `[no important files changed]` label, or not?

@rabestro Would love to hear your thoughts on this as well.

kahgoh (approved):
As suggested on the forums, I think we can add this test and disable re-running of the tests.

PR #2962, opened 2025-06-29
Fix International Calling Connoisseur hints

### LINKED ISSUE #2957: Fix International Calling Connoisseur hints
The hints for [International Calling Connoisseur](https://github.com/exercism/java/blob/main/exercises/concept/international-calling-connoisseur/.docs/hints.md) hints at using the `values()` method to extract the values from the map. However, the [exemplar solution](https://github.com/exercism/java/blob/main/exercises/concept/international-calling-connoisseur/.meta/src/reference/java/DialingCodes.java) shows this method is never used. It probably should hint at using `entries()` or `keys()` instead.

kuryuliya:
…-calling-connoisseur

- Change hint #5 from suggesting values() to entrySet()
- Update description to explain entrySet() allows searching both keys and values
- Fix link reference to point to correct Map API documentation

Closes #2957

---

Reviewer Resources:


exercises/concept/international-calling-connoisseur/.docs/hints.md
```diff
@@ -22,7 +22,7 @@
 
 ## 5. Find a country's dialing code
 
-- There is a [way][map-values-docs] to get an iterable collection of values in a map.
+- There is a [way][map-entry-set-docs] to get an iterable collection of entries in a map, which allows you to search through both keys and values.
```
kahgoh: Thanks for taking this on! My only thought with the wording is that it sounds like we could use `entrySet()` to go through the keys independently of the values (which you could, but then there are other methods for that). What do you think of describing it as going through the key-value pairs (the introduction describes maps in terms of this pair)? I think it could be something like this 👇🏻  

```suggestion
- There is a [way][map-entry-set-docs] to get an iterable collection of entries in a map, which allows you to go through the key-value pairs.
```

kuryuliya: Thank you for the feedback! Yes, your version sounds better, it will definitely be less ambiguous this way.

kahgoh (approved):

PR #2965, opened 2025-07-01
Add missing test to CryptoSquareTest (54-character input)

Snehitha9126:
This PR adds a missing canonical test case for the 'crypto-square' exercise based on the uuid `33fd914e-fa44-445b-8f38-ff8fbc9fe6e6` from problem-specifications. It verifies that a 54-character plaintext results in 8 chunks with trailing spaces.

Reference: https://github.com/exercism/java/issues/2959

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This repository contains the source for the exercises of the Java track on Exerc
 Next to the exercises, the Java track also consists of the following tooling:
 
 - [exercism/java-test-runner] - The Exercism [test runner][docs-test-runners] for the Java track that automatically verifies if a submitted solution passes all of the exercise's tests.
-- [exercism/java-representer] - The Exercism [representer][docs-representers] for the Java track that creates normalized representations of submitted solutions.
+- [exercism/java-representer]a - The Exercism [representer][docs-representers] for the Java track that creates normalized representations of submitted solutions.
 - [exercism/java-analyzer] - The Exercism [analyzer][docs-analyzers] for the Java track that automatically provides comments on submitted solutions.
 
 ## Contributing Guide
--- a/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
+++ b/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
@@ -66,8 +66,10 @@ public class CryptoSquareTest {
 
         assertThat(cryptoSquare.getCiphertext()).isEqualTo(expectedOutput);
     }
+    // Legacy test from older canonical-data.json – consider removing if outdated
 
-    @Disabled("Remove to run test")
+
+    /*@Disabled("Remove to run test")
     @Test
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ This repository contains the source for the exercises of the Java track on Exerc
 Next to the exercises, the Java track also consists of the following tooling:
 
 - [exercism/java-test-runner] - The Exercism [test runner][docs-test-runners] for the Java track that automatically verifies if a submitted solution passes all of the exercise's tests.
-- [exercism/java-representer] - The Exercism [representer][docs-representers] for the Java track that creates normalized representations of submitted solutions.
+- [exercism/java-representer]a - The Exercism [representer][docs-representers] for the Java track that creates normalized representations of submitted solutions.
```
kahgoh, line 13: I think is accidental. Could you please revert this change?

exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
kahgoh, line 1: Looking at the [history for the problem specifications](https://github.com/exercism/problem-specifications/commit/e7055f98d7d6ae21c44b6bae38b2a1917ace8676), it looks like we just need to rename the test method as the inputs, calls and expectations are still the same. Could you please update this change to just rename the test method (i.e. the diff should just show the `fiftyFourCharacterPlaintextResultsInSevenChunksWithTrailingSpaces` test method renamed to `fiftyFourCharacterPlaintextResultsInEightChunksWithTrailingSpaces` - the "Seven" is changed "Eight")?

kahgoh (changes requested):
Thanks for helping to update the tests! I have some comments below 👇🏻. In addition, could you please also update the exercise's [test.toml](https://github.com/exercism/java/blob/main/exercises/practice/crypto-square/.meta/tests.toml) by running `configlet sync -e crypto-square -u`? When it asks if you want to replace an existing test case, just choose yes and it should update tests.toml for you. 

To get configlet, run the [bin/fetch-configlet](https://github.com/exercism/java/blob/main/bin/fetch-configlet) script if you are on Linux or see [configlet releases](https://github.com/exercism/configlet/releases) for Windows or MacOS.

jagdish-15:
You'll also need to update the `.toml` file using `configlet sync`

exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
```diff
@@ -66,14 +66,26 @@ public void eightCharacterPlaintextResultsInThreeChunksWithATrailingSpace() {
 
         assertThat(cryptoSquare.getCiphertext()).isEqualTo(expectedOutput);
     }
+    // Legacy test from older canonical-data.json – consider removing if outdated
 
-    @Disabled("Remove to run test")
+
+    /*@Disabled("Remove to run test")
```
xinri, line 72: Please don't keep comments in the code. if we want to retrieve it, we can get from the history

PR #2969, opened 2025-07-04
Add intergalactic transmission

### LINKED ISSUE #2967: Add intergalactic-transmission exercise
The [intergalactic-transmission](https://github.com/exercism/problem-specifications/tree/main/exercises/intergalactic-transmission) exercise was recently added to the problem specifications. This task is to add the practice exercise to the Java track.

jagdish-15:
# pull request

closes #2967 

---

Reviewer Resources:


jagdish-15:
@kahgoh, let me know if there are any changes you'd like me to make! Also, could you please advise on the appropriate difficulty level to set for this exercise?

kahgoh:
The difficulty can be subjective at times 😅. The easiest is using the same as other tracks (6 on C#, Vlang and x86-64 assembly). 

Not so much a change, but a question - I noticed the data is given as `List<Integer>`, although the story and instruction talks in terms of bytes. I was wondering if you considered using Java's `byte` instead (e.g. `List<Byte>`)?

jagdish-15:
> The difficulty can be subjective at times 😅. The easiest is using the same as other tracks (6 on C#, Vlang and x86-64 assembly).

Sure, I'll make the change!


> Not so much a change, but a question - I noticed the data is given as `List<Integer>`, although the story and instruction talks in terms of bytes. I was wondering if you considered using Java's `byte` instead (e.g. `List<Byte>`)?

Actually, since `List.of()` infers types based on the arguments, we'd need to cast each element to `byte` for it to work with `List<Byte>`. So the current implementation `List<Integer> expected = List.of(0x03, 0x00);` would have to be updated to `List<Byte> expected = List.of((byte) 0x03, (byte) 0x00);`.
Do you know any workaround for this?

kahgoh:
Try using [Arrays.asList](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-T...-) (e.g. `List<Byte> theList = Arrays.asList(new Byte[]{0x01, 0x03});`).

Alternatively, it might be easier to use a byte array (e.g. `byte[] data = new byte[]{0x01, 0x02}`) directly instead of a list and using byte array for binary data is common in Java.

jagdish-15:
In Java, `byte` is signed by default (unlike C#), which means it only supports values from `-128` to `127`. So for any value above `127` (i.e. from `0x80` to `0xFF`), we must explicitly cast it using `(byte)`, and it's stored in 2's complement form.

Manually figuring out and casting each of these values can become tedious and error-prone, especially when writing long byte arrays in tests or encoding logic.

Is there a cleaner or more efficient way to handle such input in Java?

jagdish-15:
Yes, I was thinking the same, will add an `instruction.append.md`!

jagdish-15:
@kahgoh 
Could you take a look and let me know if it needs to be reframed?

exercises/practice/intergalactic-transmission/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+# Instructions append
+
+For the Java track, the input will be provided as `List<Integer>` instead of an `List<Byte>`, and the output should also be a `List<Integer>`.
```
kahgoh: A couple of thoughts on this one:
- I don't think we need to mention "Java track".
- Also, I think `List<Byte>` should be `byte[]`. If there is need to store binary data in low level data structure, `byte[]` is commonly used (not `List<Byte>`). There are other data structures like [ByteBuffer](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/ByteBuffer.html), but even then, it has methods that take data from `byte[]`.

Not sure if it's too much information, but we could explain its to help avoid dealing with casting or converting between "negative" bytes. Suggestion below:

```suggestion
Although we're dealing with bytes of data, the inputs and outputs are `List<Integer>` (instead of `byte[]`) to avoid the need to cast or convert negative values for bytes 128 to 255.
```

exercises/practice/intergalactic-transmission/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+# Instructions append
+
+Although we're dealing with bytes of data, the inputs and outputs are `List<Integer>` (instead of `byte[]`) to avoid the need to cast or convert negative values for bytes 128 to 255.
```
jagdish-15, line 3: @kahgoh, what about something like this?

```suggestion
Although we're dealing with bytes of data, the inputs and outputs are `List<Integer>` (instead of `byte[]`) since Java doesn't support unsigned bytes, and values between 128 and 255 would otherwise require explicit casting or conversion.
```

kahgoh (approved):
Thanks, it looks good to me now!

Issue #1813, opened 2020-04-19
paasio: implement exercise


jmrunkle:
The exercise **paasio** has not been implemented yet for the Java track. 
The description of the exercise can be found in the [problem specification repository](https://github.com/exercism/problem-specifications/tree/master/exercises/paasio). 

How to implement a new exercise for the Java track is described in detail in [CONTRIBUTING.md](https://github.com/exercism/java/blob/master/CONTRIBUTING.md#adding-a-new-exercise). Please have a look there first before starting working on the exercise. 
Also please make sure it is clear that you are currently working on this issue, either by asking to be assigned to it, or by opening an empty PR. 

When opening an PR, please reference this issue using any of the [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords). 

If you have had a look at the exercise description and you concluded that the exercise might not be possible to implement in the Java language, please leave a comment and describe the problem. 

In case you have any further questions, feel free to ask here.

CliffRoberson:
This is my first attempt at implementing an exercise and it seems like a good one to try. May I have this assigned to me and do you know of another language which has this exercise that I can copy from?

mirkoperillo:
Thank you @CliffRoberson for the help.
I found a [go implementation](https://github.com/exercism/go/tree/master/exercises/paasio) of this problem if it can help you.

Feel free to ask for every doubt

andrerfcsantos:
@CliffRoberson Still interested in working on this?

alex-md:
I would be happy to implement this exercise.

shreyasgosavi:
Heyy I can give it a try...

kahgoh:
Hi @shreyasgosavi, sure go for it! I'll assign it to you. Check this issue's description for links to guidance. Let us know if you have any further questions.

shreyasgosavi:
Hii sure you can assign the task to me

Pavankumar1407:
Hii @kahgoh I would like to try this, May I ?

shreyasgosavi:
Hello guys, will surely work on it. !!

Ayush-bit852:
Is This still open or the issue has been fixed ? 
@kahgoh May  I try ?

shreyasgosavi:
Hello guyss sorry for the delay, will try to do it by this weekend,else assign it to another member..

shreyasgosavi:
Hii was starting working on this issue, just need to understand what is expected, the flow and when should we generate the report.

[...]

PR #2970, opened 2025-07-05
Passio Practice Exercise implemented

shreyasgosavi:
# pull request

Hello, raising PR for paasio exercise implementation. Issue ID : #1813

---

Reviewer Resources:


exercises/practice/paasio/src/main/java/Paasio.java
kahgoh, line 1: This file should just contain enough of the stub for students to start. Have a look at the stub for the [Poker](https://github.com/exercism/java/blob/main/exercises/practice/poker/src/main/java/Poker.java) for example. It only contains the bare methods for students to fill.

When the student starts the Paasio exercise, the student will start with the contents of this file. They should not start with the solution.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+        try (Paasio customFileReader = new Paasio(this.dataInputStream, this.dataOutputStream)) {
+            customFileReader.read();
+            customFileReader.read();
+            customFileReader.read();
+            assertThat(customFileReader.getReadOperationCount()).isEqualTo(3);
+        } catch (IOException ioException) {
+            System.out.println("Exception occured");
+        }
+
+    }
```
kahgoh: Don't catch the exception in the tests, othewise the tests will pass if an exception is thrown. Instead, let propagate the exception so that JUnit also sees the exception.

```suggestion
    void checkReadOperationCount() throw Exception {

        try (Paasio customFileReader = new Paasio(this.dataInputStream, this.dataOutputStream)) {
            customFileReader.read();
            customFileReader.read();
            customFileReader.read();
            assertThat(customFileReader.getReadOperationCount()).isEqualTo(3);
        }
    }
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
+import org.junit.jupiter.api.*;
+
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.*;
```
kahgoh: Please don't use `*` imports, as per other tests ([poker](https://github.com/exercism/java/blob/main/exercises/practice/poker/src/test/java/PokerTest.java) for example).

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+        try (Paasio fileOperations = new Paasio(this.dataInputStream, this.dataOutputStream)) {
+
+            fileOperations.read();
+            fileOperations.read();
+            fileOperations.read();
+            fileOperations.read();
+
+            byte[] readByteData = new byte[5];
+            fileOperations.read(readByteData);
+            assertThat(fileOperations.getBytesRead()).isEqualTo(9);
```
kahgoh: I think it would be interesting to assert on the contents `readByteData`.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+            customFileReader.read();
+            customFileReader.read();
+            assertThat(customFileReader.getReadOperationCount()).isEqualTo(3);
+        } catch (IOException ioException) {
+            System.out.println("Exception occured");
+        }
+
+    }
+
+    @Test
```
kahgoh, line 41: In general, only the first test should be "enabled" in practice exercises. Students enable the test as they work through the exercise.

```suggestion
    @Disabled("Remove to run test")
    @Test
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+    void checkIfNullPointerExceptionIsThrownWhenNullIsPassedInsteadOfByteArray() {
+
+        assertThatThrownBy(() -> {
+            Paasio fileOperations = new Paasio(this.dataInputStream, this.dataOutputStream);
+            fileOperations.read(null);
+        }).isInstanceOf(NullPointerException.class);
```
kahgoh: I think this test case isn't really needed - giving `read` a `null` value isn't particularly interesting.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+    void validateReadNBytesReadOperationCount() {
+
+        try (Paasio fileOperations = new Paasio(this.dataInputStream, this.dataOutputStream)) {
+
+            fileOperations.readNBytes(5);
+            fileOperations.readNBytes(5);
+
+            assertThat(fileOperations.getReadOperationCount()).isEqualTo(2);
+            System.out.println("haha ");
+            System.out.println(fileOperations.getReadOperationCount());
```
kahgoh: Please remove all `System.out.println`.

```suggestion
            System.out.println("haha ");
            System.out.println(fileOperations.getReadOperationCount());
```

config.json
```diff
@@ -1869,6 +1869,14 @@
...
         ],
         "difficulty": 10
+      },
+      {
+        "slug": "paasio",
+        "name": "PaaS I/O",
+        "uuid": "81177978-2ed3-47c2-a6a0-fef316d94c0b",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
```
kahgoh: This really shouldn't be a one (it's not that easy). Perhaps we could use the Go track as a guide - they have used 4 as the difficulty.

exercises/practice/paasio/.meta/src/reference/java/Paasio.java
```diff
@@ -0,0 +1,140 @@
+import java.io.*;
```
kahgoh, line 1: Please don't use `*` imports. For example:

```suggestion
import java.io.Closeable;
import java.io.InputStream;
import java.io.OutputStream;
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.*;
+
+public class PaasioTest {
+
+    private ByteArrayInputStream dataInputStream;
+    private ByteArrayOutputStream dataOutputStream;
```
kahgoh, line 20: Since every test creates and uses a Paasio object, we could also use a Passio field and set it in `setUpTest` and then close it after every test in a `@AfterEach` method.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,337 @@
...
+    private static final String MESSAGECONSTANT = "This is additional Content.";
+
+    @BeforeEach
+    public void setUPTest(TestInfo testInfo) {
```
kahgoh:
```suggestion
    public void setUpTest(TestInfo testInfo) {
```

kahgoh (changes requested):

shreyasgosavi:
Thanks for time and that's really good explanation will do it 🤘

exercises/practice/paasio/src/main/java/Paasio.java
```diff
@@ -0,0 +1,98 @@
+import java.io.*;
```
jagdish-15, line 1: As mentioned earlier, please don't use `*` for import

exercises/practice/paasio/.meta/src/reference/java/Paasio.java
```diff
@@ -0,0 +1,149 @@
...
```
jagdish-15, line 1: Same goes for this!

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        paasioReaderWriter.read();
+        paasioReaderWriter.read();
+        paasioReaderWriter.read();
+        paasioReaderWriter.read();
+
+        assertThat(paasioReaderWriter.getBytesRead()).isEqualTo(4);
+    }
+
+    @Test
```
jagdish-15, line 61: Every test except the first one should have `@Disabled("Remove to run test")`

config.json
```diff
@@ -1900,6 +1900,14 @@
           "lists"
         ],
         "difficulty": 10
+      },
+      {
```
jagdish-15, line 1958: Could you please move this exercise up with the other ones that have a difficulty of 4, and make sure it stays in alphabetical order within that section?

jagdish-15 (changes requested):

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    @AfterEach
+    public void closeOpenStream() throws IOException {
+        if (!Objects.isNull(this.paasioReaderWriter.getInputStream())) {
+            this.paasioReaderWriter.getInputStream().close();
+        }
+        if (!Objects.isNull(this.paasioReaderWriter.getOutputStream())) {
+            this.paasioReaderWriter.getOutputStream().close();
+        }
```
kahgoh, line 38: I think this could be simplified to simply calling `passioReaderWriter.close()` and might allow you to remove `getInputStream` and `getOutputStream` methods from the `Passio` class.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+        paasioReaderWriter.read();
+
+        byte[] readByteData = new byte[4];
+        paasioReaderWriter.read(readByteData);
+
+        //Asserting content of the data read
+        assertThat(readByteData).isEqualTo(" is ".getBytes());
+        assertThat(paasioReaderWriter.getBytesRead()).isEqualTo(8);
+
+
```
kahgoh, line 76: There seems to be extra spacing here.

```suggestion
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    @Test
+    @Disabled("Remove to run test")
+    void checkIfDataFollowsTheOffsetWhileWritingInByteArray() throws IOException {
+
+        byte[] initialMessage = "Hello! ".getBytes();
+        byte[] messageArray = Arrays.copyOf(initialMessage, 50);
+
+        int length = messageArray.length - initialMessage.length - 2;
```
kahgoh, line 98: I wasn't sure - why do we need to subtract 2 here?

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+        assertThat(paasioReaderWriter.getBytesRead()).isEqualTo(8);
+
+
+    }
+
+    @Test
+    @Disabled("Remove to run test")
+    void checkByteValueReadFromTheFile() throws IOException {
+
+        byte[] fileData = new byte[100];
```
kahgoh, line 83: I think this `fileData` name came from an earlier iteration where we had a distinction between file and network data. However, we no longer have this distinction, so I'd suggest renaming it. Perhaps just `data`?

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Test
+    @Disabled("Remove to run test")
+    void validateReadNBytesDataReadInBytesAndVerifyReadCount() throws IOException {
+
+        byte[] helloArray = "Hello! ".getBytes();
+        byte[] dataRead = Arrays.copyOf(helloArray, 100);
+
+        int bytesRead = paasioReaderWriter.read(dataRead, helloArray.length, 40);
+        String finalVAlue = new String(dataRead, 0, helloArray.length + bytesRead, StandardCharsets.UTF_8);
```
kahgoh, line 156:
```suggestion
        String finalValue = new String(dataRead, 0, helloArray.length + bytesRead, StandardCharsets.UTF_8);
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        assertThat(bytesRead).isEqualTo(10);
+    }
+
+    @Test
+    @Disabled("Remove to run test")
+    void validateAllBytesReadingData() throws IOException {
+
+        byte[] bytes = paasioReaderWriter.readAllBytes();
+        assertThat(paasioReaderWriter.getBytesRead()).isEqualTo(bytes.length);
```
kahgoh, line 121: I'd also suggest asserting the contents of `bytes` to check the contents that were read back.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+
+    @Test
+    @Disabled("Remove to run test")
+    void validateBytesOfDataReadWhileReadingItFromOffset() throws IOException {
+        byte[] byteData = new byte[50];
+        int bytesRead = paasioReaderWriter.read(byteData, 0, 10);
+
+        assertThat(bytesRead).isEqualTo(10);
```
kahgoh, line 113: I think we should also assert the **contents** `byteData` as it should write only 10 bytes to the array. This should be less than the available content? 

Should there also be a test for when the length (or the "10") is more than what is available?

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    }
+
+
+    @Test
+    @Disabled("Remove to run test")
+    void validateReadNBytesReadOperationCount() throws IOException {
+
+        paasioReaderWriter.readNBytes(5);
+        paasioReaderWriter.readNBytes(5);
```
kahgoh, line 143: I think we are also missing tests that assert on the `byte[]` that is returned by the `readNBytes` method.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+import static org.assertj.core.api.Assertions.*;
+
+public class PaasioTest {
+
+    private Paasio paasioReaderWriter;
+
+    private ByteArrayInputStream dataInputStream;
+    private ByteArrayOutputStream dataOutputStream;
+    private static final String MESSAGECONSTANT = "This is additional Content.";
```
kahgoh, line 21: It is more idiomatic to use [screaming snake case](https://en.wiktionary.org/wiki/screaming_snake_case) for constants in Java. Could you please also move this to just under the `public class PaasioTest {` line? (its more idiomatic that way)

```suggestion
    private static final String MESSAGE_CONSTANT = "This is additional Content.";
```

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Test
+    @Disabled("Remove to run test")
+    void verifyBytesOfDataWrittenInTheOutputStream() throws IOException {
+
+        byte[] writeFileContent = MESSAGECONSTANT.getBytes();
+        paasioReaderWriter.write(writeFileContent);
+        paasioReaderWriter.write(writeFileContent);
+        paasioReaderWriter.write('s');
+
```
kahgoh, line 207: I think this is missing asserting on the contents that are written to the `dataOutputStream`.

exercises/practice/paasio/src/test/java/PaasioTest.java
```diff
@@ -0,0 +1,261 @@
...
+        assertThat(paasioReaderWriter.getBytesWritten()).isEqualTo((writeFileContent.length * 2L) + 1);
+
+    }
+
+    @Test
+    @Disabled("Remove to run test")
+    void verifyBytesWrittenFromOffsetIntoTheOutputStreamAlongWithTheCount() throws IOException {
+
+        String fileContentToBeWritten = "   This is additional Content.";
```
kahgoh, line 216: Similar to another comment, this isn't "file" specific, so suggest renaming the variable.

PR #2987, opened 2025-08-26
affine-cipher, all-your-base, allergies

AtharvOps:
# pull request

I noticed an inconsistency between the Canonical-data and the tests, and I've made the changes on line no. 134 and 143 in Practice exercise 'all-your-base'.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/acronym/src/test/java/AcronymTest.java
+++ b/exercises/practice/acronym/src/test/java/AcronymTest.java
@@ -2,11 +2,9 @@ import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
-
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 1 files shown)


exercises/practice/affine-cipher/src/test/java/AffineCipherTest.java
```diff
@@ -8,57 +9,65 @@ public class AffineCipherTest {
 
     private AffineCipher affineCipher = new AffineCipher();
 
+    @DisplayName("encode yes")
     @Test
     public void testEncodeYes() {
```
jagdish-15:
```suggestion
    @Test
    @DisplayName("encode yes")
    public void testEncodeYes() {
```

Hey @AtharvOps, could you please make sure to add all the `@DisplayName` tags just before each test method (below the other tags, as shown above) to keep it consistent with the rest of the codebase?

jagdish-15 (approved):

PR #3003, opened 2025-09-06
Add displayname annotations

bibekparajuli123:
# pull request

This PR adds @DisplayName annotations to the test methods for the following practice exercises:
pov
prime-factors
protein-translation
proverb
pythagorean-triplet
queen-attack
rail-fence-cipher
raindrops
rational-numbers
react

Descriptions were taken from the corresponding canonical-data.json files in the problem-specifications repo.

Reviewer Resources:

### DIFF
--- a/exercises/practice/queen-attack/src/test/java/QueenAttackCalculatorTest.java
+++ b/exercises/practice/queen-attack/src/test/java/QueenAttackCalculatorTest.java
@@ -1,4 +1,5 @@
 import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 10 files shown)


exercises/practice/queen-attack/src/test/java/QueenAttackCalculatorTest.java
```diff
@@ -109,6 +122,8 @@ public void testQueensCanAttackOnFourthDiagonal() {
 
     @Disabled("Remove to run test")
     @Test
+    @DisplayName("cannot attack if falling diagonals are only the same " +
+            "when reflected across the longest falling diagonal")
```
jagdish-15:
```suggestion
    @DisplayName(
        "cannot attack if falling diagonals are only the same " +
        "when reflected across the longest falling diagonal"
    )
```

Just this small change to keep it consistent with the earlier files!

PR #3002, opened 2025-09-05
Testing workflow

jagdish-15:
# Pull Request

Hey @kahgoh,  

Can you take a look at this and let me know your thoughts?  
It took about **1:25** to run when I tested it with 4 changed exercises.  

Currently, it only tests exercises where either `.java` or `.gradle` files have been modified.  

Let me know if you have any suggestions or nits!  

Closes #2693
---

Reviewer Resources:

[Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#event-checklist)

PR #3006, opened 2025-09-07
Run all tests if gradle file is changed

jagdish-15:
# pull request

Continues #3002 

---

Reviewer Resources:

### DIFF
--- a/concepts/chars/.meta/config.json
+++ b/concepts/chars/.meta/config.json
@@ -3,5 +3,7 @@
   "authors": [
     "ystromm"
   ],
-  "contributors": []
+  "contributors": [
+    "kahgoh"
+  ]
 }
--- a/concepts/chars/about.md
+++ b/concepts/chars/about.md
@@ -1,6 +1,108 @@
 # About
 
-`char`s are generally easy to use.
-They can be extracted from strings, added back (by means of a string builder), defined and initialised using literals with single quotes, as in `char ch = 'A';`, assigned and compared.
+The Java `char` primitive type is a 16 bit representation of a single Unicode character.
 
-The Character class encapsulates the char value.
+~~~~exercism/note
+The `char` type is based on the [original Unicode specification][unicode-specification], which used 16 bits to represent characters.
+This is enough to cover most of the common letters and covers characters in the range 0x0000 to 0xFFFF.
+The specification has since expanded the range of possible characters up to 0x01FFFF.
+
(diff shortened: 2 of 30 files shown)


bin/test-changed-exercise
```diff
@@ -13,9 +13,16 @@ MERGE_BASE=$(git merge-base HEAD origin/"$BASE_BRANCH")
 # Get changed files relative to merge base
 changed_files=$(git diff --name-only "$MERGE_BASE" HEAD)
 
+# 🔹 NEW: If any Gradle build file changed, run the full suite and exit
+if echo "$changed_files" | grep -qE '\.(gradle|gradlew|bat)$|settings\.gradle'; then
+  echo "Gradle build files changed, running full test suite..."
+  ./exercises/gradlew -p exercises test
```
kahgoh: Could we use the Docker to run the tests (as in the `test-with-test-runner` script)? Could we also use the test runner to test the only changed exercises?

It just occurred to me that the run all test job uses the `test-with-test-runner` script, which runs the tests in the test runner's Docker image. I think ideally that's what we want to do in our CI (use the Docker test image to run the tests) because that will highlight any potential issues when the changes run with the test runners (in both the "run everything" and "run over just the exercises that have changed" cases). 

An example of a change that is likely to be picked up by using the Docker test runner image and not the Gradle build is adding a new dependency. The test runners don't have internet access and can't download the dependency when they run. Instead, the test image needs to be updated with the dependency in the image.

bin/test-changed-exercise
```diff
@@ -13,9 +13,16 @@ MERGE_BASE=$(git merge-base HEAD origin/"$BASE_BRANCH")
 # Get changed files relative to merge base
 changed_files=$(git diff --name-only "$MERGE_BASE" HEAD)
 
+# 🔹 NEW: If any Gradle build file changed, run the full suite and exit
```
kahgoh: No need to mark what's new in the comment.

```suggestion
# If any Gradle build file changed, run the full suite and exit
```

kahgoh (approved):

bin/test-changed-exercise
kahgoh: Commit [ef1cfa9](https://github.com/exercism/java/pull/3006/commits/ef1cfa9a7db4a51417854ebbc488fd3ae3550d4e) runs tests using the Docker test runner when there are changes to the Gradle files.

PR #3008, opened 2025-09-08
[Draft] Add rate limiter exercise

### LINKED ISSUE #3007: Add rate limiter exercise
**Exercise idea**
Implement a fixed-window rate limiter that enforces a maximum number of actions per key within a time window.

**Concepts taught**
State management: tracking per-key counters and window start times.
Time handling: reasoning about elapsed time and window boundaries.
Boundary conditions: handling “at the exact moment” vs “just past” a window.

Link to discussion in forum: https://forum.exercism.org/t/new-exercise-proposal-rate-limit/19109/10

andreatanky:
# pull request

Closes #3007 

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -1900,6 +1900,14 @@
           "lists"
         ],
         "difficulty": 10
+      },
+      {
+        "slug": "rate-limiter",
+        "name": "rate-limiter",
+        "uuid": "b4b0c60e-4ce1-488e-948f-bcb6821c773c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
       }
     ],
     "foregone": [
--- a/exercises/practice/rate-limiter/.meta/config.json
+++ b/exercises/practice/rate-limiter/.meta/config.json
@@ -1,0 +1,15 @@
+{
+  "authors": [],
+  "files": {
+    "solution": [
+      "src/main/java/RateLimiter.java"
+    ],
+    "test": [
+      "src/test/java/RateLimiterTest.java"
+    ],
+    "example": [
+      ".meta/src/reference/java/RateLimiter.java"
+    ]
+  },
+  "blurb": ""
+}


exercises/practice/rate-limiter/src/main/java/RateLimiter.java
```diff
@@ -0,0 +1,3 @@
+public interface RateLimiter<K> {
```
kahgoh: Since there is only one `RateLimiter` to implement for this exercise, I'd prefer simplifying the exercise by making this _the_ class to implement instead of an interface (this means in a fully implemented solution, the code that's in `FixedWinowRateLimiter` would be here instead).

exercises/practice/rate-limiter/src/main/java/TimeSource.java
kahgoh, line 1: I have a similar comment for this interface as the exercise has just one type of `TimeSource`, I'd prefer making this a class instead and turning `TimeSource.Fake` into _the_ `TimeSource`.

exercises/practice/rate-limiter/src/main/java/TimeSource.java
```diff
@@ -0,0 +1,24 @@
+public interface TimeSource {
+    long nowNanos();
```
kahgoh: To allow practice with more of Java's time API, I'd suggest returning an [java.time.Instant](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Instant.html) instead.

```suggestion
    Instant now();
```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Test
+    void separateKeysHaveIndependentCountersAndWindows() {
+        TimeSource.Fake clock = new TimeSource.Fake(42L);
+        RateLimiter<String> limiter = new FixedWindowRateLimiter<>(1, 100L, clock);
+
+        assertThat(limiter.allow("A")).isTrue();
+        assertThat(limiter.allow("B")).isTrue(); // independent key
+        assertThat(limiter.allow("A")).isFalse();
```
kahgoh, line 86: I think calling on `A` twice before trying `B` makes the case more clear. 

```suggestion
        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isFalse();
        assertThat(limiter.allow("B")).isTrue(); // independent key

```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,88 @@
...
+        // Just before boundary: still same window
+        clock.advance(9_999L);
+        assertThat(limiter.allow("A")).isFalse();
+
+        // At exact boundary: new window
+        clock.advance(1L);
+        assertThat(limiter.allow("A")).isTrue();
+    }
+
+    @Test
```
kahgoh, line 58: The second and later test cases should be marked `@Disabled("Remove to run test")`, similar to our other tests, like [leap](https://github.com/exercism/java/blob/main/exercises/practice/leap/src/test/java/LeapTest.java).

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(limiter.allow("A")).isFalse();
+
+        // Just before boundary: still same window
+        clock.advance(9_999L);
+        assertThat(limiter.allow("A")).isFalse();
+
+        // At exact boundary: new window
+        clock.advance(1L);
+        assertThat(limiter.allow("A")).isTrue();
+    }
```
kahgoh: I think there is quite a bit going on in this test case. Could we split up this into 3 tests? I think splitting would allow students to "ease" into the exercise better than having it all in the first test.

- the first would be the assertions the from 12-15 (the simplest case)
- the second checks the new window at the boundary (lines 21-23) 
- the third could check advance just before the boundary (lines 17-19)

exercises/practice/rate-limiter/src/main/java/TimeSource.java
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        @Override
+        public long nowNanos() {
+            return now;
+        }
+
+        public void advance(long nanos) {
+            this.now += nanos;
+        }
```
kahgoh: The advance could perhaps take a `Duration` to further use Java's time API and allow other units (other than nano-seconds) to be used.

exercises/practice/rate-limiter/.meta/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "authors": [],
```
kahgoh: Add your Github handle (`andreatanky`) to the authors array.

exercises/practice/rate-limiter/.meta/config.json
```diff
@@ -0,0 +1,19 @@
...
+    "test": [
+      "src/test/java/RateLimiterTest.java"
+    ],
+    "example": [
+      ".meta/src/reference/java/RateLimiter.java",
+      ".meta/src/reference/java/TimeSource.java",
+      ".meta/src/reference/java/FixedWindowRateLimiter.java"
+    ]
+  },
+  "blurb": ""
```
kahgoh: Missing a blurb. We'll need to come up with something. I'd suggest looking at [Restful API](https://github.com/exercism/java/blob/main/exercises/practice/rest-api/.meta/config.json)'s blurb for an example or inspiration.

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,12 @@
...
+You will build a fixed‑window rate limiter: a small component that restricts how many
+operations each key may perform within a fixed period of time (a “window”).
+
+What you should implement
+
+- Implement a single method in `src/main/java/FixedWindowRateLimiter.java`:
+  - `public boolean allow(K key)` — decide whether to allow the operation for the key
+    and update the limiter’s internal state accordingly.
+
+Everything else (interfaces and constructor) is provided for you.
```
kahgoh: There isn't any need to tell students what to implement and what's provided (as per our other exercises) - this should be clear from the exercise's source files.

```suggestion
```

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,12 @@
+What this exercise is about
```
kahgoh: No need for this first sentence in the instructions.

```suggestion
```

exercises/practice/rate-limiter/src/main/java/FixedWindowRateLimiter.java
```diff
@@ -0,0 +1,32 @@
...
+        WindowState(long windowStartNanos, int usedCount) {
+            this.windowStartNanos = windowStartNanos;
+            this.usedCount = usedCount;
+        }
+    }
+
+    private final int limit;
+    private final long windowSizeNanos;
+    private final TimeSource timeSource;
+    private final Map<K, WindowState> states = new HashMap<>();
```
kahgoh: Generally, we wouldn't provide the students with the `WindowState` inner class and fields - that can be left up to the students as part of their solution as there may be different ways for them solving this.

```suggestion
```

exercises/practice/rate-limiter/src/main/java/FixedWindowRateLimiter.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public FixedWindowRateLimiter(int limit, long windowSizeNanos, TimeSource timeSource) {
+        this.limit = limit;
+        this.windowSizeNanos = windowSizeNanos;
+        this.timeSource = timeSource;
+    }
```
kahgoh: Similar, no need to implement the constructor for the students. We should let them fill the implementation. We just need a blank one to allow it to be compiled.

```suggestion
    public FixedWindowRateLimiter(int limit, long windowSizeNanos, TimeSource timeSource) {
    }
```

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,12 @@
...
+
+You will build a fixed‑window rate limiter: a small component that restricts how many
+operations each key may perform within a fixed period of time (a “window”).
```
kahgoh: - Not all students will be familiar with a rate limiter. We could expand on this and describe or provide examples.
- The instructions in other exercises generally uses the phrase "_Your task is to ...._" instead of "_You will_". 
 
For example (feel free to alter or tweak):

```suggestion
Imagine running a single server connected to a client.
The client makes a request to the server.
The server processes the request and sends back a response.
_But_, processing the request takes time.
If the client makes too many requests at one time, the server might get overwhelmed.

This is where a rate limiter comes in.
A rate limiter allows or rejects a request depending on how often the client has made a request.

Your task is to build a fixed window rate limiter.
There are types of rate limiters.
A fixed window rate limiter allows a certain number of requests within a time window per client.
```

I've used the word "client" here. If we use that term, I'd recommend using that term in place of the `key` parameter for consistency. Alternatively, we could try adjusting the explanation to use or introduce the term "key".

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
andreatanky: Thanks for the suggestion — agreed. Since there’s only one limiter, making it a single class will simplify the exercise and reduce indirection

exercises/practice/rate-limiter/src/main/java/FixedWindowRateLimiter.java
andreatanky: Thanks! I was wondering how much guidance to provide, but on second thought it’s better to avoid prescribing the data model

exercises/practice/rate-limiter/.docs/instructions.md
andreatanky: Thanks for the suggestion! I’ve incorporated a similar narrative and aligned the terminology to use "client" instead of "key"

exercises/practice/rate-limiter/.meta/config.json
andreatanky: I've updated it to "Practice stateful logic and time handling by implementing a fixed-window rate limiter". Let me know what you think!

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
```diff
@@ -0,0 +1,12 @@
+import java.time.Instant;
+
+public class RateLimiter<K> {
+
+    public RateLimiter(int limit, long windowSizeNanos, TimeSource timeSource) {
```
kahgoh: I think we could make the `windowSizeNanos` parameter a `Duration` too, otherwise students might be able to solve this without using much of the time API.

exercises/practice/rate-limiter/src/main/java/TimeSource.java
```diff
@@ -0,0 +1,22 @@
...
+        return now;
+    }
+
+    public void advance(Duration d) {
+        this.now = this.now.plus(d);
+    }
+
+    public void advanceNanos(long nanos) {
+        this.now = this.now.plusNanos(nanos);
+    }
```
kahgoh: I think you might have forgotten to update the tests to use the `advance(Duration d)` method.

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
```diff
@@ -0,0 +1,12 @@
+import java.time.Instant;
+
+public class RateLimiter<K> {
```
kahgoh, line 4: The tests seem to always be using a `RateLimiter<String>`. Do we need the generics here? Alternatively, we could make tests with some other type for `K`.

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertThat(limiter.allow("A")).isTrue();
+        assertThat(limiter.allow("A")).isFalse();
+
+        // Just before boundary: still same window
+        clock.advanceNanos(9_999L);
+        assertThat(limiter.allow("A")).isFalse();
+
+        // At exact boundary: new window
+        clock.advanceNanos(1L);
+        assertThat(limiter.allow("A")).isTrue();
```
kahgoh, line 54: I think this looks like 3 test cases in one. Could we split up this into 3 tests? I think splitting would allow students to "ease" into the exercise better than having it all in the first test.
 
* the first would be the assertions the from 14-17 (the simplest case)
* the second checks just before the boundary boundary (lines 19-21)
* the third could checks new window (lines 23-25)

For example:

```java
@Test
 void allowsUpToLimit() {
        TimeSource clock = new TimeSource(Instant.EPOCH);
        RateLimiter<String> limiter = new RateLimiter<>(3, 10_000L, clock);

        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isFalse();
}

@Test
void denyCloseToBoundary() {
        TimeSource clock = new TimeSource(Instant.EPOCH);
        RateLimiter<String> limiter = new RateLimiter<>(2, 10_000L, clock);

        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isFalse();

        clock.advanceNanos(9_999L);
        assertThat(limiter.allow("A")).isTrue();
}

@Test
void allowsNewBoundary() {
        TimeSource clock = new TimeSource(Instant.EPOCH);
        RateLimiter<String> limiter = new RateLimiter<>(2, 10_000L, clock);

        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isTrue();
        assertThat(limiter.allow("A")).isFalse();

        clock.advanceNanos(10_000L);
        assertThat(limiter.allow("A")).isTrue();
}
```

Exercism uses a [TDD approach](https://exercism.org/docs/tracks/python/test-driven-development). The idea is that students focus on getting one test to pass at a time. I think this separation will give students a more gradual easing into the exercise.

exercises/practice/rate-limiter/.docs/instructions.md
kahgoh, line 1: We prefer the markdown to be formatted [one sentence per line](https://exercism.org/docs/building/markdown/markdown#h-one-sentence-per-line). For example:

```
Imagine a single server connected to one or more clients. 
A client sends a request, the server does some work, and returns a response.
But processing takes time. 
If a client sends too many requests too quickly, the server can become overwhelmed - everything slows down or fails.
```

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
andreatanky: I think it is useful to include Generics because different callers may key by String, Integer, UUID, or a custom type.

I have added in other tests for UUID, long, int, and custom client objects as keys

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
andreatanky: Thanks for the suggestion — I split the initial test into three to better match the TDD flow.

I kept the “just before boundary” assertion as false (line 34) because we’re still within the same fixed window; at 9,999 ns the next call should still be denied

exercises/practice/rate-limiter/.docs/instructions.md
andreatanky: Thanks for pointing this out, updated the instructions to one sentence per line

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
```diff
@@ -0,0 +1,13 @@
+import java.time.Duration;
+import java.time.Instant;
+
+public class RateLimiter<K> {
+
+    public RateLimiter(int limit, Duration windowSize, TimeSource timeSource) {
+
+    }
+
+    public boolean allow(K key) {
```
kahgoh: In the instructions we've used the word `client` rather than `key`, so this might be a bit confusing. What do you think about something like `clientId` for the parameter name?

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+Different strategies exist; in this exercise you’ll implement a fixed‑window rate limiter.
+
+Fixed‑window rate limiting groups time into equal‑length windows (for example, every 10 seconds) and allows up to a certain number of requests within each window for each client.
+Once the window resets, the allowance refreshes for the next window.
+Each client is tracked separately.
+
+Examples:
+- Limit: 3 requests per 10 seconds per client.
+  - A client’s first three requests in the same window are allowed; a fourth is rejected.
+  - After the window resets, the next request is allowed again and the counting starts fresh.
```
kahgoh: I think this example could be better expanded. For example, what about other clients? When does the window "start"?

```suggestion
For example, consider a rate limiter configured to limit 2 requests per 10 seconds per client.
Lets say client A sends a request. 
- Being its first request, the request is permitted.
- A second request within 10 seconds after the first one is also permitted.
- However, further requests after that would be denied _until_ at least 10 seconds has elapsed since the first request.
- If a second client sends its first request within 10 seconds with of the first client's first request, it would also be permitted, _regardless_ of whether the first client has sent a second request. 
```

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,18 @@
...
+A client sends a request, the server does some work, and returns a response.
+But processing takes time.
+If a client sends too many requests too quickly, the server can become overwhelmed — everything slows down or fails.
+
+A rate limiter is a small component that decides whether to allow or reject a request based on how frequently that client has been making requests.
+Different strategies exist; in this exercise you’ll implement a fixed‑window rate limiter.
+
+Fixed‑window rate limiting groups time into equal‑length windows (for example, every 10 seconds) and allows up to a certain number of requests within each window for each client.
+Once the window resets, the allowance refreshes for the next window.
+Each client is tracked separately.
```
kahgoh:
```suggestion
Each client is tracked separately, so another client can make requests within that same period.
```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,210 @@
...
+            @Override
+            public boolean equals(Object o) {
+                return (o instanceof ClientId) && ((ClientId) o).id.equals(this.id);
+            }
+
+            @Override
+            public int hashCode() {
+                return id.hashCode();
+            }
+        }
```
kahgoh: To be honest, I'm not a fan of having to rely on the correct implementation of the `equals` and `hashCode` in the test and I think having the `UUID`, `Integer`, `Long` and `String` variants is more than enough to test the generics side - perhaps we could remove this test case?

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,210 @@
...
+        assertThat(limiter.allow("B")).isFalse();
+
+        clock.advance(Duration.ofNanos(100L)); // new window for both at boundary
+        assertThat(limiter.allow("A")).isTrue();
+        assertThat(limiter.allow("B")).isTrue();
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    void longGapsResetWindowDeterministically() {
```
kahgoh:
```suggestion
    void longGapsResetWindow() {
```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,210 @@
...
+        assertThat(limiter.allow("key")).isFalse();
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    void separateKeysHaveIndependentCountersAndWindows() {
+        TimeSource clock = new TimeSource(Instant.EPOCH.plusNanos(42L));
+        RateLimiter<String> limiter = new RateLimiter<>(1, Duration.ofNanos(100L), clock);
+
+        assertThat(limiter.allow("A")).isTrue();
```
kahgoh, line 83: Most of the tests don't seem to advance the clock after the first call to `allow` (even if its just a tiny little bit). Is it worth considering doing it here?

config.json
```diff
@@ -1901,6 +1901,14 @@
...
         ],
         "difficulty": 10
+      },
+      {
+        "slug": "rate-limiter",
+        "name": "rate-limiter",
+        "uuid": "b4b0c60e-4ce1-488e-948f-bcb6821c773c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
```
kahgoh: For the difficulty, would you say it is similar in difficulty as Luhn? If so, we could say it is a 4 (a 4 is one of the lower medium difficulty, 3 is easy). I think 1 is too low.

config.json
```diff
@@ -1901,6 +1901,14 @@
           "lists"
         ],
         "difficulty": 10
+      },
+      {
+        "slug": "rate-limiter",
+        "name": "rate-limiter",
+        "uuid": "b4b0c60e-4ce1-488e-948f-bcb6821c773c",
+        "practices": [],
+        "prerequisites": [],
```
kahgoh: We should put `generic-types` as the prerequisite. Don't worry about `duration` - the concepts here are generally ones that we have added to the syllabus. We haven't added one for Java's time API and duration yet.

exercises/practice/rate-limiter/.meta/config.json
```diff
@@ -0,0 +1,17 @@
...
+      "src/main/java/TimeSource.java"
+    ],
+    "test": [
+      "src/test/java/RateLimiterTest.java"
+    ],
+    "example": [
+      ".meta/src/reference/java/RateLimiter.java",
+      ".meta/src/reference/java/TimeSource.java"
+    ]
+  },
```
kahgoh, line 19: I assum the student shouldn't change the `TimeSource` code? If so, it should be in the `editor` key, like this:

```suggestion
  "files": {
    "solution": [
      "src/main/java/RateLimiter.java"
    ],
    "test": [
      "src/test/java/RateLimiterTest.java"
    ],
    "example": [
      ".meta/src/reference/java/RateLimiter.java",
      ".meta/src/reference/java/TimeSource.java"
    ],
    "editor": [
      "src/main/java/TimeSource.java"
    ]
  },
```

exercises/practice/rate-limiter/src/main/java/TimeSource.java
```diff
@@ -0,0 +1,18 @@
+import java.time.Duration;
+import java.time.Instant;
+
```
kahgoh, line 3: Students might choose to download this exercise to work on their machines. We should add a Javadoc comment to tell them not to edit this file.

```suggestion

/**
 * NOTE: There is no need to change this file and is treated as read only by the Exercism test runners.
 */
```

kahgoh (changes requested):

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
kahgoh, line 1: The tests seem to all be using nano-seconds. Since we're using `Duration`, I'd suggest having a bit more of a mix and making use of milliseconds and seconds somewhere. 

I don't think there we need to add a new test case to use these other units - just updating one of the existing test cases to use them would be enough. Perhaps we could even express the window in one of the units and advance the clock in one of the "lower" units (e.g. the window could be x seconds and clock advance y milliseconds or nanoseconds).

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
andreatanky: I think this makes sense! Updated

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
andreatanky: Thanks for the feedback — could you share a bit more on what feels uncomfortable about the custom‑object test?
Is it that it couples outcomes to equals/hashCode correctness and can cause confusing failures unrelated to the rate‑limiter logic?

I have removed this custom object test case to reduce possible confusion

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
andreatanky: I think advancing the clock makes intent clearer and avoids identical-timestamp edge cases.
I added a tiny tick (clock.advance(Duration.ofNanos(1))) after the first allow in some of the tests for consistency

config.json
andreatanky: I think low‑medium (4) is reasonable! it only requires per‑key state and fixed‑window boundary handling with java.time, but avoids concurrency and advanced algorithms like sliding windows or token buckets.

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
andreatanky: In the updated test (UUID), the window is expressed in seconds and the advances use lower units:

Window: Duration.ofSeconds(1L) at RateLimiterTest.java:131
Lower-unit advances: Duration.ofMillis(1L) at RateLimiterTest.java:139 and boundary Duration.ofSeconds(1L) at RateLimiterTest.java:144

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
kahgoh: Yeah, it is that we'd be relying on the `equals` and `hashCode` being implemented correctly for this test to work properly. If this had been for an actual project, I'd be wondering if we need tests to prove that they have been implemented correctly and honor the `equals` and `hashCode` contract too. I think an alternative would be to use records which would take care of this for you, but I also felt it was already sufficient to use UUID objects.

config.json
kahgoh: Great! The exercises are sorted by difficulty and then name though. Could you please move this entry to between `proverb` and `rotational-cipher` in the exercises list?

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,20 @@
+Your task is to build a fixed‑window rate limiter.
```
kahgoh, line 3: The [Markdown linter](https://github.com/exercism/java/actions/runs/17914115798/job/50935078328?pr=3008#step:3:11) requires a heading. 

```suggestion
# Instructions

Your task is to build a fixed‑window rate limiter.
```

exercises/practice/rate-limiter/.docs/instructions.md
```diff
@@ -0,0 +1,20 @@
...
+Fixed‑window rate limiting groups time into equal‑length windows (for example, every 10 seconds) and allows up to a certain number of requests within each window for each client.
+Once the window resets, the allowance refreshes for the next window.
+Each client is tracked separately, so another client can make requests within that same period.
+
+For example, consider a rate limiter configured to limit 2 requests per 10 seconds per client.
+Lets say client A sends a request. 
+- Being its first request, the request is permitted.
+- A second request within 10 seconds after the first one is also permitted.
+- However, further requests after that would be denied _until_ at least 10 seconds has elapsed since the first request.
+- If a second client sends its first request within 10 seconds with of the first client's first request, it would also be permitted, _regardless_ of whether the first client has sent a second request.
```
kahgoh: Minor wording / grammar corrections and add blank line for [linter](https://github.com/exercism/java/actions/runs/17914115798/job/50935078328?pr=3008#step:3:11)

```suggestion
Lets say a client sends a request: 

- Being its first request, the request is permitted.
- A second request within 10 seconds after the first one is also permitted.
- However, further requests after that would be denied _until_ at least 10 seconds has elapsed since the first request.
- If a second client sends its first request within 10 seconds with of the first client's first request, it would also be permitted, _regardless_ of whether the first client has sent a second request. 
```

exercises/practice/rate-limiter/.meta/config.json
kahgoh, line 1: The CI is [failing](https://github.com/exercism/java/actions/runs/17914115884/job/50935078638?pr=3008) due to formatting error in this file. To fix it:

- If you don't already have configlet, run `bin/fetch-configlet` (this a Linux script is in this [repository](https://github.com/exercism/java/blob/main/bin/fetch-configlet)). Alternatively, you can also download [configlet](https://github.com/exercism/configlet/releases).
- Run ` bin/configlet fmt -e rate-limiter -u`.

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,185 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class RateLimiterTest {
+
+    @Test
+    void allowsUpToLimit() {
```
kahgoh, line 14: We've started putting `@DisplayName` tags in the tests in the other PRs (example in suggestion). Could you please do the same?

```suggestion
    @Test
    @DisplayName("Allows up to window limit")
    void allowsUpToLimit() {
```

config.json
andreatanky: Shifted! Thanks!

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
andreatanky: Updated with the tests with `@DisplayName`

exercises/practice/rate-limiter/.meta/config.json
andreatanky: Formatted `config.json` file

exercises/practice/rate-limiter/.docs/instructions.md
kahgoh, line 1: Markdown linter is [failing](https://github.com/exercism/java/actions/runs/17932687155/job/50994733697?pr=3008) in this file due to trailing spaces on lines 18 and 23.

config.json
kahgoh, line 1: The CI is [failing](https://github.com/exercism/java/actions/runs/17932687233/job/50994733976?pr=3008) on this `config.json`. I think you should be able to run `bin/configlet fmt -u` to fix it.

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("Supports UUID keys with mixed time units")
+    void supportsUuidKeys() {
+        TimeSource clock = new TimeSource(Instant.EPOCH);
+        // Use a seconds-long window and advance in smaller units to mix Duration usage
```
kahgoh: I think it is clear enough from the code that this comment isn't needed.

```suggestion
```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,196 @@
...
+            clock.advance(Duration.ofNanos(10L));
+            assertThat(limiter.allow("k")).isTrue();
+            assertThat(limiter.allow("k")).isFalse();
+        }
+    }
+
+
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("Supports UUID keys with mixed time units")
```
kahgoh: I think being able to use UUIDs is the main thing in this test, so we could probably remove the mixed time units bit from the test description.

```suggestion
    @DisplayName("Supports UUID keys")
```

exercises/practice/rate-limiter/src/test/java/RateLimiterTest.java
```diff
@@ -0,0 +1,196 @@
...
+class RateLimiterTest {
+
+    @Test
+    @DisplayName("Allows up to window limit")
+    void allowsUpToLimit() {
+        TimeSource clock = new TimeSource(Instant.EPOCH);
+        RateLimiter<String> limiter = new RateLimiter<>(3, Duration.ofNanos(10_000L), clock);
+
+        assertThat(limiter.allow("A")).isTrue();
+        // Advance minimally to model time passage between requests
```
kahgoh: I think it is clear enough that the comment isn't needed here.

```suggestion
```

exercises/practice/rate-limiter/src/main/java/RateLimiter.java
```diff
@@ -0,0 +1,13 @@
+import java.time.Duration;
+import java.time.Instant;
+
+public class RateLimiter<K> {
+
+    public RateLimiter(int limit, Duration windowSize, TimeSource timeSource) {
+
```
kahgoh: In some the other exercises, we also do this for the constructor ([rotational cipher](https://github.com/exercism/java/blob/main/exercises/practice/rotational-cipher/src/main/java/RotationalCipher.java), for example).

```suggestion
        throw new UnsupportedOperationException("Delete this statement and write your own implementation.");
```

andreatanky:
Appreciate the thorough review!

config.json
```diff
@@ -810,6 +810,16 @@
         ],
         "difficulty": 4
       },
+      {
+        "slug": "rate-limiter",
+        "name": "rate-limiter",
```
kahgoh: Name shouldn't be kebab case.

```suggestion
        "name": "Rate Limiter",
```

PR #3020, opened 2025-09-22
fix(FootBallMatchReportsTest.java): remove unused import

AlvesJorge:
# pull request

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/football-match-reports/.docs/instructions.md
+++ b/exercises/concept/football-match-reports/.docs/instructions.md
@@ -27,11 +27,11 @@ FootballMatchReports.onField(10);
 // => "striker"
 ```
 
-## 2. Raise an alert if an unknown shirt number is encountered
+## 2. Output "unknown" if the shirt number is not part of the official list
 
-Modify the `FootballMatchReports.onField()` method to throw an `IllegalArgumentException` when a shirt number outside the range 1-11 is processed.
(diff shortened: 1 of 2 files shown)


kahgoh:
👆🏻 that and to properly mark them as no longer passing requires re-testing them all, which can be a rather expensive operation (more info in the [docs](https://exercism.org/docs/building/tracks/ci/workflows/no-important-files-changed) if you're interested).

AlvesJorge:
Yeah I understand the not wanting to check if all submissions pass the latest version. It's just that they explicitly say _"This solution correctly solves the latest version of the exercise"_. Which is awfully specific, and very wrong. If i was a newbie and I copy pasted the solution code I'd be so confused as to why it's not passing.

jagdish-15:
Are you sure it shows that? As far as I know, and even after checking, it doesn’t seem to. Could you double-check on your side?  
@kahgoh do you happen to know more about this?

kahgoh:
Yeah ... I see it ... the tooltip does say that when hovering over the check mark. I can see how it can be confusing and using `[no important files changed]` would have prevented it from re-running the check. 

On another note ... I wonder if the test display name "_The onField method returns 'invalid' for invalid shirt number_" should describe why it is invalid (e.g. "_The onField method returns 'invalid' for shirt positive number outside range_"). It would be more consistent with the one below it which says "_negative number_". If so, we could make the change and re-run the check at that point (@AlvesJorge, if agreed, would you want to raise the PR?), otherwise I'll see if someone can trigger the re-testing.

PR #3026, opened 2025-09-27
Add deprecated workflow

### LINKED ISSUE #2690: Add GitHub Actions workflow that fails if any deprecated exercise files change
Some exercises on this track have been deprecated and should no longer be updated.

To make sure this doesn't happen accidentally anyway, we could add a workflow to GitHub actions that checks if any of the files changed in a PR belong to a deprecated exercise and if so, fail. This action should only run on PRs.

For more information on the deprecation of exercises, see https://exercism.org/docs/building/tracks/deprecated-exercises.

jagdish-15:
# Pull Request

closes #2690 

---

Reviewer Resources:


jagdish-15:
@kahgoh  
Do you think there’s a way we can load the deprecated exercises dynamically, so we don’t have to manually update the list in the script every time an exercise gets deprecated? Or is that not really necessary?

kahgoh:
We could look in `config.json` (check if they have the status `deprecated`). Might be able to use [jq](https://jqlang.org/) to process JSON in the shell, if it is required.

.github/workflows/check-deprecated-exercises.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  test-deprecated:
+    name: Check for deprecated exercises
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+        with:
+          fetch-depth: 0
+      - name: Install jq
+        run: sudo apt-get update && sudo apt-get install -y jq
```
kahgoh: By any chance, have you tested without this step? I'm just wondering whether jq might happen to be available without having to install.

jagdish-15: I had assumed we’d need to install jq, but after testing in my fork I confirmed it’s already preinstalled, so we can safely remove this step.

kahgoh: Thanks for testing and confirming!

kahgoh (approved):

PR #3039, opened 2025-10-22
Add @DisplayName to pig latin

gregorioRQ:
# pull request

This is my first PR ever, so I'd really appreciate some feedback.
In the pig-latin exercise tests, I added the "@DisplayName" annotation to each test.

---

Reviewer Resources:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "java.debug.settings.onBuildFailureProceed": true
-}


exercises/practice/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -14,138 +15,139 @@ public void setup() {
     }
 
     @Test
+    @DisplayName("ay is added to words that begin with a vowel")
     public void testWordBeginningWithA() {
         assertThat(pigLatinTranslator.translate("apple")).isEqualTo("appleay");
     }
 
-    @Disabled("Remove to run test")
```
kahgoh, line 21: There is no need to remove the `@Disabled` annotations. Please put them back.

exercises/practice/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -14,138 +15,139 @@ public void setup() {
     }
 
     @Test
+    @DisplayName("ay is added to words that begin with a vowel")
```
kahgoh: The text for the `@DisplayName` should come from the `description` in the exercise's [canonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/pig-latin/canonical-data.json). For example, this particular test case, it the description is on [line 9](https://github.com/exercism/problem-specifications/blob/main/exercises/pig-latin/canonical-data.json#L9), so the text should be `word beginning with a`. The other `@DisplayName`s should be updated to match theirs too.

exercises/practice/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -14,138 +15,162 @@ public void setup() {
...
+    @Disabled
+    @DisplayName("word beginning with e")
     public void testWordBeginningWithE() {
         assertThat(pigLatinTranslator.translate("ear")).isEqualTo("earay");
     }
 
-    @Disabled("Remove to run test")
     @Test
+    @Disabled
+    @DisplayName("word beginning with i")
```
jagdish-15, line 32:
```suggestion
    @Disabled("Remove to run test")
    @Test
    @DisplayName("word beginning with i")
```


Avoid making unnecessary changes, such as altering the order between @Test and @Disabled.
Also, make changes across all instances according to the above suggestion.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.debug.settings.onBuildFailureProceed": true
+}
```
jagdish-15: It seems this file was unintentionally added by your IDE.
Please remove it!

exercises/practice/pig-latin/src/test/java/PigLatinTranslatorTest.java
```diff
@@ -14,138 +15,162 @@ public void setup() {
     }
 
     @Test
+    @Disabled
```
kahgoh: Sorry I should've mentioned earlier, the only exception is the first test - the first test doesn't get `@Disabled`. The idea is that Exercism uses TDD where students would first try to get the first test working first. Once they have done that, they can  remove the tag from the second test and get that working and so on.

```suggestion
```

exercises/practice/pig-latin/src/test/java/PigLatinTranslatorTest.java
kahgoh: If it helps, the overall diff in the _Files changed_ tab should show just the addition of the `@DisplayName` annotations. There shouldn't be any changes to `@Disabled`.

PR #2983, opened 2025-08-17
feat: add streams concept

navaneedan07:
# pull request

Closes #2982 

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/java/blob/main/POLICIES.md#event-checklist)

PR #3041, opened 2025-10-31
Add new Streams concept with intro and resources

ZaildarAnmol:
# pull request

---

Reviewer Resources:

### DIFF
--- a/concepts/streams/about.md
+++ b/concepts/streams/about.md
@@ -1,0 +1,13 @@
+# About Streams
+
+Java Streams provide a functional, declarative approach to processing collections of data.
+
+Instead of writing explicit loops, Streams let you build **pipelines** of operations — transforming data step-by-step in a clean, readable way.
+
+---
+
+### Creating Streams
+Streams can be created from collections, arrays, or I/O channels:
+```java
+List<Integer> numbers = List.of(1, 2, 3, 4, 5);
+Stream<Integer> stream = numbers.stream();
--- a/concepts/streams/introduction.md
+++ b/concepts/streams/introduction.md
@@ -1,0 +1,22 @@
+# Introduction
+
+The `Stream` API, introduced in Java 8, provides a modern, functional way to process data.
+
+A **Stream** is a sequence of elements that supports operations like filtering, mapping, and reducing.  
+It allows you to transform and analyze collections without using traditional loops.
+
+Example:
+```java
+import java.util.*;
+import java.util.stream.*;
+
+public class Example {
+    public static void main(String[] args) {
+        List<String> names = List.of("Arjun", "Riya", "Sam", "Aman");
+
+        names.stream()
+             .filter(n -> n.startsWith("A"))
+             .map(String::toUpperCase)
+             .forEach(System.out::println);
+    }
+}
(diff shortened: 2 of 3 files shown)


kahgoh (changes requested):
Thanks for wanting to contribute the Streams concept, but there is already an open PR for this (#2983) so I'm going to put this one on hold by putting this in "requested changes" state. Admittedly, the other one hasn't had activity for a while, so I've pinged the other PR's author on that one. If there's no reply or if they've lost interest, I'll come back to review this one.

concepts/streams/about.md
```diff
@@ -0,0 +1,13 @@
+# About Streams
+
+Java Streams provide a functional, declarative approach to processing collections of data.
+
+Instead of writing explicit loops, Streams let you build **pipelines** of operations — transforming data step-by-step in a clean, readable way.
+
```
xinri, line 6: I would also introduce the intermediate and terminal methods to introduce the possiblity of chaining function.
+ I would also introduce the usage of the lambda with only one function to execute which makes more readable

concepts/streams/introduction.md
```diff
@@ -0,0 +1,22 @@
...
+```java
+import java.util.*;
+import java.util.stream.*;
+
+public class Example {
+    public static void main(String[] args) {
+        List<String> names = List.of("Arjun", "Riya", "Sam", "Aman");
+
+        names.stream()
+             .filter(n -> n.startsWith("A"))
```
xinri, line 18: I would introduce, as an intermediate operation:
- filter
- map, mapToInt, mapToLong...
- peek
- limit
- distinct
- sorted
- skip

And Termination operations:
- collect
- count
- reduce
- anyMatch 
- findFirst

concepts/streams/about.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+### Creating Streams
+Streams can be created from collections, arrays, or I/O channels:
+```java
+List<Integer> numbers = List.of(1, 2, 3, 4, 5);
+Stream<Integer> stream = numbers.stream();
```
kahgoh, line 13: This content is incomplete

concepts/streams/introduction.md
kahgoh: I agree it would be good to introduce some of theses operations. Note we don't need a complete list, but having some would be great.

concepts/streams/introduction.md
```diff
@@ -0,0 +1,22 @@
+# Introduction
+
+The `Stream` API, introduced in Java 8, provides a modern, functional way to process data.
+
+A **Stream** is a sequence of elements that supports operations like filtering, mapping, and reducing.  
+It allows you to transform and analyze collections without using traditional loops.
+
```
kahgoh, line 7: I think this is also missing how to get a `Stream`

concepts/streams/introduction.md
xinri: agree. I would think that at least the collect and findfirst or anyMatch should be introduced:
- Collect is one of the terminal operation that is widely used.
- a terminal operation liike findfirst to show we can return only one element and anyMatch to show that a stream can return a boolean

PR #3046, opened 2025-11-09
add the design.md for the calculator-conundrum

xinri:
# pull request

Add the design.md file for the calculator-conundrum for the analyzer.

Fixes: https://github.com/exercism/java/issues/2673

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/calculator-conundrum/.meta/design.md
+++ b/exercises/concept/calculator-conundrum/.meta/design.md
@@ -18,3 +18,15 @@
 
 - `conditionals-if`: know how to do conditional logic.
 - `switch-statement`: know how to work with a `switch` statement.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution is using the try catch statement for the division.
+- `actionable`: If the solution is wrapping all the code in a try catch statement, instruct the student to only use the `try catch` for the division statement
+- `actionable`: If the solution uses only `if` statement, instruct the student that he/she can use the `switch case` statement to perform the operations.
+- `informative`: If the solution does not throw the exception for `Operation cannot be null` and `Operation cannot be empty` at the beginning, suggest the fail-fast approach to the student.
+  Inform this way of implementation can be less error-prone and more readable as suggested by Martin Fowler: <https://martinfowler.com/ieeeSoftware/failFast.pdf>
+
+[analyzer]: https://github.com/exercism/java-analyzer


kahgoh (commented):
Thanks for wanting to contribute @xinri! My thoughts on this one:
 
> - `essential`: Verify that the solution is not wrapping all the code in a try catch statement

I agree that it is worth recommending to the student not to wrap all the code in a try catch statement.  However, I think this should be __actionable__ because the objective is only to be able to use `try catch` (which they would've demonstrated).  To me, this comment is  an improvement. I think a more `essential` comment would be for something that doesn't meet the exercise's objective (for example, somehow solving this exercise without  a `try-catch`.

>  - `actionable`: If the solution uses an `if` statement, instruct the student to use the `switch case` statement instead.

My understanding of this one is that we would also tell to use a `switch` statement if they use both an `if` and `switch`, like this one:

```java
    public String calculate(int operand1, int operand2, String operation) {
        if (operation == null) {
            throw new IllegalArgumentException("Operation cannot be null");
        }

        if (operation.isEmpty()) {
            throw new IllegalArgumentException("Operation cannot be empty");
        }

        int result;
        switch (operation) {
            case "+" -> result = operand1 + operand2;
            case "*" -> result = operand1 * operand2;
            case "/" -> {
                try {
                    result = operand1 / operand2;
                } catch (ArithmeticException e) {
                    throw new IllegalOperationException("Division by zero is not allowed", e);
                }
            }
            default -> throw new IllegalOperationException("Operation '" + operation + "' does not exist");
        }
```

Did you mean something more like "If the solution does not use a `switch` statement, instruct the student to use a `switch` statement"? 

>  - `informative`: If the solution does not use `HashMap`, instruct the student to use the operator as a key and a BiFunction implementing the calculation
>  Explain that it is more efficient for performance, and it removes the cyclomatic complexity

I don't understand how or why a `HashMap` is needed? I thought the students would use  `switch` as in earlier example.  I'd expect a `switch` would be more efficient because Java can turn it into a [table](https://medium.com/@AlexanderObregon/how-javas-switch-statement-works-internally-326ba8485f31), but `HashMap` requires [hashing operations](https://medium.com/@akashnadar14/understanding-java-hashmap-performance-why-its-not-always-o-1-cab51a0b358d) and need to account for keys that hash to the same value. Using a  `HashMap` is also _not_  an objective of this exercise either.

> - `informative`: If the solution does not throw the exception for `Operation cannot be null` and `Operation cannot be empty` at the beginning, instruct the student to do so
>  Explain that it is better to fail fast, and it will make the code less error-prone and more readable

I generally agree with this one. The only part I'm not sure about is the readability as that can be a bit subjective.

exercises/concept/calculator-conundrum/.meta/design.md
```diff
@@ -18,3 +18,18 @@
...
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution is using the try catch statement for the division
+- `essential`: Verify that the solution is throwing an exception if the parameter "operation" is null
+- `essential`: Verify that the solution is throwing an exception if the parameter "operation" is empty
+- `essential`: Verify that the solution is throwing an exception if the parameter "operation" is not a mathematical operation.
+- `actionable`: If the solution is wrapping all the code in a try catch statement, instruct the student to only use the try catch for the division statement
+- `actionable`: If the solution uses only `if` statement, instruct the student that he/she can use the `switch case` statement also.
```
kahgoh: I think it might be worth mentioning where the `switch case` could be used. For example "... can use the `switch case` statement to perform the operation".

kahgoh (changes requested):
Ah, oops, there is lint failure CI:

> Error: exercises/concept/calculator-conundrum/.meta/design.md:30:110 MD034/no-bare-urls Bare URL used [Context: "https://martinfowler.com/ieeeS..."] https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md034.md
Error: Failed with exit code: 1

(see also: https://github.com/exercism/java/actions/runs/19278839635/job/55183291934?pr=3046#step:3:11)

kahgoh (approved):

Issue #2676, opened 2024-01-30
elons-toy-car


sanderploegsma:

bishabadkri:
## Analyzer

This section outlines how the analyzer should review submissions for the `satellite` concept exercise and what types of feedback it should provide. The aim is to guide students constructively, reinforcing correct approaches and suggesting improvements where necessary.

The feedback should be **targeted**, **respectful**, and **encouraging**, helping learners better understand vector math, coordinate manipulation, and algorithmic design in the context of satellite data interpretation.

---

### 1. ✅ Code Structure and Clarity
**What to check**: Ensure that the solution follows clean coding practices—such as using descriptive variable and function names, breaking logic into reusable components, and avoiding code duplication.

**Why**: Clear structure improves readability and maintainability, which are crucial in real-world software engineering.

**Feedback type**:  
- *Informative*: When the code is functional but could be cleaner.  
- *Actionable*: When structure significantly hinders readability.

**Example comment**:  
_"Your solution works well! To make it even more readable, consider splitting the coordinate transformation into a separate helper function."_

---

### 2. 🛰️ Correctness of Coordinate Processing
**What to check**: Validate whether the student is correctly applying vector math, triangulation logic, or geometry operations to determine the final position based on satellite data.

**Why**: The core learning goal is to understand how to process and triangulate satellite coordinates accurately.

**Feedback type**:  
- *Essential*: If the output is incorrect due to faulty logic.  
- *Actionable*: If the logic is overly complex or fragile.

**Example comment**:  
_"It looks like the coordinates are being summed directly rather than triangulated. You may want to review how trilateration works with distances from multiple points."_

---

### 3. 🧮 Use of Mathematical Abstractions
**What to check**: Encourage the use of vector operations (like dot product, cross product, distance formulas) instead of manually coded logic that’s harder to generalize.

**Why**: Reinforces abstraction and the use of standard mathematical tools rather than reinventing them manually.

**Feedback type**:  
- *Actionable*

**Example comment**:  
_"Great effort! To simplify your solution and reduce potential errors, you might consider using built-in vector operations or libraries."_

---

### 4. 🧠 Algorithmic Efficiency
**What to check**: Examine if the solution unnecessarily loops through the same data or recalculates results multiple times.

**Why**: Promotes optimization and efficient problem-solving—a key concept in all technical interviews and real applications.

**Feedback type**:  
- *Actionable*

**Example comment**:  
[...]

PR #3047, opened 2025-11-11
Add analyzer design for jedliks toy car

xinri:
# pull request

Add the design.md file for the jedliks-toy-car for the analyzer.

Fixes: https://github.com/exercism/java/issues/2676

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/jedliks-toy-car/.meta/design.md
+++ b/exercises/concept/jedliks-toy-car/.meta/design.md
@@ -28,3 +28,15 @@
 - `strings`: know how to do basic string interpolation.
 - `numbers`: know how to compare numbers.
 - `conditionals`: know how to do conditional logic.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution keeps the void type for the drive function.
+- `essential`: Verify that the solution has fields in the class
+- `actionable`: If the solution defines the fields as `public`, instruct the student to use `private` and explain the encapsulation principle.
+- `informative`: If the solution does not use a primitive as a type for the fields, inform the student to use it.
+   Explain that the values cannot be null and it is less error-prone
+
+[analyzer]: https://github.com/exercism/java-analyzer


exercises/concept/jedliks-toy-car/.meta/design.md
```diff
@@ -28,3 +28,19 @@
...
 - `conditionals`: know how to do conditional logic.
+
+## Analyzer
+
+This exercise could benefit from the following rules in the [analyzer]:
+
+- `essential`: Verify that the solution keeps the void type for the drive function.
+- `essential`: Verify that the solution has the fields in the class for battery percentage and the driving distance.
+- `essential`: Verify that the solution updates the variables in the drive method.
+- `essential`: Verify that the solution is not updating values when the battery is empty.
```
kahgoh: > - `essential`: Verify that the solution has the fields in the class for battery percentage and the driving distance.


I don't think we should be insisting that there are specific fields for battery percentage and driving distance. To be honest, I'm not sure how reliably an analyzer implementation can tell which fields is for battery percentage and driving distance. Instead, I'd suggest simply checking that the class has fields.

> - `essential`: Verify that the solution updates the variables in the drive method.

Did you mean "updates the _fields_ in the class"?

> `essential`: Verify that the solution is not updating values when the battery is empty.

I don't think we need this one - I think the last two tests cover this.

exercises/concept/jedliks-toy-car/.meta/design.md
```diff
@@ -28,3 +28,19 @@
...
+- `actionable`: If the solution defines the fields as `public`, instruct the student to use `private` and explain the encapsulation principle.
+- `informative`: If the solution does not use a primitive as a type for the fields, inform the student to use it.
+   Explain that the values cannot be null and it is less error-prone
+- `informative`: If the solution updates variables outside the drive function, instruct the student to move there.
```
kahgoh: I didn't quite understand what you mean by "move there". Is it checking that the fields are only changed in the `drive` method?

exercises/concept/jedliks-toy-car/.meta/design.md
xinri: Sorry yes I mean fields.

exercises/concept/jedliks-toy-car/.meta/design.md
xinri: I will remove this, I saw the unit tests that just create the instance with the new (and not the buy) and will test the battery and the distance. I thought perhaps the mentee can do a -- for battery or ++ for the distance when calling the function batterydisplay and distanceDisplay

exercises/concept/jedliks-toy-car/.meta/design.md
xinri: > I don't think we should be insisting that there are specific fields for battery percentage and driving distance. To be honest, I'm not sure how reliably an analyzer implementation can tell which fields is for battery percentage and driving distance. Instead, I'd suggest simply checking that the class has fields.

mmmh ok. thanks for taking account the analyzer

kahgoh (approved):
Looks good now! Thanks!

PR #3051, opened 2025-11-16
sync the largest-series-product tests and fix / add missing unit tests

xinri:
# pull request
Sync the largest-series-product. Missing reimplement (re-generated with configlet)

related to #2959 

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/check-deprecated-exercises.yml
+++ b/.github/workflows/check-deprecated-exercises.yml
@@ -2,6 +2,12 @@ name: Deprecated
 
 on:
   pull_request:
+    paths:
+      - "exercises/**"
+  push:
+    branches:
+      - main
+  workflow_dispatch:
 
 jobs:
   test-deprecated:
--- a/exercises/practice/crypto-square/.meta/config.json
+++ b/exercises/practice/crypto-square/.meta/config.json
@@ -21,6 +21,7 @@
     "sshine",
     "stkent",
     "vdemeester",
+    "Xinri",
     "Zaldrick"
   ],
   "files": {
--- a/exercises/practice/crypto-square/.meta/tests.toml
+++ b/exercises/practice/crypto-square/.meta/tests.toml
@@ -32,3 +32,8 @@ description = "8 character plaintext results in 3 chunks, the last one with a tr
 
 [fbcb0c6d-4c39-4a31-83f6-c473baa6af80]
 description = "54 character plaintext results in 7 chunks, the last two with trailing spaces"
+include = false
+
+[33fd914e-fa44-445b-8f38-ff8fbc9fe6e6]
+description = "54 character plaintext results in 8 chunks, the last two with trailing spaces"
+reimplements = "fbcb0c6d-4c39-4a31-83f6-c473baa6af80"
--- a/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
+++ b/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
@@ -77,8 +77,8 @@ public class CryptoSquareTest {
 
     @Disabled("Remove to run test")
     @Test
-    @DisplayName("54 character plaintext results in 7 chunks, the last two with trailing spaces")
-    public void fiftyFourCharacterPlaintextResultsInSevenChunksWithTrailingSpaces() {
+    @DisplayName("54 character plaintext results in 8 chunks, the last two with trailing spaces")
+    public void fiftyFourCharacterPlaintextResultsInEightChunksWithTrailingSpaces() {
         CryptoSquare cryptoSquare = new CryptoSquare("If man was meant to stay on the ground, god would have " +
                 "given us roots.");
         String expectedOutput = "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau ";
--- a/exercises/practice/two-bucket/.meta/config.json
+++ b/exercises/practice/two-bucket/.meta/config.json
@@ -14,7 +14,8 @@
     "muzimuzhi",
     "sjwarner-bp",
     "SleeplessByte",
-    "sshine"
+    "sshine",
+    "Xinri"
   ],
   "files": {
     "solution": [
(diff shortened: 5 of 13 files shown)


exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
jagdish-15, line 127:
```suggestion
                .withMessage("span must not exceed string length");
```

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
```diff
@@ -130,7 +130,29 @@ public void testSeriesLengthLongerThanLengthOfStringToTestIsRejected() {
...
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("reports 1 for nonempty string and empty product (0 span)")
+    public void testNonEmptyStringToSearchAndEmptyProduct() {
+        LargestSeriesProductCalculator calculator = new LargestSeriesProductCalculator("123");
+
+        long actualProduct = calculator.calculateLargestProductForSeriesLength(0);
+
+        assertThat(actualProduct).isEqualTo(1);
+    }
```
jagdish-15:
```suggestion
```

Remove the tests with `include = false` in the toml file

jagdish-15: Remove the tests which are marked as `include = false` in the toml file

jagdish-15 (changes requested):
Also check if there are any other tests that are marked as `include = false` but still exist in the test file. And please make all the error messages consistent with those in the [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/main/exercises/largest-series-product/canonical-data.json) file. I’ve suggested changes for one of them, but I believe there are more cases like that, for example the `rejects negative span` test.

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: Well it is indeed what the canonical said but I read that the implementation of the error messages should be made easy to read by the devs. I checked a lot of unit tests and their canonical file, and they have their own way to express what the canonical file raised as an error. So we should not just copy paste the error message but we should create an explicit message that will represent the error message in the canonical file no?

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: done thanks for checking that

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: Indeed, the message in the specification is not representing the error message.
I have replace it with "Series length must not exceed the length of the string to search." to match with the function.

By the way, if some mentee works on this exercise, they will have to rework again no ?

kahgoh: I'd prefer to leave the exception messages as they were. I think your wording is saying the same thing as before, just with different wording.  

> By the way, if some mentee works on this exercise, they will have to rework again no ?

Yeah, if we change the wording, the students will need to update their solutions. Since the wording is saying the same thing, I don't think there is much benefit with changing the exception message (hence, my preference for leaving the exception message as is).

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
```diff
@@ -124,18 +124,18 @@ public void testSeriesLengthLongerThanLengthOfStringToTestIsRejected() {
...
-    @DisplayName("reports 1 for empty string and empty product (0 span)")
-    public void testEmptyStringToSearchAndSeriesOfNonZeroLengthIsRejected() {
+    @DisplayName("rejects empty string and nonzero span")
+    public void testEmptyStringAndNonZeroSpanIsRejected() {
         LargestSeriesProductCalculator calculator = new LargestSeriesProductCalculator("");
 
         assertThatExceptionOfType(IllegalArgumentException.class)
                 .isThrownBy(() -> calculator.calculateLargestProductForSeriesLength(1))
-                .withMessage("Series length must be less than or equal to the length of the string to search.");
+                .withMessage("Series length must not exceed the length of the string to search.");
```
kahgoh: As mentioned in another comment, let's keep the message as it was.

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
```diff
@@ -155,7 +155,7 @@ public void testNegativeSeriesLengthIsRejected() {
 
         assertThatExceptionOfType(IllegalArgumentException.class)
                 .isThrownBy(() -> calculator.calculateLargestProductForSeriesLength(-1))
-                .withMessage("Series length must be non-negative.");
+                .withMessage("Series length must not be negative.");
```
kahgoh: As mentioned in another comment, let's keep the message as it was.

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: ok I have done a rollback for this

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: ok I have done a rollback for this

exercises/practice/largest-series-product/src/test/java/LargestSeriesProductCalculatorTest.java
xinri: sorry I didn't see this answer. Thanks for the confirmation. I also prefer to keep the same error message because they are quite explicits

Issue #2971, opened 2025-07-07
Add `@DisplayName` to Test Methods in Practice Exercises


jagdish-15:
## 📝 Issue Description

This issue aims to add `@DisplayName()` annotations for **every test method** in the test files of all **practice exercises** in the Java track.

### 🔧 What You Need to Do

1. **Skip unsynced exercises:**  
   Before starting, make sure the exercise is **not** listed among the ones with unsynced tests in [issue #2959](https://github.com/exercism/java/issues/2959).

2. **Locate the test file:**
   - Go to the exercise folder inside `exercises/practice/<exercise-name>/src/test`.
   - Open the Java test file (typically named `<ExerciseName>Test.java`).

3. **Find the canonical source:**
   - Navigate to the [problem-specifications](https://github.com/exercism/problem-specifications) repo.
   - Open the corresponding exercise folder and locate `canonical-data.json`.

4. **Match and annotate tests:**
   - The test cases in the `.json` file are listed in a specific order.
   - Match the tests in the Java file with the ones in `canonical-data.json`.
   - For each test, copy the `"description"` and use it as a `@DisplayName("")` annotation above the corresponding method in the Java test file.

## ⚠️ Important Guidelines

- **Max 15 exercises per PR:**  
  To avoid large PRs, limit the number of exercises being updated in each PR to a maximum of 15 exercises (less is ok too). 
  
- **Do not close this issue via PRs:**  
  Avoid referencing this issue in your PR with terms like `Fixes #XXX` or `Resolves #XXX`, since this would automatically close the issue. 

- **Avoid reimplemented exercises:**  
  Make sure you are not copying descriptions from tests that have been reimplemented.

- **Need help?**  
  If you have any questions, feel free to ask in the comments!

---

### 📋 Checklist of Practice Exercises

- [x] acronym
- [x] affine-cipher
- [x] all-your-base
- [x] allergies
- [x] alphametics
- [x] anagram
- [x] armstrong-numbers
- [x] atbash-cipher
- [x] bank-account
- [x] binary-search-tree
- [x] binary-search
- [x] bob
- [x] book-store
- [x] bottle-song
- [x] bowling
- [x] change
- [x] circular-2buffer
- [x] clock
- [x] collatz-conjecture
- [x] complex-numbers
- [x] connect
- [x] crypto-square
- [x] custom-set
- [x] darts
- [x] diamond
- [x] difference-of-squares
- [x] dnd-character
- [x] dominoes
- [x] dot-dsl
[...]

PR #3052, opened 2025-11-16
sync the two-bucket tests and fix / add missing unit tests with display name 

xinri:
# pull request

Sync two-bucket tests (implemented but not sync)
And implements the two missing tests 

related to #2959 and https://github.com/exercism/java/issues/2971

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/crypto-square/.meta/config.json
+++ b/exercises/practice/crypto-square/.meta/config.json
@@ -21,6 +21,7 @@
     "sshine",
     "stkent",
     "vdemeester",
+    "Xinri",
     "Zaldrick"
   ],
   "files": {
--- a/exercises/practice/crypto-square/.meta/tests.toml
+++ b/exercises/practice/crypto-square/.meta/tests.toml
@@ -32,3 +32,8 @@ description = "8 character plaintext results in 3 chunks, the last one with a tr
 
 [fbcb0c6d-4c39-4a31-83f6-c473baa6af80]
 description = "54 character plaintext results in 7 chunks, the last two with trailing spaces"
+include = false
+
+[33fd914e-fa44-445b-8f38-ff8fbc9fe6e6]
+description = "54 character plaintext results in 8 chunks, the last two with trailing spaces"
+reimplements = "fbcb0c6d-4c39-4a31-83f6-c473baa6af80"
--- a/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
+++ b/exercises/practice/crypto-square/src/test/java/CryptoSquareTest.java
@@ -77,8 +77,8 @@ public class CryptoSquareTest {
 
     @Disabled("Remove to run test")
     @Test
-    @DisplayName("54 character plaintext results in 7 chunks, the last two with trailing spaces")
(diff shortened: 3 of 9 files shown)


exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -67,6 +73,8 @@ public void testBucketOneSizeOneBucketTwoSizeThreeStartWithTwo() {
 
     @Disabled("Remove to run test")
     @Test
+    @DisplayName("Measure using bucket one of size 2 and bucket two of size 3 - " +
```
xinri: The message exceeds 120 characters. So either we modify this in the [specification](https://github.com/exercism/problem-specifications/blob/main/exercises/two-bucket/canonical-data.json) or I have to split in two lines

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -67,6 +73,8 @@ public void testBucketOneSizeOneBucketTwoSizeThreeStartWithTwo() {
...
+            "start with bucket one and end with bucket two")
```
jagdish-15:
```suggestion
    @DisplayName(
        "Measure using bucket one of size 2 and bucket two of size 3 - " +
        "start with bucket one and end with bucket two"
    )
```
You can do this for the line being more than 120 characters. It aligns with what's been followed throughout the repo.

jagdish-15 (changes requested):
@xinri
Thanks for picking this up! Please apply the following change:

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
xinri: Thanks @jagdish-15 for the suggestion, I will update the MR when I am back home. I prefer to amend my commit with the change ;)

xinri:
> I think you accidentally pushed the same commit without amending the changes. Could you please check?

No I really push it. It is just that the second line has been indented by the code formatter.
I put the name on the next line and the second line just after.

exercises/practice/two-bucket/src/test/java/TwoBucketTest.java
```diff
@@ -67,6 +73,9 @@ public void testBucketOneSizeOneBucketTwoSizeThreeStartWithTwo() {
 
     @Disabled("Remove to run test")
     @Test
+    @DisplayName(
+            "Measure using bucket one of size 2 and bucket two of size 3 - " +
+                    "start with bucket one and end with bucket two")
```
jagdish-15:
```suggestion
    @DisplayName(
        "Measure using bucket one of size 2 and bucket two of size 3 - " +
        "start with bucket one and end with bucket two"
    )
```
> No I really push it. It is just that the second line has been indented by the code formatter.

Press the commit suggestion button. It won't amend the change into the same commit like you wanted, but it will get the job done and you won’t run into the code formatter issue. And since we squash and merge, the extra commits don’t matter much anyway.

xinri: Thanks done

PR #3077, opened 2026-01-11
Update tests bird watcher

jagdish-15:
# pull request

This would catch solutions like these:

```
    public int[] getLastWeek() {
        return birdsPerDay;
    }
```

```
    public void incrementTodaysCount() {
        for (int i = 1; i < birdsPerDay.length; i++) {
            birdsPerDay[i] = birdsPerDay[i] + 1;
        }
    }
```

Variants like this have appeared in student solutions and currently pass the tests, even though they don’t align with the exercise instructions

---

Reviewer Resources:


exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java
```diff
@@ -1,34 +1,35 @@
...
 
     private BirdWatcher birdWatcher;
-    private final int[] lastWeek = {DAY1, DAY2, DAY3, DAY4, DAY5, DAY6, TODAY};
+    private final int[] currentWeek = {DAY1, DAY2, DAY3, DAY4, DAY5, DAY6, TODAY};
+    private final int[] lastWeek = {0, 2, 5, 3, 7, 8, 4};
 
     @BeforeEach
     public void setUp() {
-        birdWatcher = new BirdWatcher(lastWeek);
+        birdWatcher = new BirdWatcher(currentWeek);
```
kahgoh: This is going to invalidate solutions that return the array in `getLastWeek`, like [this](https://exercism.org/tracks/java/exercises/bird-watcher/solutions/kapitanov) one. 

If our objective is to just capture the `incrementTodaysCount`, I think we could just change `itIncrementTodaysCount`.

exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java
```diff
@@ -42,13 +43,16 @@ public void itTestGetToday() {
     @Tag("task:3")
     @DisplayName("The incrementTodaysCount method correctly increments today's counts")
     public void itIncrementTodaysCount() {
+        int firstSixDaysBeforeIncrement = birdWatcher.getCountForFirstDays(6);
         birdWatcher.incrementTodaysCount();
+        int firstSixDaysAfterIncrement = birdWatcher.getCountForFirstDays(6);
         assertThat(birdWatcher.getToday()).isEqualTo(TODAY + 1);
+        assertThat(firstSixDaysAfterIncrement).isEqualTo(firstSixDaysBeforeIncrement);
```
kahgoh: Ah, this is a clever way (using `getCountForFirstDays`). Just couple of minor comments for this one:
- We could simplify this by doing calculating the expected value directly, like we currently do in the other test for `getCountForFirstDays` (i.e. `assertThat(birdWatcher.getCountForFirstDays(6)).isEqualTo(DAY1 + DAY2 + DAY3 + DAY4 + DAY5 + DAY6);`).
- I think it can be confusing why we're asserting `getCountForFirstDays` in this test. I'd suggest moving to a separate test so that the test can be described in the `@DisplayName` and test name. For example:

```java
public void itIncrementDoesNotChangeCountForOtherDays() {
  birdWatcher.incrementTodaysCount();
  assertThat(birdWatcher.getCountForFirstDays(6)).isEqualTo(DAY1 + DAY2 + DAY3 + DAY4 + DAY5 + DAY6);
}
```

exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java
jagdish-15: I see what you’re getting at, but this solution is fundamentally incorrect:

```
    public int[] getLastWeek() {
        return birdsPerDay;
    }
```

This method returns the **current week**, not the **previous week**. 
According to the exercise instructions: "For comparison purposes, you always keep a copy of last week's counts nearby, which were: 0, 2, 5, 3, 7, 8 and 4. "

If `birdsPerDay` were truly representing the previous week, then the last index of the array could not correspond to `today`. That alone makes this implementation inconsistent with the intended model.

These solutions were effectively getting a free pass because the test data happened to use the same values for both the current week and the “previous week” mentioned in the introduction. As a result, logically incorrect implementations still passed the tests.

Curious to hear your thoughts on this.

kahgoh: I can see where you are coming from. The instructions for `getLastWeek` seem to imply students should always be defining the returned array. This would be within scope of this exercise (`design.md` mentions one of the objects is to be able to define an array).

But I think the setup with the two arrays in the test class could be confusing as their usage or intent might not be as clear. I prefer the [C# implementation](https://github.com/exercism/csharp/blob/main/exercises/concept/bird-watcher/BirdWatcherTests.cs), where each test creates the `BirdWatcher` with its own input and `getLastWeek` is like a `static` method (so that we don't have to make an instance of `BirdWatcher` to call `getLastWeek`). 

_Alternatively_, if we were to stick with a common array that gets reused, I'd preferring sticking to just one array in the test class that is used by the test method instead of having two arrays. In this case, we could change the array contents in this PR so that it is no longer the matches the expected `lastWeek`.

exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java
```diff
@@ -1,102 +1,107 @@
...
+
+    @Test
+    @Tag("task:3")
+    @DisplayName("The incrementTodaysCount does not change count for other days")
+    public void itIncrementDoesNotChangeCountForOtherDays() {
+        int[] counts = new int[] {5, 1, 0, 4, 2, 3, 0};
+        BirdWatcher birdWatcher = new BirdWatcher(counts);
         birdWatcher.incrementTodaysCount();
-        assertThat(birdWatcher.getToday()).isEqualTo(TODAY + 1);
+        assertThat(birdWatcher.getCountForFirstDays(6)).isEqualTo(15);
```
kahgoh: I wonder if using `getCountForFirstDays` here might be confusing for students (implementing `getCountForFirstDays` is task 5, which comes later). Would it make sense to tag it "task:5" and move it after the other `getCountForFirstDays` test? The `assertThat` is still using `getCountForFirstDays`. The test could also be described in terms of `getCountForFirstDays` instead. For example "_incrementTodaysCount adds one to getCountForFirstDays_"

jagdish-15: Makes sense!

exercises/concept/bird-watcher/src/test/java/BirdWatcherTest.java
```diff
@@ -87,6 +77,18 @@ public void itTestGetCountForMoreDaysThanTheArraySize() {
...
+    @Test
+    @Tag("task:5")
+    @DisplayName("The incrementTodaysCount method adds one to getCountForFirstDays method")
+    public void itIncrementDoesNotChangeCountForOtherDays() {
+        int[] counts = new int[] {5, 1, 0, 4, 2, 3, 0};
+        BirdWatcher birdWatcher = new BirdWatcher(counts);
+        int countPriorIncrement = birdWatcher.getCountForFirstDays(7);
+        birdWatcher.incrementTodaysCount();
+        int countAfterIncrement = birdWatcher.getCountForFirstDays(7);
+        assertThat(countPriorIncrement).isEqualTo(countAfterIncrement - 1);
```
kahgoh: I'd recommend using `assertThat(countPriorIncrement).isEqualTo(15);` and `assertThat(countAfterIncrement ).isEqualTo(16);` instead. If both `countPriorIncrement` and `countAfterIncrement` are wrong, it may still be possible for `assertThat(countPriorIncrement).isEqualTo(countAfterIncrement - 1)` to pass.

kahgoh (approved):

PR #3102, opened 2026-04-12
Update Need for Speed drained battery instructions for clarity

ronvoluted:
The tests check that if battery level is 99/100 but battery drain would be 100, that the car is technically 'drained'. This PR updates the instructions to reflect this to avoid ambiguity. A greater change would be change the wording to not describe it as drained (i.e. `batteryDrained` -> `batteryEnough`) but that's more of a subjective change, so this  PR keeps to a smaller scope.

---

Reviewer Resources:

### DIFF
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -45,7 +45,7 @@ car.distanceDriven();
 
 ## 4. Check for a drained battery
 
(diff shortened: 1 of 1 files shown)


kahgoh (approved):
Hi @ronvoluted, thanks for this!

ronvoluted:
Cheers @kahgoh! It certainly had me stuck for longer than expected 😅

Issue #1803, opened 2020-04-15
Should we allow other testing frameworks in the java track?


jmrunkle:
Expanding this discussion from #1634.

Currently, we *mostly* only use the generic JUnit asserting framework (eg. `org.junit.Assert.*`), but we do not have any officially documented policy in our contributing guide about using other testing frameworks. We should either explicitly allow or disallow the other frameworks.

Arguments for using only the generic JUnit Assert methods:

*   We want it to be as easy as possible for people to do these exercises and therefore want to limit the different libraries and syntax they need to understand to use the tests.
*   This is consistent with our policy to only use the Java standard library for our exercises.

Arguments for allowing one or more additional testing frameworks:

*   They provide more detailed errors that make it easier for students to debug their tests. 
*   `org.junit.Assert` methods do not distinguish clearly between the expected and the actual so students have to look at the test code to see what was actually expected rather than getting that information straight from the error itself.
*   Exposure to the larger set of test frameworks is a useful skill for the students to aquire.
*   Any reasonable testing framework should be easy to read even without lots of experience. 

If we are going to allow another test framework, I suggest that we should choose one framework to make the experience consistent while they complete exercises. Here are the possible contenders:

*   [Hamcrest](http://hamcrest.org/JavaHamcrest/)
*   [Truth](https://truth.dev/)
*   [AssertJ](https://assertj.github.io/doc/)

In my experience, Hamcrest is the least clear and the most bizarre to write tests with its fuzzy matcher style. As such, I would prefer to ban Hamcrest from the Java track (even though it has already been used in a couple places).

I have most experience with Truth (it is what I use daily at work) and find that the errors and style of writing the tests is most straightforward (for example, an IDE will suggest helpful autocompletions). This is my preferred option.

Looking at AssertJ, it seems to also have a very reasonable style for writing tests, but I have little to no experience there to comment on how effective the error messages are, etc. According to Truth's documents comparing with AssertJ, they suggest that it is another good option with both having their own strengths and weaknesses.

So, there are (up to) two questions open for maintainers:

1. Should we allow any additional test frameworks?
2. If so, which test framework(s) should we allow?

jmrunkle:
@jtigger @matthewmorgan @sit @stkent @Smarticles101 @FridaTveit @sjwarner-bp @lemoncurry @sshine @msomji @mirkoperillo @pdmoore 

I think that's the only way to get everyone's attention here. If you don't have an opinion, just ignore this.

[...]

PR #3109, opened 2026-04-22
[hangman] Sync tests & Update solution to match canonical rules

thibault2705:
# pull request

What has been done:
- Rename and simplify method `play(Observable<String> words, Observable<String> letters)` to `guess(String word, Observable<String> letters)`
- Add, rename output params `maskedWord`, `state`, and `remainingFailures` to match canonical expectations
- Introduced MAX_FAILURES = 10 based on canonical rules
- Added error handling for guesses after win and lose states
- Aligned test cases with canonical scenarios

This PR partially solves issues #2959.

---

Reviewer Resources:


exercises/practice/hangman/src/main/java/Hangman.java
```diff
@@ -2,7 +2,7 @@
 
 class Hangman {
 
-    Observable<Output> play(Observable<String> words, Observable<String> letters) {
+    Observable<Output> guess(String word, Observable<String> letters) {
```
kahgoh: I think this change is because we have removed the "consecutive games" test, which isn't in the problem specifications. To me, if we are going to change the first argument from `Observable` to a `String`, then I think we should also remove the other `Observable` usages. `letters` can become a  list and the return could be `Output` directly. This would simplify the exercise and remove the RxJava dependency. The exercise's `.docs/instructions.append.md` should also be deleted if we do this.

The other alternative is to keep the "consecutive games" test case and keep using RxJava. We can have Java track specific tests.

exercises/practice/hangman/src/main/java/Output.java
```diff
@@ -36,4 +40,51 @@ static Output empty() {
...
+        return "Output{" +
+            "secret='" + word + '\'' +
+            ", discovered='" + maskedWord + '\'' +
+            ", guess=" + guesses +
+            ", misses=" + misses +
+            ", parts=" + parts +
+            ", status=" + state +
+            ", remainingFailures=" + remainingFailures +
+            '}';
+    }
```
kahgoh: Do all these methods really need to be added? I don't see them used by the tests. If they needed by the student solution, students could write them.

exercises/practice/hangman/src/main/java/Output.java
```diff
@@ -1,29 +1,33 @@
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
kahgoh: Please don't use the `*` import.

```suggestion
import java.util.Collections;
import java.util.List;
import java.util.Set;
```

exercises/practice/hangman/src/main/java/Output.java
thibault2705: You're reasonable, I removed these methods, let the students write them in case.

exercises/practice/hangman/src/main/java/Output.java
thibault2705: I updated my IDE settings to prevent this style in the future.

exercises/practice/hangman/src/main/java/Hangman.java
thibault2705: I resolved this review by removing the usage of `Observable` from the RxJava dependency, making `letters` a `List`, and simplifying the method return with `Output` directly. The unused instructions file was also removed.

exercises/practice/hangman/src/main/java/Output.java
```diff
@@ -1,39 +1,32 @@
...
+    static final int MAX_FAILURES = Part.values().length;
+
+    public final String word;
+    public final String maskedWord;
+    public final Set<String> guesses;
     public final Set<String> misses;
     public final List<Part> parts;
-    public final Status status;
+    public final Status state;
+    public final int remainingFailures;
```
kahgoh, line 14: Looking at the tests, I notice not all of these fields are checked. Do we need all of them? If not, consider removing the ones no longer needed.

exercises/practice/hangman/src/test/java/HangmanTest.java
```diff
@@ -1,29 +1,15 @@
...
 import org.junit.jupiter.api.Test;
 
-import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Stream;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
kahgoh, line 9: We generally prefer using AssertJ over JUnit assertions (see issue [#1803](https://github.com/exercism/java/issues/1803)).

kahgoh (commented):
Hey @thibault2705, sorry this review is taking a while. I have noticed over on the forums, there is a [proposal to replace this Hangman exercise with Save the Cow](https://forum.exercism.org/t/replace-hangman-with-non-violent-content/34156/16), which is going through review at the moment. I would like to suggest holding off on this one for a bit so that we can move to Save the Cow. I think the canonical data is largely the same, so I think you'd be able to most of this to the new exercise and rename things.

kahgoh:
@thibault2705, [Save the Cow](https://github.com/exercism/problem-specifications/tree/main/exercises/save-the-cow) has just been merged into the problem specifications, deprecating Hangman. If you're interested in updating the exercise to _Save the Cow_, if you need a rough guide, I think this would roughly involve:
1. Use `configlet` to create the exercise (`configlet create --exercise save-the-cow --practice-exercise save-the-cow`)
2. Copy the Java files from your Hangman work to Save the Cow and update the sources accordingly.
3. You'll probably also need to copy over the Gradle files and some settings from `config.json`.
4. Revert changes to the Hangman exercise. 
5. In the track's config.json, mark Hangman as deprecated (add ` "status": "deprecated"` to Hangman).
6. In `exercises/settings.gradle`, comment out `include 'practice:hangman'` and add `include 'practice:save-the-cow'`

kahgoh:
@thibault2705 are you still working on this?

IsaacG:
@thibault2705 Ping? Are you still interested here?

PR #3112, opened 2026-04-26
Curate practice exercises per concept

bull1210:
## What's the problem?

The concept tree for the Java track is pretty hard to navigate right now. Every concept ends up with a huge wall of practice exercises listed under it â€” strings alone shows 33+ exercises. That number is misleading because most of those exercises don't really *teach* strings, they just happen to use strings somewhere along the way.

This seems to have crept in during the v3 migration, where the `practices` field got auto-populated based on prerequisites rather than being thoughtfully assigned.

## What this PR does

Goes through every concept and hand-picks 2â€“4 practice exercises that actually make sense as follow-up practice for a student who just learned that concept. Exercises were chosen based on whether the concept is genuinely the *main thing* you're working with in that exercise, not just a side dependency.

Here's what ended up assigned:

| Concept | Practice exercises |
|---|---|
| strings | reverse-string, acronym, isogram, bob |
| numbers | armstrong-numbers, grains, difference-of-squares, space-age |
| chars | pangram, isbn-verifier, rotational-cipher, run-length-encoding |
| maps | nucleotide-count, relative-distance, word-count, etl |
| classes | dot-dsl, robot-name, linked-list, clock |
| if-else-statements | leap, darts, raindrops |
| for-loops | hamming, luhn, twelve-days |
| generic-types | flatten-array, sublist, custom-set |
| exceptions | collatz-conjecture, error-handling, variable-length-quantity |
| lists | high-scores, sieve, grade-school |
| enums | secret-handshake, yacht, robot-simulator |
| constructors | triangle, bank-account, clock |
| booleans | leap, allergies |
| bit-manipulation | eliuds-eggs, allergies |
| foreach-loops | rna-transcription, atbash-cipher |
| datetime | gigasecond, meetup |
| randomness | dnd-character, robot-name |
| switch-statement | scrabble-score, robot-simulator |
| arrays | resistor-color, protein-translation, food-chain |
| sets | baffling-birthdays |

Six concepts (`basics`, `inheritance`, `interfaces`, `method-overloading`, `nullability`, `ternary-operators`) have no practice exercises assigned â€” there just aren't existing exercises in the track that specifically target those concepts, so I left them empty rather than forcing a bad fit.

## Notes

- No exercise content was changed, only `config.json`
- Deprecated exercises were skipped entirely
- A few exercises appear under two concepts where they genuinely practice both (e.g. `leap` under both `if-else-statements` and `booleans`, `allergies` under both `booleans` and `bit-manipulation`)


kahgoh:
> The concept tree for the Java track is pretty hard to navigate right now. Every concept ends up with a huge wall of practice exercises listed under it â€” strings alone shows 33+ exercises. 

Are you talking about the tree on the [concepts](https://exercism.org/tracks/java/concepts) page? I only see only one exercise (Log Levels) listed for Strings, so this doesn't seem right to me ...

bull1210:
You're right, sorry for the confusing description. 
The current state of the repo has "practices": [] on almost all practice exercises, so the concept tree shows very few (or no) practice exercises per concept.

The PR is actually is populating the empty practices fields with a curated selection of 2–4 exercises per concept. The goal is to give students who have just finished a concept exercise some concrete practice exercises to try next, rather than leaving the concept page with no suggested follow-up.

IsaacG:
See https://exercism.org/docs/building/tracks/config-json#h-practice-exercises which documents the `practices` array.

@exercism/java I think this PR looks reasonable. Am I missing something?

kahgoh:
Sorry @bull1210 , I had somehow forgotten about this one. At a glance, I think it looks like a good starting point. Just a couple of thoughts on this one:

Leap - Boolean seems good, but I noticed there are many community solutions that solve it without if-else statements. Suggest dropping if-else statement.
Bank Account - Similar, there is quite a number of solutions without having to implement a constructor and would suggest removing this mapping.

PR #3114, opened 2026-04-26
fix: correct factual errors in docs and CI config

bull1210:
Small factual corrections scattered across the docs and CI config.

**INSTALLATION.md: wrong Java LTS version**

The file says "Java 25 LTS" but Java 25 hasn't been released — the current LTS is Java 21. Changed to 21.

**RESOURCES.md: stale API link**

The official Java documentation link pointed at the Java 11 API. The track CI runs on Java 21, so updated the link to match.

**nullability/about.md: C# syntax error**

Line 31 uses `arr.Length` which is C# — in Java it's `arr.length` (lowercase). Small but confusing when it's in a code example students are expected to learn from.

**dependabot.yml: add Gradle ecosystem**

The existing Dependabot config only watches GitHub Actions. Adding a Gradle entry so that JUnit and AssertJ version bumps get picked up automatically instead of going stale.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,10 +7,3 @@ updates:
       interval: 'monthly'
     labels:
       - 'x:size/tiny'
-
-  - package-ecosystem: 'gradle'
-    directory: '/'
-    schedule:
-      interval: 'monthly'
-    labels:
-      - 'x:size/tiny'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,10 @@ updates:
       interval: 'monthly'
     labels:
       - 'x:size/tiny'
+
+  - package-ecosystem: 'gradle'
+    directory: '/'
+    schedule:
+      interval: 'monthly'
+    labels:
+      - 'x:size/tiny'
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -9,7 +9,7 @@ Here at Exercism we recommend using [Eclipse Temurin by Adoptium][adoptium].
 
 ## Supported Java versions
 
-Exercism's Java track supports Java 21 LTS and earlier.
+Exercism's Java track supports Java 25 LTS and earlier.
 
 ## Installing Eclipse Temurin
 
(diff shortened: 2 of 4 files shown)


kahgoh:
> The file says "Java 25 LTS" but Java 25 hasn't been released — the current LTS is Java 21. Changed to 21.

No, Java 25 _is_ LTS and has been released. The [Adoptium](https://adoptium.net/temurin/releases) and [Oracle](https://www.oracle.com/anz/java/technologies/downloads/) builds are available.

docs/RESOURCES.md
```diff
@@ -2,5 +2,5 @@
 
 - [Stack Overflow](http://stackoverflow.com/questions/tagged/java).
 - [The Java subreddit](https://www.reddit.com/r/java)
-- [Official Java documentation](https://docs.oracle.com/en/java/javase/11/docs/api/index.html)
+- [Official Java documentation](https://docs.oracle.com/en/java/javase/21/docs/api/index.html)
```
kahgoh: Might as well update this to the Java 25 docs: https://docs.oracle.com/en/java/javase/25/docs/api/index.html

.github/dependabot.yml
kahgoh, line 1: I think we should this out for now since updating things like JUnit and AssertJ also requires the test runner to be updated.

docs/INSTALLATION.md
```diff
@@ -9,7 +9,7 @@ Here at Exercism we recommend using [Eclipse Temurin by Adoptium][adoptium].
 
 ## Supported Java versions
 
-Exercism's Java track supports Java 25 LTS and earlier.
+Exercism's Java track supports Java 21 LTS and earlier.
```
kahgoh: Please revert this one. We currently support Java 25.

kahgoh (changes requested):

docs/RESOURCES.md
bull1210: Updated this link to Java 25 doc location.

.github/dependabot.yml
bull1210: removed that change in dependabot.yml

docs/INSTALLATION.md
bull1210: Reverted it.

kahgoh (approved):

PR #3115, opened 2026-04-26
improve: expand strings intro and rewrite inheritance concept docs

bull1210:
Quality improvements to two concept docs that had real problems.

**strings/introduction.md**

The existing file was 13 lines and only mentioned that strings exist and "have methods". Expanded it to cover the most commonly used String methods with examples, explain immutability properly, show why `.equals()` matters over `==`, and introduce `StringBuilder` and `String.format` for building strings dynamically.

**inheritance/introduction.md**

Rewrote this one from scratch. The old version had a few issues:
- Used `bark()` on a parent `Animal` class — semantically odd since lions don't bark
- "It literally means in programming as it means in english" is awkward
- Only linked to javatpoint.com instead of the Oracle docs
- Didn't mention `super` or abstract classes at all

The rewrite uses a proper `makeSound()` method, introduces `super` with a real example, covers abstract classes, and links to the official Java tutorial.

**inheritance/about.md**

Fixed a syntax error in the interface example: `interface Animal()` — interfaces don't have `()` after the name. Changed to `interface Animal` and updated the method signature to `makeSound()` to match the introduction.

### DIFF
--- a/concepts/inheritance/introduction.md
+++ b/concepts/inheritance/introduction.md
@@ -1,58 +1,88 @@
 # Introduction
 
-Inheritance is a core concept in OOP (Object-Oriented Programming).
-It represents an IS-A relationship.
-It literally means in programming as it means in english, inheriting features from parent (in programming features is normally functions and variables).
-
-Consider a class, `Animal` as shown,
+Inheritance lets one class build on top of another, reusing its fields and methods without rewriting them.
+The class being extended is called the **parent** (or superclass); the class doing the extending is the **child** (or subclass).
+The relationship is described as IS-A: a `Dog` IS-A `Animal`.
 
 ```java
-//Creating an Animal class with bark() as a member function.
 public class Animal {
-
-    public void bark() {
-        System.out.println("This is an animal");
+    public void makeSound() {
+        System.out.println("...");
     }
-
 }
 ```
 
-`Animal` is a parent class, because the properties this class has can be extended to all the animals in general.
-
-Consider an animal named `Lion`, having a class like,
+Use the `extends` keyword to inherit from a class:
 
 ```java
-//Lion class is a child class of Animal.
-public class Lion extends Animal {
+public class Dog extends Animal {
 
     @Override
-    public void bark() {
-        System.out.println("Lion here!!");
+    public void makeSound() {
+        System.out.println("Woof!");
     }
-
 }
 ```
 
 ~~~~exercism/note
-The `Override` annotation is used to indicate that a method in a subclass is overriding a method of its superclass.
-It's not strictly necessary but it's a best practice to use it.
+The `@Override` annotation tells the compiler that you intend to replace a parent method.
+It's not required, but it's considered good practice — the compiler will warn you if the method name doesn't actually match anything in the parent.
 ~~~~
 
-Now whenever we do,
+Because `Dog` extends `Animal`, a `Dog` instance can be stored in an `Animal` variable.
+The actual type at runtime decides which method runs — this is called **polymorphism**:
 
 ```java
-Animal animal = new Lion(); //creating instance of Animal, of type Lion
-animal.bark();
+Animal animal = new Dog();
+animal.makeSound(); // prints "Woof!", not "..."
 ```
 
-Note: Initialising the `Animal` class with `Lion`.
-The output will look like
+## The `super` keyword
+
+Inside a child class, `super` refers to the parent.
+You can use it to call the parent's constructor or a method you've overridden:
 
 ```java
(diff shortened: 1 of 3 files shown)


concepts/inheritance/introduction.md
```diff
@@ -1,58 +1,88 @@
...
 
 ```java
-//Lion class is a child class of Animal.
-public class Lion extends Animal {
+public class Dog extends Animal {
 
     @Override
-    public void bark() {
-        System.out.println("Lion here!!");
+    public void makeSound() {
```
kahgoh, line 21: Is it worth having an example of a child using something from the parent? The introduction opened by saying we can reuse fields  and methods, but the first example looks like we're rewriting (by overriding) `makeSound`.

concepts/inheritance/introduction.md
```diff
@@ -1,58 +1,88 @@
 # Introduction
 
-Inheritance is a core concept in OOP (Object-Oriented Programming).
-It represents an IS-A relationship.
-It literally means in programming as it means in english, inheriting features from parent (in programming features is normally functions and variables).
-
-Consider a class, `Animal` as shown,
+Inheritance lets one class build on top of another, reusing its fields and methods without rewriting them.
+The class being extended is called the **parent** (or superclass); the class doing the extending is the **child** (or subclass).
+The relationship is described as IS-A: a `Dog` IS-A `Animal`.
```
kahgoh, line 5: Suggest explicitly stating that `Dog` is extending `Animal` to make it really explicit and clear:

```suggestion
The relationship is described as IS-A. For example, if `Dog` class extends an `Animal` class, the relationship can be described as a `Dog` IS-A `Animal`.
```

concepts/inheritance/introduction.md
```diff
@@ -1,58 +1,88 @@
...
+    @Override
+    public void makeSound() {
+        super.makeSound(); // calls Animal's version first
+        System.out.println(name + " says: Woof!");
+    }
+}
+```
+
+## Abstract classes
+
```
kahgoh, line 63: I think Abstract classes is a bit out of place in the inheritance introduction and would probably be better in class's [about](https://github.com/exercism/java/blob/main/concepts/classes/about.md) instead or a separate concept that comes after both inheritance and classes concepts (of the two, I favor as a separate concept).

concepts/strings/introduction.md
```diff
@@ -7,7 +7,64 @@ Double quotes are used to define a `String` instance:
...
+
+```java
+StringBuilder sb = new StringBuilder();
+sb.append("Hello");
+sb.append(", ");
+sb.append("World!");
+String result = sb.toString(); // "Hello, World!"
+```
+
+For simple one-liner formatting, `String.format` is handy:
```
kahgoh: The general advice to use `StringBuilders` over `+` and `String.format` over `+` bugs me a bit.

concepts/strings/introduction.md
```diff
@@ -7,7 +7,64 @@ Double quotes are used to define a `String` instance:
...
+```java
+String a = "hello";
+String b = "hello";
+a.equals(b);            // true  (correct)
+a.equalsIgnoreCase(b);  // true  (case-insensitive version)
+```
+
+## Building strings dynamically
+
+When you need to construct a string piece by piece (e.g. inside a loop), use `StringBuilder` instead of concatenating with `+`.
```
kahgoh, line 54: Using `+` hasn't been introduced and there doesn't seem to be an example. I'd suggest adding one.

concepts/inheritance/introduction.md
```diff
@@ -1,58 +1,88 @@
...
+    }
+}
 ```
 
-According to OOP, there are many types of inheritance, but Java supports only some of them (Multi-level and Hierarchical).
-To read more about it, please read [this][java-inheritance].
+To read more about inheritance in Java, see the [official documentation][java-inheritance].
 
-[java-inheritance]: https://www.javatpoint.com/inheritance-in-java#:~:text=On%20the%20basis%20of%20class,will%20learn%20about%20interfaces%20later.
+[java-inheritance]: https://docs.oracle.com/javase/tutorial/java/IandI/subclasses.html
```
kahgoh, line 88: I'd suggest using Dev.java where possible. The top of Oracle page states that Dev.java has the latest stuff. The Dev.java page for inheritance is at For inheritance, the page is at https://dev.java/learn/inheritance/.

kahgoh (changes requested):

IsaacG:
@bull1210 Do you want to work out this or should it be closed out at abandoned?

PR #3116, opened 2026-04-26
add: hints.md for all difficulty-5 exercises (policy compliance)

bull1210:
Per the [track policy][policy], every exercise with difficulty 5 should have a `hints.md` explaining that no starter implementation is provided. Students coming from difficulty-4 exercises are used to getting a stub — hitting a blank file for the first time with no explanation is a bad experience.

Added the same standard hints file to all 21 difficulty-5 exercises that were missing it:

`allergies`, `atbash-cipher`, `bob`, `flatten-array`, `game-of-life`, `grep`, `knapsack`, `ledger`, `matching-brackets`, `nucleotide-count`, `pascals-triangle`, `phone-number`, `pig-latin`, `prime-factors`, `relative-distance`, `robot-name`, `run-length-encoding`, `series`, `square-root`, `state-of-tic-tac-toe`, `word-count`

The two exercises that already had hints.md (`mazy-mice` and `parallel-letter-frequency`) were left untouched.

[policy]: https://github.com/exercism/java/blob/main/POLICIES.md#add-hint-for-the-first-exercises-without-starter-implementation

### DIFF
--- a/exercises/practice/allergies/.docs/hints.md
+++ b/exercises/practice/allergies/.docs/hints.md
@@ -1,0 +1,10 @@
+# Hints
+
+## General
+
+- Unlike lower-difficulty exercises, this one does not come with a starter implementation file.
+  You will need to create the solution file yourself.
+- Look at the test file to figure out the expected class name and method signatures.
+- If you are not sure how to structure the file, check the [StubTemplate][stub-template] in the repository for a starting point.
+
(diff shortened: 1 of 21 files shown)


kahgoh:
Sorry, this might also be an outdated policy. Most exercises now have a stub implementation, so the hint may no longer be needed. See commit 2493e118d43216627e103bb0562890e15eafbf9f for example. Feel free to update the policy docs accordingly.

kahgoh:
The policy has been updated in PR #3122. Closing this PR as I think it isn't needed anymore.

PR #3118, opened 2026-04-28
[tree-building] Sync tests

thibault2705:
# pull request

Sync the `tree-building` exercise by adding the following missing test cases:
CleanShot 2026-04-28 at 16 44 30@2x

This PR partially solves issue [#2959 ](https://github.com/exercism/java/issues/2959)

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/tree-building/.meta/config.json
+++ b/exercises/practice/tree-building/.meta/config.json
@@ -7,7 +7,8 @@
     "lemoncurry",
     "msomji",
     "muzimuzhi",
-    "uzilan"
+    "uzilan",
+    "thibault2705"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/tree-building/.meta/src/reference/java/Record.java
+++ b/exercises/practice/tree-building/.meta/src/reference/java/Record.java
@@ -1,17 +1,3 @@
-class Record {
-    private int recordId;
-    private int parentId;
+record Record(int recordId, int parentId) {
 
-    public Record(int recordId, int parentId) {
-        this.recordId = recordId;
-        this.parentId = parentId;
-    }
-
-    int getParentId() {
-        return parentId;
-    }
-
-    int getRecordId() {
-        return recordId;
-    }
 }
--- a/exercises/practice/tree-building/src/main/java/TreeNode.java
+++ b/exercises/practice/tree-building/src/main/java/TreeNode.java
@@ -1,8 +1,8 @@
 import java.util.ArrayList;
 
 class TreeNode {
-    private int nodeId;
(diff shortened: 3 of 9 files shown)


exercises/practice/tree-building/src/main/java/TreeNode.java
```diff
@@ -1,8 +1,8 @@
 import java.util.ArrayList;
 
 class TreeNode {
-    private int nodeId;
-    private ArrayList<TreeNode> children;
+    private final int nodeId;
+    private final ArrayList<TreeNode> children;
```
kahgoh, line 5: I think when this class was first written, the `final` word was [deliberartely left out](https://github.com/exercism/java/blob/main/POLICIES.md#avoid-using-final).

exercises/practice/tree-building/.meta/config.json
```diff
@@ -7,7 +7,8 @@
     "lemoncurry",
     "msomji",
     "muzimuzhi",
-    "uzilan"
+    "uzilan",
+    "thibault2705"
```
jagdish-15, line 11: Please maintain alphabetical order while adding yourself as a contributor

exercises/practice/tree-building/.meta/src/reference/java/Record.java
```diff
@@ -1,17 +1,3 @@
-class Record {
-    private int recordId;
-    private int parentId;
+record Record(int recordId, int parentId) {
```
jagdish-15, line 1: I’m in favour of this change to convert the POJO to a Java record. @kahgoh, can we have your thoughts on this one?

jagdish-15 (changes requested):

exercises/practice/tree-building/.meta/src/reference/java/Record.java
kahgoh: Yeah, I'm also in favour of sticking with the original POJO. There seems to be a lot of changes just to sync the tests, some of which I think will definitely break existing solutions (this change to `record` is one of them). I'd certainly prefer trying to support existing solutions where possible, so I'd also prefer going back to the previous `class` implementation.

kahgoh (changes requested):
As mentioned in an earlier comment, there seems to be quite a number of changes that will break existing community solutions. I'd prefer to avoid them if possible. A couple of the other breaking changes I noticed:
- Changing to use a `static` method.
- The change to expected exception messages. I assume they were copied from the canonical data, but the change will cause existing solutions to fail.

IsaacG:
@thibault2705 Are you still wanting to work out this or should it be closed out?

PR #3124, opened 2026-05-08
Add lint-changed task in Java workflow

jagdish-15:
# Pull Request

We previously added `test-changed` to avoid the waiting time required to run tests on all exercises when only a few, or even just one, exercise was changed.

Even though linting took much less time than testing, there was still room to save around one minute per PR using this approach. I've tested it on my fork.

The working is analogous to the `test-changed` task, except this skips the result archive part.

---

Reviewer Resources:

### DIFF
--- a/bin/lint-changed-exercise
+++ b/bin/lint-changed-exercise
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Determine the base branch of the PR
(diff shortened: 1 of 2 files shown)


bin/lint-changed-exercise
kahgoh, line 1: This script seems quite similar to the build changed one in the other PR. Have you considered seeing if there is some way to share the common parts?

jagdish-15:
Superseded by #3126

PR #3145, opened 2026-06-23
Fixed faulty test.

MiaLisen:
Fault was: Test expects 3 Strings instead of 2 as as described in the task.

# Pull Request

---

Reviewer Resources:


kahgoh (changes requested):
I don't see anything wrong with the test. The instruction for task 5 states that the default travel method is walking (i.e. if only character and destination is given, then the travel method is walking). The relevant part:

> For convenience, players are allowed to omit mentioning their travel method, in which case walking will be assumed to be the travel method.

MiaLisen:
Omg.. Sorry you're right. ><

PR #3146, opened 2026-06-24
simple-linked-list: sync missing test cases and add peek/toList

### LINKED ISSUE #3142: [configlet] simple-linked-list: missing test cases
## Missing test cases for `simple-linked-list`

Part of #2959.

The following test cases from [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/simple-linked-list) are not yet implemented in this track:

- `962d998c-c203-41e2-8fbd-85a7b98b79b9` count -> Empty list has length of zero
- `9760262e-d7e4-4639-9840-87e2e2fbb115` count -> Singleton list has length of one
- `d9955c90-637c-441b-b41d-8cfb48e924a8` count -> Non-empty list has correct length
- `0c3966db-58f9-4632-b94c-8ea13e54c2c8` pop -> Pop from empty list is an error
- `a4f9d2e1-7425-49ef-9ee8-6c0cb3407cf0` pop -> Can pop from singleton list
- `6dcbb2c9-d98a-47bc-a010-9c19703d3ea2` pop -> Can pop from non-empty list
- `e83aade9-f030-4096-aaf0-f9dc6491e6cf` pop -> Can pop multiple items
- `5c46bcf2-c0a9-4654-ae17-f3192436fcf1` pop -> Pop updates the count
- `70d747a1-2e84-4ebc-bc3f-dcbee6a05f6b` push -> Can push to an empty list
- `f3197f0a-1fea-45a5-939f-4a5ea60387ec` push -> Can push to an empty list
- `391e332e-1f91-4033-b1e0-0e0c17812fa7` push -> Can push to a non-empty list
- `ed4b0e01-3bbd-4895-af25-152b5914b3da` push -> Push updates count
- `41666790-b932-4e5a-b323-e848a83d12d5` push -> Push and pop
- `930a4a5c-76f6-47ec-9be3-4e70993173a1` peek -> Peek on empty list is an error
- `43255a50-d919-4e81-afce-e4a271eaedbd` peek -> Can peek on singleton list
- `48353020-e25d-4621-a854-e35fb1e15fa7` peek -> Can peek on non-empty list
- `96fcead9-a713-46c2-8005-3f246c873851` peek -> Peek does not change the count
- `7576ed05-7ff7-4b84-8efb-d34d62c110f5` peek -> Can peek after a pop and push
- `b97d00b6-2fab-435d-ae74-3233dcc13698` toList LIFO -> Empty linked list to list is empty
- `eedeb95f-b5cf-431d-8ad6-5854ba6b251c` toList LIFO -> To list with multiple values
- `838678de-eaf3-4c14-b34e-7e35b6d851e8` toList LIFO -> To list after a pop
- `03fc83a5-48a8-470b-a2d2-a286c5e8365f` toList FIFO -> Empty linked list to list is empty
- `1282484e-a58c-426a-972e-90746bda61fc` toList FIFO -> To list with multiple values
- `05ca3109-1249-4c0c-a567-a3b2f8352a7c` toList FIFO -> To list after a pop
- `5e6c1a3d-e34b-46d3-be59-3f132a820ed5` reverse -> Reversed empty list has same values
- `93c87ed3-862a-474f-820b-ba3fd6b6daf6` reverse -> Reversed singleton list is same list
- `92851ebe-9f52-4406-b92e-0718c441a2ab` reverse -> Reversed non-empty list is reversed
- `1210eeda-b23f-4790-930c-7ac6d0c8e723` reverse -> Reversed non-empty list is reversed
- `9b53af96-7494-4cfa-9b77-b7366fed5c4c` reverse -> Double reverse

### How to help

For detailed instructions on how to fetch configlet and update the tests, please see the **"How to do this task"** section in the main tracking issue:
👉 **[Read the instructions here](https://github.com/exercism/java/issues/2959)**

_This issue is managed automatically by the [Create Configlet Sync Issues](https://github.com/exercism/java/actions/workflows/create-configlet-sync-issues.yml) workflow._

AdityaPudasaini:
# Pull Request

Syncs the simple-linked-list exercise with the latest problem specifications.

Changes:
- Added 29 missing test cases via `configlet sync`
- Added `peek()` method to stub and reference solution
- Added `toList()` method to stub and reference solution
- Updated `.meta/tests.toml` with all new test UUIDs

All 34 tests pass locally.

Fixes #3142 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/simple-linked-list/src/main/java/SimpleLinkedList.java
+++ b/exercises/practice/simple-linked-list/src/main/java/SimpleLinkedList.java
@@ -22,9 +22,6 @@ class SimpleLinkedList<T> {
     List<T> toList() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.toList() method.");
     }
-    T[] asArray(Class<T> clazz) {
-        throw new UnsupportedOperationException("Please implement the SimpleLinkedList.asArray() method.");
-    }
     int size() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.size() method.");
     }
--- a/exercises/practice/simple-linked-list/src/main/java/SimpleLinkedList.java
+++ b/exercises/practice/simple-linked-list/src/main/java/SimpleLinkedList.java
@@ -1,28 +1,30 @@
+import java.util.List;
+
 class SimpleLinkedList<T> {
     SimpleLinkedList() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList() constructor.");
     }
-
     SimpleLinkedList(T[] values) {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList(T[]) constructor.");
     }
-
     void push(T value) {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.push() method.");
     }
-
     T pop() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.pop() method.");
     }
-
+    T peek() {
+        throw new UnsupportedOperationException("Please implement the SimpleLinkedList.peek() method.");
+    }
     void reverse() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.reverse() method.");
     }
-
+    List<T> toList() {
+        throw new UnsupportedOperationException("Please implement the SimpleLinkedList.toList() method.");
+    }
     T[] asArray(Class<T> clazz) {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.asArray() method.");
     }
-
     int size() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.size() method.");
     }
--- a/exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
+++ b/exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
@@ -8,87 +8,6 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 public class SimpleLinkedListTest {
 
     @Test
-    @DisplayName("A new list is empty")
-    public void aNewListIsEmpty() {
-        SimpleLinkedList<Integer> list = new SimpleLinkedList<>();
-        assertThat(list.size()).isEqualTo(0);
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Create list from array")
-    public void canCreateFromArray() {
-        Character[] values = new Character[]{'1', '2', '3'};
-        SimpleLinkedList<Character> list = new SimpleLinkedList<Character>(values);
-        assertThat(list.size()).isEqualTo(3);
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Popping an empty list throws NoSuchElementException")
-    public void popOnEmptyListWillThrow() {
-        SimpleLinkedList<String> list = new SimpleLinkedList<String>();
-        assertThatExceptionOfType(NoSuchElementException.class).isThrownBy(list::pop);
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Pop returns last added element (LIFO)")
-    public void popReturnsLastAddedElement() {
-        SimpleLinkedList<Integer> list = new SimpleLinkedList<Integer>();
-        list.push(9);
-        list.push(8);
-        assertThat(list.size()).isEqualTo(2);
-        assertThat(list.pop()).isEqualTo(8);
-        assertThat(list.pop()).isEqualTo(9);
-        assertThat(list.size()).isEqualTo(0);
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Reverse reverses the list order")
-    public void reverseReversesList() {
-        SimpleLinkedList<String> list = new SimpleLinkedList<String>();
-        list.push("9");
-        list.push("8");
-        list.push("7");
-        list.push("6");
-        list.push("5");
-        list.reverse();
-        assertThat(list.pop()).isEqualTo("9");
-        assertThat(list.pop()).isEqualTo("8");
-        assertThat(list.pop()).isEqualTo("7");
-        assertThat(list.pop()).isEqualTo("6");
-        assertThat(list.pop()).isEqualTo("5");
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Can return list as an array")
-    public void canReturnListAsArray() {
-        SimpleLinkedList<Character> list = new SimpleLinkedList<Character>();
-        list.push('9');
-        list.push('8');
-        list.push('7');
-        list.push('6');
-        list.push('5');
-        Character[] expected = {'5', '6', '7', '8', '9'};
-        assertThat(list.asArray(Character.class)).isEqualTo(expected);
-    }
-
-    @Disabled("Remove to run test")
-    @Test
-    @DisplayName("Can return empty list as an empty array")
-    public void canReturnEmptyListAsEmptyArray() {
-        SimpleLinkedList<Object> list = new SimpleLinkedList<Object>();
-        Object[] expected = {};
-        assertThat(list.asArray(Object.class)).isEqualTo(expected);
-    }
-
-    // count tests
-
-    @Disabled("Remove to run test")
-    @Test
     @DisplayName("count -> Empty list has length of zero")
     public void countEmptyListHasLengthOfZero() {
         SimpleLinkedList<Integer> list = new SimpleLinkedList<>();
@@ -110,8 +29,6 @@ public class SimpleLinkedListTest {
         SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2, 3});
         assertThat(list.size()).isEqualTo(3);
     }
-
-    // pop tests
 
     @Disabled("Remove to run test")
     @Test
@@ -158,8 +75,6 @@ public class SimpleLinkedListTest {
         assertThat(list.size()).isEqualTo(0);
     }
 
-    // push tests
-
     @Disabled("Remove to run test")
     @Test
     @DisplayName("push -> Can push to an empty list")
@@ -201,8 +116,6 @@ public class SimpleLinkedListTest {
         assertThat(list.pop()).isEqualTo(1);
         assertThat(list.size()).isEqualTo(0);
     }
-
-    // peek tests
 
     @Disabled("Remove to run test")
     @Test
@@ -251,8 +164,6 @@ public class SimpleLinkedListTest {
         assertThat(list.peek()).isEqualTo(3);
     }
 
-    // toList tests
-
     @Disabled("Remove to run test")
     @Test
     @DisplayName("toList LIFO -> Empty linked list to list is empty")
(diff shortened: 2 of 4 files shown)


AdityaPudasaini:
These changes do affect test outcomes as they add 29 missing test cases to sync with the latest problem specifications. Happy to add [no important files changed] to the commit message if the maintainers prefer, but I believe re-running student tests is appropriate here.

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -88,4 +85,273 @@ public void canReturnEmptyListAsEmptyArray() {
...
     }
 
+    // count tests
+
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("count -> Empty list has length of zero")
+    public void countEmptyListHasLengthOfZero() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>();
+        assertThat(list.size()).isEqualTo(0);
```
kahgoh, line 14: This test is the same as `aNewTestIsEmpty`. In fact, I noticed there are more tests that have also been duplicated (e.g. `popFromEmptyListIsAnError` is a copy of `popOnEmptyListWillThrow`). Could you please remove the duplicates? I suspect you'll be able to remove the "old" version of the tests.

exercises/practice/simple-linked-list/src/main/java/SimpleLinkedList.java
```diff
@@ -1,29 +1,31 @@
...
     void reverse() {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.reverse() method.");
     }
-
+    List<T> toList() {
+        throw new UnsupportedOperationException("Please implement the SimpleLinkedList.toList() method.");
+    }
     T[] asArray(Class<T> clazz) {
         throw new UnsupportedOperationException("Please implement the SimpleLinkedList.asArray() method.");
     }
```
kahgoh: I think it is a bit strange to have both these methods for this exercise. I think we should choose one to go with and remove the tests for the other. If we go with `asArray` then the `toList` tests can use `asArray` instead.

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -88,4 +85,273 @@ public void canReturnEmptyListAsEmptyArray() {
...
+    @DisplayName("toList FIFO -> To list after a pop")
+    public void toListFifoToListAfterAPop() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>();
+        list.push(1);
+        list.push(2);
+        list.push(3);
+        assertThat(list.pop()).isEqualTo(3);
+        list.push(4);
+        assertThat(list.toList()).containsExactly(4, 2, 1);
+    }
```
kahgoh: It doesn't really make sense to have _both_ FIFO and LIFO tests. It should have just the FIFO set _or_ the LIFO set. This is mentioned in the comments at the top of the [canonical-data](https://github.com/exercism/problem-specifications/blob/e2c23e29aa99997be4c95e48b549681b88f4b380/exercises/simple-linked-list/canonical-data.json#L7) 👇🏻:

>    "The linked list implements a LIFO. A pop retrieves the last-in value.    ",
>    "As such, the toList operation can be thought as being build using pops,  ",
>    "which means list([1, 2, 3]).toList() would return [3, 2, 1].             ",
>    "Conversely, the toList is more intuitive to some if it returns data matching how the list was build, ",
>    "so list([1, 2, 3]).toList() would return [1, 2, 3].",
>    "",
>    "To preserve both options, this data has two test groups: toList LIFO and toList FIFO.",
>    "The tests in these groups are mutually exclusive.",

Also, the expectations don't match the canonical data in some of your tests either. For example, the expectation for `toListLifoToListWithMultipleValues` should be `[3, 2, 1]`, not `[1, 2, 3]`, and for `toListFifoToListAfterAPop`, it should be `[1, 2, 4]`, not `[4, 2, 1]`.

Anyway, to fix this:
1. Decide whether to use the LIFO set or FIFO set. I'd prefer using the set that aligns best with the existing `asArray` method. I think that is the LIFO set.
2. In `tests.toml` mark the removed tests as skipped. To do this, you can add `include = false`. See line 29 in the one for [hamming](https://github.com/exercism/java/blob/0443ae57ff3d4842259bd30bc0b71aeb593fe39b/exercises/practice/hamming/.meta/tests.toml#L29) for an example.

kahgoh (changes requested):
Hi @AdityaPudasaini, thanks for raising the PR to update the tests.

> These changes do affect test outcomes as they add 29 missing test cases to sync with the latest problem specifications. 

Actually, some of the tests have already been written - the `tests.toml` just haven't been synced so `configlet` couldn't see them. As a result I noticed some of the tests have now been duplicated (this is mentioned in one of comments below).

> Happy to add [no important files changed] to the commit message if the maintainers prefer, but I believe re-running student tests is appropriate here.

No need to add this comment - we'll add it at merge time if we think it is appropriate. But you are also right - adding new tests and new methods is going to break all the existing solutions.

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
AdityaPudasaini: Thanks for the detailed review @kahgoh! I'll remove the duplicate tests, go with the LIFO set, mark the FIFO tests as include = false in tests.toml, and fix the expectations. I'll push the changes shortly.

AdityaPudasaini: Hi @kahgoh , I've addressed all the feedback:

Removed duplicate tests (old aNewListIsEmpty, popOnEmptyListWillThrow, etc.)
Kept only the LIFO set and marked FIFO tests as include = false in tests.toml
Fixed the toList expectations to match canonical data
All 27 tests pass locally

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -1,68 +1,238 @@
...
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1});
+        assertThat(list.pop()).isEqualTo(1);
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("pop -> Can pop from non-empty list")
+    public void canPopFromNonEmptyList() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2});
+        assertThat(list.pop()).isEqualTo(1);
```
kahgoh: Should this be equal to 2?

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -1,68 +1,238 @@
...
+    }
+
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("pop -> Can pop multiple items")
+    public void canPopMultipleItems() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2});
+        assertThat(list.pop()).isEqualTo(1);
+        assertThat(list.pop()).isEqualTo(2);
+    }
```
kahgoh, line 64: Should this be equal to 2 then 1?

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -1,68 +1,238 @@
...
+    @Test
+    @DisplayName("pop -> Pop updates the count")
+    public void popUpdatesTheCount() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2});
         assertThat(list.size()).isEqualTo(2);
-        assertThat(list.pop()).isEqualTo(8);
-        assertThat(list.pop()).isEqualTo(9);
+        assertThat(list.pop()).isEqualTo(1);
+        assertThat(list.size()).isEqualTo(1);
+        assertThat(list.pop()).isEqualTo(2);
```
kahgoh, line 72: Similar to other comment, should the expectation for `pop` be 2 then 1?

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -1,68 +1,238 @@
...
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("reverse -> Double reverse")
+    public void doubleReverse() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2, 3});
+        list.reverse();
+        list.reverse();
+        assertThat(list.pop()).isEqualTo(1);
+        assertThat(list.pop()).isEqualTo(2);
+        assertThat(list.pop()).isEqualTo(3);
```
kahgoh: Should the the results from `pop` be 3, 2 then 1?

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
```diff
@@ -1,68 +1,238 @@
...
+    @Disabled("Remove to run test")
+    @Test
+    @DisplayName("reverse -> Reversed non-empty list is reversed")
+    public void reversedNonEmptyListIsReversed() {
+        SimpleLinkedList<Integer> list = new SimpleLinkedList<>(new Integer[]{1, 2, 3});
+        list.reverse();
+        assertThat(list.size()).isEqualTo(3);
+        assertThat(list.pop()).isEqualTo(3);
+        assertThat(list.pop()).isEqualTo(2);
+        assertThat(list.pop()).isEqualTo(1);
```
kahgoh: Should the result from `pop` be 1, 2 then 3?

kahgoh (changes requested):
Thanks, that is looking better! Some of the expectations are still different to what is in the canonical data though (see comments below). 

Also, did you see my [comment](https://github.com/exercism/java/pull/3146#discussion_r3467355910) about having both `toList` and `asArray`? What were your thoughts about having just one or the other?

exercises/practice/simple-linked-list/src/test/java/SimpleLinkedListTest.java
AdityaPudasaini: Hi @kahgoh, thanks for the continued feedback! I want to make sure I understand correctly before making changes.
Regarding canPopFromNonEmptyList — the current constructor pushes elements in reverse order so that 1 ends up at the head, making pop() return 1 first. Should the constructor instead push in forward order so that the last element 2 is at the head?
Also regarding having both toList and asArray — I'd suggest keeping toList and removing asArray tests since toList returns a List<T> which is more idiomatic modern Java. Happy to go with whatever you prefer though.
Please let me know and I'll update accordingly!

kahgoh: > Should the constructor instead push in forward order so that the last element 2 is at the head?

Yes, I think that's what the canonical data suggests. 

> Also regarding having both toList and asArray — I'd suggest keeping toList and removing asArray tests since toList returns a List which is more idiomatic modern Java.

Yes to keeping `toList` and removing `asArray`.

AdityaPudasaini: Hi @kahgoh, updated! Fixed the constructor to push in forward order, updated all test expectations accordingly, removed asArray and kept toList. All 25 tests pass locally.

kahgoh (approved):

jagdish-15 (approved):

PR #3152, opened 2026-07-26
Import DateTimeFormatter in AppointmentScheduler

devdavco:
Added import for DateTimeFormatter to format date and time.

# Pull Request

---

Reviewer Resources:


kahgoh:
I don't think  the stub need to have the import statement up front - isn't it something the students can add? The introduction already lists the class as `java.time.format.DateTimeFormatter`.

